Issue #10, opened 2017-12-29
Move from stock etree to lxml etree


wrznr:
It's superior (and offers full XPATH functionality).

cneud:
Agreed - better to do this sooner than later. http://lxml.de/api.html#lxml-etree has come handy in the past. 
[I'll look into it](https://github.com/cneud/pyocrd/tree/lxml).

wrznr:
Fixed via #12.

PR #12, opened 2018-02-13
change xml.etree to lxml.etree

cneud:
Seems as simple as replacing the import statement (at this early state of dev).


wrznr:
Resolves #10.

PR #32, opened 2018-03-16
Recognition of line images with tesseract

### LINKED ISSUE #17: Text recognition: Tesseract interface
Implement text recognition using tesserocr.

wrznr:
also fixes #17 and #20 

There is still an issue with the recognition: it did work very, very poorly on the sample image from the spec repository. Might have to do something with tesseract invocation or the missing preprocessing (i.e. binarization).

### DIFF
--- a/ocrd/init/initializing.py
+++ b/ocrd/init/initializing.py
@@ -89,7 +89,10 @@ class Initializer:
         """
         Loads or creates missing PAGE XML from internal METS tree.
         """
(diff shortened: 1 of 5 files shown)


kba:
I've implemented the same algorithm in #40. It works, only produces gibberish but the main problem is that it's incredibly slow to do it like this, line image by line image. On a single core of an i7 6600U, it takes almost a second per line.

We should optimize this (alternativeImage) but also discuss how we support multicore processing. In this case: Have the recognize.tesserocr module create a Pool and spawn workers that work on a line each. Perhaps the workspace module could also offer an API for processors to use.

Issue #61, opened 2018-04-13
Naming conflict in model spec


wrznr:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/kmw/projects/dwds/ocr/src/ocrd_kraken/env/lib/python3.5/site-packages/ocrd/model/yaml/ocrd_oas3.spec.yml'
```
but we have
`pyocrd/ocrd/model/yaml/ocrd_oas3.yml`

PR #62, opened 2018-04-13
* renamed according to ocrd/constants.py

wrznr:
@kba @VolkerHartmann  Only a minor change, but prevents from continuing with `ocr_kraken`.


wrznr:
Fixes #61

wrznr:
I see. So a `make` call would fix my problem?

wrznr:
Not the right way!

PR #63, opened 2018-04-13
Update spec

kba:
#61 #62


kba:
I'll merge this, since it breaks CI and MP tools.

PR #75, opened 2018-05-08
Refactor CLI

kba:
* Subcommands in dedicated files
* `workspace` subcommand nestable
  * `workspace add` to add files
  * `workspace pack` to pack to ZIP
  * `workspace unpack` to unpack from ZIP
  * `workspace validate` to validate METS for consistency and compliance


ocrd/cli/workspace.py
```diff
@@ -0,0 +1,146 @@
...
+''')
+@click.option('-m', '--mets-url', help="METS URL to validate")
+@pass_workspace
+def validate_workspace(ctx, mets_url=None):
+    if mets_url is None:
+        mets_url = 'file://%s/mets.xml' % ctx.directory
+    report = WorkspaceValidator.validate_url(ctx.resolver, mets_url)
+    print(report.to_xml())
+    if not report.is_valid:
+        return 128
```
wrznr: Magic number?

kba: Yeah, `128` is random. Must just be something non-zero, so you can do

```sh
if ocrd workspace validate -m https://server/path/to/mets.xml;then
  echo 'ok'
else
	echo 'not ok'
fi
```

But it didn't work because return of the click handler !== retval of process. I'll fix that.

PR #80, opened 2018-05-22
WIP: Wrap execeutable webservice

kba:
This is blocked by OCR-D/spec#30 and needs cleaning up, so WIP for now.

Basically, it allows to do

```
ocrd server wrap-executable --port 1234 ocrd-kraken-binary
```

which will wrap any spec-compliant executable, inspect its [tool description](https://ocr-d.github.io/ocrd_tool) and start a webservice for it that adheres to [our HTTP API specs](https://ocr-d.github.io/swagger).

Then these calls can be used for the same thing:

```
ocrd-kraken-binary -m /path/to/mets.xml

==

http --form POST http://localhost:1234 mets@/path/to/mets.xml
```

We really only need the names of the executables provided, the rest (CLI call, workflow integration, HTTP service, Swagger, service list etc) can be generated from introspection.

### DIFF
--- a/ocrd/__init__.py
+++ b/ocrd/__init__.py
@@ -1,4 +1,4 @@
-from ocrd.processor.base import run_processor, run_cli, Processor
+from ocrd.processor.base import run_processor, run_executable, Processor
 from ocrd.model import OcrdMets, OcrdExif, OcrdFile, OcrdSwagger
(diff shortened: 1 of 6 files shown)


wrznr:
@kba What is the status here?

PR #112, opened 2018-06-18
set groupId when adding files, fix #110

### LINKED ISSUE #110: groupId isn't set while adding file to workspace.
```
ocrd workspace add --file-grp OCR-D-TEST --file-id test_file1 --group-id image0001 --mimetype image/tiff anyImage.tif
```
Will not check for duplicated entries (same ID) and won't create a groupID.

if file is the original image (USE=OCR-D-IMG) GROUPID and ID has to be identical!
There should be at least a test if a file with ID=GROUPID in fileGrp "OCR-D-IMG" exists

kba:


kba:
> ocrd_mets line 62: Why don't you reuse code from the find command?

Because find_files returns a list of files, not a fileGrp. Return value is not indicative of whether fileGrp exists or not.

> Is groupId still optional? Then you should test both cases.

It's optional. I'll add a test.

Issue #143, opened 2018-07-18
cache_enabled


finkf:
Maybe it is better to default cache_enabled to false. 
If one needs the extra speedup one can enable it manually.
Its weird during development if changes are not visible :)

finkf:
see #132

kba:
Caching was purely for practical reasons, transmitting image files via mobile network before I had high speed internet available. I'll fix this next, thanks for reminding.

kba:
fixed (caching removed) in v0.5.0

Issue #132, opened 2018-06-27
Caching of parameter files


kba:
`/tmp/cache-pyocrd` is a major nuisance.

kba:
fixed (caching removed) in v0.5.0

Issue #126, opened 2018-06-19
Enable cache only explicitly?


kba:
The `resolver_cache` was implemented solely to save bandwidth while developing. There is no sensible cache invalidation strategy and I see no merit in spending time on this.

Disable the cache by default and require a flag or cache directory to be explicitly set to enable it?

VolkerHartmann:
That's fine for me. In this case the users are aware that they work with a cached workspace.

Issue #92, opened 2018-06-12
Allow disabling/emptying cache in CLI


kba:
And/or have a sane cache invalidation strategy in place.

kba:
The cache is on the resolver level, so its file-type-agnostic. The only reason a URL won't be cached at the moment is when HTTP request fails.

The background of caching was: Slow/Limited internet connection and not wanting to download the same multi-MB TIFF for each run of the tests. It's a very simple implementation, effectively 30 lines of code or so. The simplest "solution" would be to make the default **not** to cache.

The only solution to invalidate a cache entry for a URL is to delete the file corresponding to it in the cache directory. There is no programmatic way to do this.

We could add cache control functionality to the CLI, like `ocrd cache purge`, `ocrd cache rm FILE` etc. with little effort but it's still hacky.

kba:
cache can be circumvented with `--no-cache` #100

PR #144, opened 2018-07-19
:fire: Drop all caching, #132, #143, #126, #100, #92, #45, #85

kba:
> There are only two hard things in Computer Science: cache invalidation and naming things.
>
> -- Phil Karlton

The caching was deliberate premature optimization but now it's just a source for confusing errors.


ocrd/resolver.py
```diff
@@ -160,25 +155,15 @@ def download_to_directory(self, directory, url, basename=None, overwrite=False,
...
-                with open(outfilename, 'wb') as outfile:
-                    response = requests.get(url)
-                    if response.status_code != 200:
-                        raise Exception("Not found: %s (HTTP %d)" % (url, response.status_code))
-                    outfile.write(response.content)
-
-        if self.cache_enabled and not cached_filename:
-            cached_filename = self.cache.put(url, filename=outfilename)
-            log.debug("Stored in cache <%s> at '%s'", url, cached_filename)
+            with open(outfilename, 'wb') as outfile:
```
wrznr: Indent correct here?

PR #149, opened 2018-07-24
:fire: Use the directory of an existing mets.xml file as the working dir

kba:
…if none other was specified.

This will use the "original" mets file for output files/file groups and write files to the directory of the original files.

Less clean but more intuitive.

Needs testing to make sure we're not clobbering files.

### DIFF
--- a/ocrd/resolver.py
+++ b/ocrd/resolver.py
@@ -159,10 +159,10 @@ class Resolver(object):
         if url.startswith('file://'):
             self._copy_or_symlink(url[len('file://'):], outfilename, prefer_symlink)
         else:
+            response = requests.get(url)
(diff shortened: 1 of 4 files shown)


kba:
> So we have to copy mets.xml in beforehand, right?

That was the old behavior. The new behavior is that no temporary directory is created when creating a workspace from a URL if that URL is a `file://` URL. Instead, we reuse the existing METS XML file.

> That's fine for me. But all other files must not be changed! (e.g. page.xml)

That behavior is not changed, this should only affect directory creation and copying/downloading of the METS file. If you find any flaws in the logic do not hesitate to report, I fear this is getting a bit convoluted.

PR #150, opened 2018-07-26
Replace exiftool

kba:


ocrd/model/ocrd_exif.py
```diff
@@ -77,14 +77,23 @@ def from_filename(image_filename):
             raise Exception("Must pass 'image_filename' to OcrdExif.from_filename")
         with exiftool.ExifTool() as et:
```
cneud: `exiftool` still being used here?

kba: That was 2 commits ago, check https://github.com/OCR-D/core/pull/150/commits/015832233000e89619bc3f0edc9dd9e2196aba3f

ocrd/model/ocrd_exif.py
```diff
@@ -75,36 +9,34 @@ class OcrdExif(object):
...
+            self.resolutionUnit = 'cm' if img.tag[296] == 3 else 'inches'
+        elif img.format == 'JPEG':
+            self.xResolution = img.info['jfif_density'][0]
+            self.yResolution = img.info['jfif_density'][1]
+            self.resolutionUnit = img.info['jfif_unit']
+        elif img.format == 'PNG' and 'dpi' in img.info:
+            print(img.info['pnginfo'])
+            self.xResolution = img.info['dpi'][0]
+            self.yResolution = img.info['dpi'][1]
+        else:
```
cneud: There is a test below, but we do not support JP2K yet in the exif reading code?

ocrd/model/ocrd_exif.py
kba: default value should be `1`, `-1` was for testing, I'll fix that.

cneud: Ok, spotted it!

PR #151, opened 2018-07-26
replace xmllint callout with lxml solution, fix #72

### LINKED ISSUE #72: Drop dependency on xmllint
xmllint from libxml2 arguably produces the most predictable pretty-printed XML but it should be an optional dependency, gracefully fall back to lxml tools when not available.

kba:


Makefile
```diff
@@ -41,8 +41,7 @@ deps-ubuntu:
 	sudo apt install -y \
 		python3 \
 		python3-pip \
-		libimage-exiftool-perl \
-		libxml2-utils
+		libimage-exiftool-perl
```
cneud: I assume this reference to `libimage-exiftool-perl` will be gone once https://github.com/OCR-D/core/pull/150 is merged?

kba: Correct. Without it, the CI builds break, c.f. https://circleci.com/gh/kba/ocrd-core/217 vs https://circleci.com/gh/kba/ocrd-core/219

cneud: Oh yeah, that's why we have tests! :-)

PR #155, opened 2018-08-04
add parameter description property to ocrd_tool schema 

bertsky:
(as used by website [example](https://ocr-d.github.io/ocrd_tool))


kba:
Thanks for spotting this. That schema should be a 1:1 copy of [the one we keep in spec](https://github.com/OCR-D/spec/blob/master/ocrd_tool.schema.yml), so the change should happen over there. I will fix it there.

PR #159, opened 2018-08-14
Handle any number of unordered points

finkf:


kba:
Great, with a test even, thanks! Will merge tomorrow.

PR #165, opened 2018-08-20
resolver: when downloading a file-url fall back to url for basename, …

kba:
When downloading, to a known directory, a `file://` URL referencing a path within that directory, the resolver assumed the URL to be opaque and created a safe filename version of that URL. Which adds up to exponentially long filenames when repeated.

This PR catches that case (basename not set, file-URL pointing to the workspace directory) and should result in a no-op if that file already exists.

### DIFF
--- a/ocrd/resolver.py
+++ b/ocrd/resolver.py
@@ -119,6 +119,8 @@ class Resolver(object):
         """
         Download a file to the workspace.
 
+        Early Shortcut: If iit's a file-URL and already in the directory, keep it there.
+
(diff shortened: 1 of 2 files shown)


ocrd/resolver.py
```diff
@@ -119,6 +119,8 @@ def download_to_directory(self, directory, url, basename=None, overwrite=False,
         """
         Download a file to the workspace.
 
+        Early Shortcut: If iit's a file-URL and already in the directory, keep it there.
```
cneud: Typo: "If it's a file-URL"

PR #170, opened 2018-08-27
merge_ocr_txt: Run only as script

kba:
merge_ocr_txt is a  legacy script from the pyocrd days of yonder. This commit will make sure that it is only run when explicitly called as a script.

I just noticed it when rebuilding the apidocs with sphinx.

@wrznr Do you still use that script? Could we factor it out of core?


wrznr:
We will factor it out as ocrd-merge-txt, okay?

kba:
Fine with me 👍

Issue #173, opened 2018-08-28
logging levels now override CLI setting by config file


bertsky:
Since #164 and v0.8 (with `CONFIG_PATHS` in _core/ocrd/logging.py_), the log level set via command line parameter (e.g. "-l DEBUG") can get overriden by settings in those config files. I think it should be the other way round.

kba:
Setting the log level via command line switch does this:

```
logging.getLogger('').setLevel(logging.getLevelName(value))
```

It sets the (minimum) log level for the root logger. More specific rules in the `ocrd_logging.py` files will override this.

What do you mean by "the other way round"? Do you want the log level specified on the command line override all loggers? That would defeat the purpose of the `ocrd_logging.py` override mechanism. And it's kinda difficult because loggers get created at runtime.

Couldn't you just disable the customizations in the `ocrd_logging.py` that override the root logger settings?

bertsky:
If I set the log level via command line switch it will affect _all_ loggers (e.g. DEBUG) – this is to be expected.

But the specific rules in `ocrd_logging.py` files will be _unaffected_ by that CLI setting (e.g. stay at WARN) – this goes against the general principle; command line should always have precedence over config files.

bertsky:
> I understand what you're saying but I still don't understand why you're using both mechanisms (command line and config file). Yes, CLI params should override config file default but not more specific rules. 

Here I disagree. If the CLI is defined such that specifics are
impossible to articulate, so be it: it is a last resort then. Why not
have a (coarse but effective) last resort at CLI and a (more refined but
elaborate) configuration facility? (Isn't this what cfg vs cli is all
about after all?)

(Of course, one could define the CLI differently, like applying loglevel
only to the top processor's logger, or by adding another parameter for
the name of the logger to be changed.)

> Minimum level for a specific logger namespace prefix should override the generic rule.

True for everything between defaults in code and any config files. But
having an override mechanism above those only makes sense if it actually
overrides them.

>> If I set the log level via command line switch it will affect all loggers (e.g. DEBUG) – this is to be expected.
> 
> No, it will affect the root logger. The config file is parsed afterwards and its more specific rules overrride the root logger.

That's strange. I was merely describing my observation there. If I use
"-l DEBUG" I get debug level messages from all loggers (`ocrd.resolver`,
`ocrd.processor`, `ocrd.workspace` etc). Otherwise I only get info level
messages. Is that not intentional?

[...]

PR #181, opened 2018-09-17
Logging cli

kba:
This should implement the behavior we agreed on in #173.

### DIFF
--- a/ocrd/decorators.py
+++ b/ocrd/decorators.py
@@ -5,7 +5,7 @@ import click
 from ocrd.constants import VERSION as OCRD_VERSION
 from ocrd.resolver import Resolver
 from ocrd.processor.base import run_processor
-from ocrd.utils import setOverrideLogLevel
+from ocrd.logging import setOverrideLogLevel
 
 def _set_root_logger_version(ctx, param, value):    # pylint: disable=unused-argument
     setOverrideLogLevel(value)
--- a/ocrd/decorators.py
+++ b/ocrd/decorators.py
@@ -5,10 +5,10 @@ import click
 from ocrd.constants import VERSION as OCRD_VERSION
 from ocrd.resolver import Resolver
 from ocrd.processor.base import run_processor
-from ocrd.utils import logging
+from ocrd.utils import setOverrideLogLevel
 
-def _set_root_logger_version(ctx, param, value):
-    logging.getLogger('').setLevel(logging.getLevelName(value))
+def _set_root_logger_version(ctx, param, value):    # pylint: disable=unused-argument
+    setOverrideLogLevel(value)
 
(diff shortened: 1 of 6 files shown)


ocrd/decorators.py
```diff
@@ -49,8 +49,8 @@ def cli(mets_url):
         click.option('-p', '--parameter', type=click.Path()),
         click.option('-J', '--dump-json', help="Dump tool description as JSON and exit", is_flag=True, default=False),
         click.option('-l', '--log-level', help="Log level",
-            type=click.Choice(['OFF', 'ERROR', 'WARN', 'INFO', 'DEBUG', 'TRACE']),
-            default='INFO', callback=_set_root_logger_version),
+                     type=click.Choice(['OFF', 'ERROR', 'WARN', 'INFO', 'DEBUG', 'TRACE']),
+                     default='INFO', callback=_set_root_logger_version),
```
bertsky: I believe this must not have a `default` any more. If it does have a default, then the callback will **always** be triggered, overriding all configuration by the `INFO` level. (I have tested this, too.)

kba: Well spotted, thanks. We should document that the default log level `INFO` and explicitly providing `--log-level INFO` cause different behaviors when using core.

kba: should be fixed and tested now.

bertsky: It is, thanks. But do we really need that bitter pill of not having a default (in the user sense, not in the `click.option` implementation sense) any more? We would have to set the respective level to the root logger in a configuration file for that now. Doesn't the spec say that INFO should be the default setting? 

(I think all we need is an additional `logging.getLogger('').setLevel(logging.INFO)` in ocrd/logging.py.)

bertsky: ok, I guess you did then :-) I am on your 8b99b512 ... there is only a setting for `PIL` in there.

bertsky: Since we already stumbled across that other twist, how about adding `self.assertEqual(logging.getLogger('').getEffectiveLevel(), logging.INFO)` as first assertion to `test_loglevel_override`?

kba: That should also be in the Code. We think alike apparently :D Ill push tomorrow sorry for the confusion thanks for reviewing!

bertsky: Works like a charm, thanks! 

No idea what causes these new assertion failures. Tests pass for me with python 3.6.5, 3.4.3, 2.7.15rc1, and 2.7.6.

kba: Thanks for testing. I guess travis has a different test harness. It works on CircleCI and Scrutinizer, but travis possibly hooks into `setUp` and sets root logger level to `DEBUG` or something like that.

kba: Fixed with explicitly initiating loggin in `setUp`.

PR #188, opened 2018-10-17
Remove unneeded pass statements

stweil:
This fixes warnings from LGTM:

    Unnecessary 'pass' statement.


stweil:
There are [many more warnings and recommendations](https://lgtm.com/projects/g/OCR-D/core/alerts).

kba:
That file is auto-generated from XSD. It's not perfect code but consistent. We're using [pylint to check code style](https://github.com/OCR-D/core/blob/master/.pylintrc).

Good point though that we should re-enable the `unused XYZ` warnings, I've disabled it at some point because they were mostly noise.

kba:
@stweil Is there a way to filter the warnings about the file `ocrd_page_generateds.py` on lgtm.com?

stweil:
Pull request #214 should do that. It excludes the whole file for LGTM.

stweil:
Closing this PR as it was for a generated file which should be ignored.

stweil:
The generated file was renamed, so the LGTM rules need an update. I'll send a PR.

Issue #46, opened 2018-03-29
Line detection via ocrd: PAGE XML is overwritten, result is empty


wrznr:
Using
```shell
ocrd process -m ocrd-assets/dist/mets.xml characterize/exif segment-region/tesserocr
```
works great (i.e.  metadata and regions show up in OUTPUT PAGE XML).
However, adding line detection
```shell
ocrd process -m ocrd-assets/dist/mets.xml characterize/exif segment-region/tesserocr segment-line/tesserocr
```
results in “empty” XML:
```xml
<PcGts xmlns="http://schema.primaresearch.org/PAGE/gts/pagecontent/2017-07-15" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://schema.primaresearch.org/PAGE/gts/pagecontent/2017-07-15 http://schema.primaresearch.org/PAGE/gts/pagecontent/2017-07-15/pagecontent.xsd">
        <Page imageFileName="http://localhost:5001/00000005.tif">
        </Page>
</PcGts>
```

wrznr:
The problem could be
```python
page = OcrdPage.from_file(self.workspace.download_file(input_file))
```
line 20 in `https://github.com/OCR-D/pyocrd/blob/master/ocrd/processor/segment_line/tesserocr.py`.

kba:
Since this is now six months old, and the `process` subcommand has been outdated for a while, this is hard to reproduce right now. But the bug is still there and we should fix it once #199 is in and test it stays that way.

PR #199, opened 2018-10-25
fix CLI `ocrd process`

bertsky:
- renamed argument for METS file inherited from general CLI
- iterate through tool json tools dict (not list), taking 'executable' (not 'binary')
- enforce tool json as path name, add help

### DIFF
--- a/ocrd/cli/process.py
+++ b/ocrd/cli/process.py
@@ -16,32 +16,34 @@ from ocrd.decorators import ocrd_cli_options
 @click.argument('steps', nargs=-1)
 def process_cli(mets, steps, **kwargs):
     """
-    Execute OCR-D processors for a METS file directly.
-    
+    Execute OCR-D processor executables for a METS file.
+
     Run each of the STEPS executables one after another.
+
+    Order of comma-separated values for -I, -O, -p corresponds to order of steps.
     """
     resolver = Resolver()
     workspace = resolver.workspace_from_url(mets)
-    
+
     cmds = []
     for ocrd_tool_file in kwargs['ocrd_tool']:
         with codecs.open(ocrd_tool_file, encoding='utf-8') as f:
             obj = json.loads(f.read())
             for tool in obj['tools'].values():
                 cmds.append(tool['executable'])
-    
+
     for cmd in steps:
         if cmd not in cmds:
             raise Exception("Tool not registered: '%s'" % cmd)
-    
+
     # fixme: this really should be coming from workflow engine (giving steps, i/o fileGrps, parameter files)
     # fixme: using the command line here is inefficient
     # fixme: someone should go and check if the first input_file_grp actually exists
-    if not 'input_file_grp' in kwargs:
+    if 'input_file_grp' not in kwargs:
         raise Exception("Need input_file_grp USE(s)")
-    if not 'output_file_grp' in kwargs:
+    if 'output_file_grp' not in kwargs:
         raise Exception("Need output_file_grp USE(s)")
-    if not 'parameter' in kwargs:
+    if 'parameter' not in kwargs:
         raise Exception("Need parameter file(s)")
     input_file_grps = kwargs['input_file_grp'].split(',')
     if len(input_file_grps) != len(steps):
@@ -52,16 +54,16 @@ def process_cli(mets, steps, **kwargs):
     parameters = kwargs['parameter'].split(',')
     if len(parameters) != len(steps):
         raise Exception("Number of parameter files (%d) does not match number of steps (%d)" % (len(parameters), len(steps)))
-    
+
     for cmd, input_file_grp, output_file_grp, parameter in zip(steps, input_file_grps, output_file_grps, parameters):
-        run_cli(cmd, mets, resolver, workspace, 
-                log_level=kwargs['log_level'], 
-                group_id=kwargs['group_id'], 
-                input_file_grp=input_file_grp, 
-                output_file_grp=output_file_grp, 
+        run_cli(cmd, mets, resolver, workspace,
+                log_level=kwargs['log_level'],
+                group_id=kwargs['group_id'],
+                input_file_grp=input_file_grp,
+                output_file_grp=output_file_grp,
                 parameter=parameter)
-    
+
     workspace.reload_mets()
-    
+
     #  print('\n'.join(k + '=' + str(kwargs[k]) for k in kwargs))
     print(workspace)
--- a/ocrd/cli/process.py
+++ b/ocrd/cli/process.py
@@ -14,7 +14,7 @@ from ocrd.decorators import ocrd_cli_options
 @ocrd_cli_options
 @click.option('-T', '--ocrd-tool', multiple=True, type=click.Path(exists=True, dir_okay=False), help='register all executables from a ocrd-tool.json file')
 @click.argument('steps', nargs=-1)
-def process_cli(mets, **kwargs):
+def process_cli(mets, steps, **kwargs):
     """
     Execute OCR-D processors for a METS file directly.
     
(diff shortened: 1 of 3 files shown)


ocrd/cli/process.py
```diff
@@ -12,30 +12,56 @@
...
+    if len(input_file_grps) != len(steps):
+        raise Exception("Number of input_file_grp arguments (%d) does not match number of steps (%d)" % (len(input_file_grps), len(steps)))
+    output_file_grps = kwargs['output_file_grp'].split(',')
+    if len(output_file_grps) != len(steps):
+        raise Exception("Number of output_file_grp arguments (%d) does not match number of steps (%d)" % (len(output_file_grps), len(steps)))
+    parameters = kwargs['parameter'].split(',')
+    if len(parameters) != len(steps):
+        raise Exception("Number of parameter files (%d) does not match number of steps (%d)" % (len(parameters), len(steps)))
+    
+    for cmd, input_file_grp, output_file_grp, parameter in zip(steps, input_file_grps, output_file_grps, parameters):
```
kba: Clever.

ocrd/processor/base.py
```diff
@@ -90,12 +89,21 @@ def run_cli(
...
     """
     workspace = _get_workspace(workspace, resolver, mets_url, working_dir)
-    log.debug("Running binary '%s'", binary)
-    subprocess.call([
-        binary,
-        '--mets', mets_url,
-        '--working-dir', workspace.directory
-    ])
+    args = [binary, '--working-dir', workspace.directory]
+    if mets_url:
```
kba: required.

ocrd/cli/process.py
```diff
@@ -12,30 +12,56 @@
...
+    for cmd in steps:
         if cmd not in cmds:
             raise Exception("Tool not registered: '%s'" % cmd)
-
-    for cmd in kwargs['steps']:
-        run_cli(cmd, mets_url, resolver, workspace)
-
+    
+    # fixme: this really should be coming from workflow engine (giving steps, i/o fileGrps, parameter files)
+    # fixme: using the command line here is inefficient
```
kba: You mean "as opposed to running python code in the same thread"? Or that the way the processes are spawned?

ocrd/cli/process.py
```diff
@@ -12,30 +12,56 @@
...
+    # fixme: someone should go and check if the first input_file_grp actually exists
```
kba: Better yet: Check input requirements before each step and process success&output file group existence after every step.

bertsky:
> Zipping arguments via comma-separated list position is brittle, e.g. what to do if a processor takes multiple input groups.

Oh, right. I forgot about that. 

> Input/Output groups are partly redundant because this is a straight forward pipeline.

Er, but that's not true in the above case anymore, is it? If processors can take multiple inputs, saying which becomes which would be necessary.

> The "register ocrd-tool.json" mechanism is not necesary anymore

I had hoped so. My intention was to be as conservative as I can to get it running "somehow". 

BTW, shouldn't we attempt to solve #46 now too? (I don't know how this used to break, but I believe it does not, anymore, because each step CLI saves to METS again, and local files will stay local.)

> Ideally I would aim for a syntax/DSL resembling the pipeline

Interesting. That would straighten out the non-linear problem and make it much more readable for complex workflows. But since we are in full control of the syntax anyway, maybe one single string argument (that can also be read from a file) would be more practical?

ocrd/cli/process.py
bertsky: BTW, I also find it cumbersome to always have to provide an empty parameter json file for steps that don't take or need parameters. I cannot even provide an empty file, because the json parser wants `{}`. But from the looks of your new syntax you already thought of that.

ocrd/cli/process.py
bertsky: The latter. Going to CLI also prevents running data-parallel I think. Couldn't we look up the processor class and call `run_processor`?

ocrd/cli/process.py
bertsky: Yes. At the moment there is no error if a processor does not find any data. But that should be mocked IMO, and before the whole pipeline starts. Otherwise the user may need to undo the parts that were successful.

bertsky:
httpie is nice indeed. But this is supposed to be a feature for quick use by humans, right? IMO we need very limited functionality here, no data types, no HTTP. The best metaphor would still be the Unix command shell. 

So how about (option A):
```sh
ocrd process --group-id=OCR-D-IMG_0001 "preprocess1 <INPUT >GRP1 | process2 path/to/params.json <GRP1,OTHERGROUP >GRP2 | postprocess3 <GRP2 >OUTPUT"
```
In other words: borrow shell i/o fd redirector syntax for `fileGrp USE` (but with comma extension), borrow pipeline syntax as command separator (but without the usual redirection semantics), use one optional argument for parameter files (relative to PWD). These are the only CLI arguments we need to pass individually.

Alternatively (staying closer to the shell syntax, option B):
```sh
ocrd process --group-id=OCR-D-IMG_0001 "preprocess1 -I INPUT -O GRP1" "process2 -p path/to/params.json -I GRP1,OTHERGROUP -O GRP2" "postprocess3 -I GRP2 -O OUTPUT"
```
In other words: directly pass arguments to processor CLIs, and directly use shell expression syntax (grouping strings via quotes) as command separator.

Note that option B does not preclude storing configurations in strings and files as well, because one could always have them expanded separately like so:
```sh
config=("preprocess1 -I INPUT -O GRP1" "process2 -p path/to/params.json -I GRP1,OTHERGROUP -O GRP2" "postprocess3 -I GRP2 -O OUTPUT")
# or file-based (one line per processor):
readarray -t config <best-workflow.cfg
ocrd process "${config[@]}"
```

kba:
Option A would be more fun to implement but option B is more intuitive and easier to parse since we already have library code for that and stay consistent (e.g. `-I` means "input file group" in all CLI).

Having the individual calls as a single string has the disadvantage of disabling autocompletion in the shell though but that is minor and is only relevant for parameters anyway (and would also apply to the httpie-`@`-convention (unless we generate autocompletion scripts for bash/zsh which would be overkill but, again, great fun 😄)).

So option B is fine by me. @VolkerHartmann @wrznr?

VolkerHartmann:
I also prefer Option B because it is the more intuitive method and code reuse is always better.

bertsky:
> disadvantage of disabling autocompletion in the shell

Yes, inevitably. But as with the `-I` or `-L` option to gcc, one can always type and autocomplete the path string first and then fit the actual command syntax afterwards. This is very simple for both A and B: one merely needs to add the quotes as final editing step. Jumping over words (`M-f` / `M-b` with standard=emacs readline bindings) or over shell expressions makes that easier, of course. I prefer the latter, but these are unset by default. Being used to jump over s-expressions in Emacs via `C-M-f` / `C-M-b`, I configured them like this in my .inputrc:
```
$if mode=emacs
    $if Bash
    set convert-meta on
    Meta-C-f: shell-forward-word
    Meta-C-b: shell-backward-word
    set convert-meta off # required for UTF-8 locales
    $endif
$endif
```

ocrd/cli/process.py
```diff
@@ -16,32 +16,34 @@
...
+    Execute OCR-D processor executables for a METS file.
+
     Run each of the STEPS executables one after another.
+
+    Order of comma-separated values for -I, -O, -p corresponds to order of steps.
     """
     resolver = Resolver()
     workspace = resolver.workspace_from_url(mets)
-    
+
```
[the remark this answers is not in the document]
kba: > `no-space-check=empty-line`

Fine by me. No point in wasting time because of conflicting matters of taste about insignificant whitespace.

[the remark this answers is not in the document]
kba: 94bfa1e

kba:
OK, option B it is. I'll merge this now and send a PR for it soon.

PR #222, opened 2018-12-11
Group to pageid

kba:
Blocked by #221


ocrd/model/ocrd_file.py
```diff
@@ -57,14 +58,19 @@ def ID(self, ID):
...
     @property
-    def groupId(self):
-        return self._el.get('GROUPID')
-
-    @groupId.setter
-    def groupId(self, groupId):
-        if groupId is None:
+    def pageId(self):
+        if self.mets is None:
+            raise Exception("OcrdFile %s has no member mets pointing to parent OcrdMets" % self)
```
wrznr: Pls. quote (or hoghlight otherwise) `mets`.

ocrd/model/ocrd_file.py
```diff
@@ -57,14 +58,19 @@ def ID(self, ID):
...
+        return self.mets.get_physical_page_for_file(self)
+
+    @pageId.setter
+    def pageId(self, pageId):
+        if pageId is None:
             return
-        self._el.set('GROUPID', groupId)
+        if self.mets is None:
+            raise Exception("OcrdFile %s has no member mets pointing to parent OcrdMets" % self)
```
wrznr: See above.

ocrd/model/ocrd_mets.py
```diff
@@ -99,32 +103,44 @@ def find_files(self, ID=None, fileGrp=None, groupId=None, mimetype=None, local_o
...
+        # instantiate / get from cache
+        for file_id in file_ids:
+            el = self._tree.getroot().find('.//mets:file[@ID="%s"]' % file_id, NS)
             if file_id not in self._file_by_id:
-                self._file_by_id[file_id] = OcrdFile(el, baseurl=self.baseurl)
+                self._file_by_id[file_id] = OcrdFile(el, baseurl=self.baseurl, mets=self)
             if local_only:
                 url = el.find('mets:FLocat', NS).get('{%s}href' % NS['xlink'])
-                if not url.startswith('file://'):
+                if not (url.startswith('file://') or '://' not in url):
```
wrznr: First condition is a special case of the second and thus superfluous.

kba: The condition checks whether the URL is either a `file://` URL or not a URL at all (i.e. a relative path). If it is neither, skip the file.

I'll add a comment to make that clearer.

PR #223, opened 2018-12-17
page validator to check/fix consistency, spec#82

kba:


ocrd/validator/page_validator.py
```diff
@@ -0,0 +1,174 @@
...
+        self.tag = tag
+        self.ID = ID
+        self.actual = actual
+        self.expected = expected
+        super(ConsistencyError, self).__init__("INCONSISTENCY in %s ID '%s': text results '%s' != concatenated '%s'" % (tag, ID, actual, expected))
+
+class PageValidator(object):
+
+    @staticmethod
+    def validate_filename(filename, **kwargs):
```
wrznr: I do not like the naming scheme here: `validate_filename` is misleading especially given its counterpart `validate_ocrd_page`.

ocrd/validator/page_validator.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Returns:
+            report (:class:`ValidationReport`) Report on the validity
+        """
+        if self.strictness == 'off':
+            return self.report
+        self.find_or_fix_inconsistencies(self.page, strictness=self.strictness)
+        return self.report
+
+    def concatenate_index_1_text(self, node, concatenate_with):
```
wrznr: Not sure whether it makes much sense in real life but `index` (c|sh)ould be parameter.

ocrd/validator/page_validator.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            if len(textEquivs) > 1:
+                index1 = [x for x in textEquivs if x.index == 1]
+                if index1:
+                    index1[0].set_Unicode(text)
+                    return
+            textEquivs[0].set_Unicode(text)
+
+
+    def find_or_fix_inconsistencies(self, node, strictness='strict'):
```
wrznr: Split into two methods? `find_inconsistencies` returns a list of inconsistencies which can be fixed via `fix_inconsistencies`.

ocrd/validator/page_validator.py
kba: I don't think it would ever make sense to have a "concatenate by index 2" method. But it **would** make sense to have a "concatenate most confident" method. So maybe have the signature

```
def concatenate_text(self, node, strategy='index1', concatenate_with=' '):
  if strategy not in ('index1', 'conf'):
    raise Exception("No such strategy %s" % strategy)
```

ocrd/validator/page_validator.py
kba: It's a bit smelly that `report` is an instance property. Unfortunately that means that the whole instance has a state. Before implementing this in a PageValidator, I had the methods as functions in `ocrd_page`, where the behaviour you propose was easier.

I could add facades `find_inconsistencies` and `fix_inconsistencies` that only differ in what they accept as strictness?

Or just prepend an underscore and to make it clear that these methods are not part of the public API.

ocrd/validator/page_validator.py
kba: The naming scheme is `validate_<source-of-the-page-xml>`. `validate_ocrd_page` means that the source is an `OcrdPage` object, `validate_filename` means PAGE XML comes from a file on disk.

What would you prefer? `validate_by_filename` or similar?

ocrd/validator/page_validator.py
wrznr: The latter.

In understand that the distinction between `fix` and `find` depends on the `strictness` parameter, right? Maybe `handle_inconsistencies` could be an alternative?

ocrd/validator/page_validator.py
wrznr: YES!

ocrd/validator/page_validator.py
wrznr: Difficult. Is it technically possible to have a `validate` method which calls the appropriate method depending on the parameter type?

ocrd/validator/page_validator.py
kba: You mean `handle_inconsistencies` is clearer than `find_or_fix_inconsistencies`?

ocrd/validator/page_validator.py
kba: Now checked in constructor

ocrd/validator/page_validator.py
kba: called handle_inconsistencies now and again a function not a method (as are the former `*_index_1_*` methods).

PR #241, opened 2019-05-17
ocrd_page: Additional region type ImageRegion

kba:


bertsky:
What is the reason for this – isn't it enough to have `ImageRegionType` in ocrd_page_generateds?

kba:
> What is the reason for this – isn't it enough to have ImageRegionType in ocrd_page_generateds?

Yeah, that is possible. We just thought it would be best to encapsulate access to the generated code.

PR #257, opened 2019-07-11
Improve pixel density logic in OcrdExif

bertsky:
OCR-D#256, OCR-D#37, OCR-D/ocrd_tesserocr#54

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Changed:
+
+  * Improve pixel density logic in OcrdExif, #256, #37, OCR-D/ocrd_tesserocr#54
+
 ## [1.0.0b10] - 2019-06-25
 
 Fixed:
--- a/ocrd_models/ocrd_models/ocrd_exif.py
+++ b/ocrd_models/ocrd_models/ocrd_exif.py
@@ -2,9 +2,22 @@
 Technical image metadata
 """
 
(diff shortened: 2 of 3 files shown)


bertsky:
Looking into Travis failure now... (I get `IFDRational` instead of integer for `xResolution`, which is not allowed in the multiplication, at least in Python 3.5)

bertsky:
It's strange that [IFDRational](https://github.com/python-pillow/Pillow/blob/1c57a41ececc4529520a423b9bec13c78927cb25/src/PIL/TiffImagePlugin.py#L295) would not be an allowed operand type for `*` in Python 3.5 and Python 3.7 ...

bertsky:
The Python 3.7-dev test seems to broken already.

kba:
Yeah, the 3.7-dev failure is unrelated, though strange.

```
pytest.py: error: unrecognized arguments: --duration=10
```

Version problem with the tooling it seems.

bertsky:
Python 3.4 has been removed from support in Pillow >= 6.0:
> Drop support for Python 3.4 #3596 [hugovk]

But we get the issue with 5.4.1 according to Travis' log

bertsky:
No GH issues for the `IFDRational` problem in Pillow yet. I will try the Pillow selftest in a 3.4 installation...

kba:
Or we just stop supporting 3.4. IIRC the reason we did support it was because of CentOS still shipping it a year ago. Development stopped, there will be no more 3.4 releases. @cneud @wrznr what do you think?

PR #261, opened 2019-07-16
:fire: stop supporting python <= 3.4

kba:


kba:
This works but why `--duration=10` would not be recognized by pytest.py still puzzles me. Happened for Python 3.7-dev on Travis but for Python 3.5, 3.6 and 3.7 on CircleCI. Problem "solved" by removing the parameter (which is purely aesthetic anyway, but still...).

PR #268, opened 2019-08-01
Refactoring of ocrd_tesserocr common functionality into core

kba:
Start implementing https://github.com/OCR-D/ocrd_tesserocr/issues/49


ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -5,29 +5,48 @@
...
-points a string encoding a polygon: "0,0 100,0 100,100, 0,100"
-polygon an array of x-y-tuples of a polygon: [[0,0], [100,0], [100,100], [0,100]]
-xywh a dict with keys for x, y, width and height: {'x': 0, 'y': 0, 'w': 100, 'h': 100}
-points is for PAGE
+* bbox is a 4-tuple of x0, y0, x1, y1 of the bounding box
+* points a string encoding a polygon: "0,0 100,0 100,100, 0,100". points is for PAGE
+* polygon an array of x-y-tuples of a polygon: [[0,0], [100,0], [100,100], [0,100]]
+* xywh a dict with keys for x, y, width and height: {'x': 0, 'y': 0, 'w': 100, 'h': 100}
 
 polygon is what opencv2 expects
```
bertsky: Please add our own coordinate provisions (`coordinates_of_segment`, `coordinates_for_segment`, `polygon_mask`, `image_from_polygon`, `rotate_coordinates`) as use case for `polygon` here.
```suggestion
polygon is what opencv2 and higher-level coordinate functions in ocrd_utils expect
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -5,29 +5,48 @@
```
bertsky: Please add more description items __above__ (being more abstract than the converter functions) the first item on line 4.
```suggestion
* coordinates_of_segment, coordinates_for_segment, rotate_coordinates

These functions convert polygon outlines for PAGE elements on all hierarchy levels (page, region, line, word, glyph) between relative coordinates w.r.t. parent segment and absolute coordinates w.r.t. the top-level (source) image. This includes rotation and offset correction.

* polygon_mask, image_from_polygon, crop_image

These functions combine PIL.Image with polygons or bboxes.
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -5,29 +5,48 @@
 
 The functions have the syntax X_from_Y, where X/Y can be
 
-points a string encoding a polygon: "0,0 100,0 100,100, 0,100"
-polygon an array of x-y-tuples of a polygon: [[0,0], [100,0], [100,100], [0,100]]
-xywh a dict with keys for x, y, width and height: {'x': 0, 'y': 0, 'w': 100, 'h': 100}
-points is for PAGE
+* bbox is a 4-tuple of x0, y0, x1, y1 of the bounding box
+* points a string encoding a polygon: "0,0 100,0 100,100, 0,100". points is for PAGE
```
bertsky:
```suggestion
* points a string encoding a polygon: "0,0 100,0 100,100, 0,100"
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -5,29 +5,48 @@
...
+* polygon an array of x-y-tuples of a polygon: [[0,0], [100,0], [100,100], [0,100]]
+* xywh a dict with keys for x, y, width and height: {'x': 0, 'y': 0, 'w': 100, 'h': 100}
 
 polygon is what opencv2 expects
 
 xywh is what tesserocr expects/produces.
```
bertsky: Tesseract also expects/gives x0y0x1y1 in some places (like `BoundingBox` on a page iterator). 
```suggestion
xywh and x0y0x1y1 are what tesserocr expects/produces.
points is what PAGE-XML uses.
bbox is what PIL.Image uses.
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -5,29 +5,48 @@
...
 """
```
bertsky: Please mention the remaining functions in the top-level docstring.
```suggestion
* is_local_filename, safe_filename, abspath

* is_string, membername, concat_padded

* MIMETYPE_PAGE, EXT_TO_MIME, VERSION
"""
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -82,6 +207,19 @@ def is_string(val):
...
+    return "%i,%i %i,%i %i,%i %i,%i" % (
+        minx, miny, maxx, miny, maxx, maxy, minx, maxy)
+
+def points_from_polygon(polygon):
+    """Convert polygon coordinates from a numeric list representation to a page representation."""
+    return " ".join("%i,%i" % (x, y) for x, y in polygon)
+
 def points_from_xywh(box):
     """
     Constructs a polygon representation from a rectangle described as a dict with keys x, y, w, h.
```
bertsky: Please use the same system for describing the converter docstrings in the old functions as well (making them more precise and consistent with the new ones).
```suggestion
    Construct polygon coordinates in page representation from numeric dict representing a bounding box.
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -125,6 +263,10 @@ def points_from_x0y0x1y1(xyxy):
         x0, y1
     )
 
+def polygon_from_bbox(minx, miny, maxx, maxy):
+    """Construct polygon coordinates in numeric list representation from a numeric list representing a bounding box."""
+    return [[minx, miny], [maxx, miny], [maxx, maxy], [minx, maxy]]
+
 def polygon_from_points(points):
     """
     Constructs a numpy-compatible polygon from a page representation.
```
bertsky: Same here.
```suggestion
    Convert polygon coordinates in page representation to polygon coordinates in numeric list representation.
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -151,6 +340,15 @@ def unzip_file_to_dir(path_to_zip, output_directory):
...
+    return {
+        'x': minx,
+        'y': miny,
+        'w': maxx - minx,
+        'h': maxy - miny,
+    }
+
 def xywh_from_points(points):
     """
     Constructs an dict representing a rectangle with keys x, y, w, h
```
bertsky: And here.
```suggestion
    Construct a numeric dict representing a bounding box from polygon coordinates in page representation.
```

ocrd_utils/requirements.txt
```diff
@@ -0,0 +1,2 @@
+Pillow >= 5.3.0
```
bertsky: Maybe this needs to be `>= 6.1.0` – see #270.

bertsky:
Wait a minute! Where is `coordinates_for_segment`?

When you add this, please be sure to apply OCR-D/ocrd_tesserocr#68 as well!

ocrd_utils/requirements.txt
kba: I'll debug that separately.

kba:
> Wait a minute! Where is coordinates_for_segment?

It's there now. The rest of common.py has been turned into workspace methods.

More tests would be wise but the ocrd_tesserocr test suite passes.

ocrd_utils/ocrd_utils/__init__.py
bertsky: GH does not allow me to make suggestions outside the 3 lines of context within existing hunks. Please read my original comment: this docstring is meant as the _first item_ on __line 4__. 

Also, GH does not offer a proper interface to calculate suggestions from the result of proper editing, so my suggestions were necessarily incomplete. Please have a look at the final result – I see missing indentation and quotation, suboptimal newlines etc.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Added:
+
+  * many utility methods for image manipulation and coordinate handling, #268
```
bertsky: Could we advertise this some more? After all, this is supposed to become the _recommended_ or even _required_ way of processing images on a workspace (when working with core). This should at least mention that it offers spec-conforming `AlternativeImage` processing. Also, we should refer to OCR-D/spec#116 (regarding the details that are still missing in the spec but already assumed here).

ocrd/ocrd/workspace.py
```diff
@@ -194,3 +205,184 @@ def resolve_image_as_pil(self, image_url, coords=None):
             np.min(poly[:, 0]):np.max(poly[:, 0])
         ]
         return Image.fromarray(region_cut)
+
+    def image_from_page(self, page, page_id):
+        """Extract the Page image from the workspace.
+
+        Given a PageType object, `page`, extract its PIL.Image from
+        AlternativeImage if it exists. Otherwise extract the PIL.Image
+        from imageFilename and crop it if a Border exists. Otherwise
```
bertsky: @wrznr is right about insisting that the term `cropping` (de: `Beschneidung`) should only apply to the process of finding the `Border` (and perhaps also removing the margins from the image by cutting), not of other elements down the hierarchy. This we should rather call `cutting` (de: `Freistellen`) – it is only due to `PIL.Image.crop` that I was led astray.

If this is correct, then the docstrings must be fixed accordingly throughout.

ocrd/ocrd/workspace.py
```diff
@@ -6,12 +7,22 @@
...
+    is_local_filename,
+    polygon_from_points,
+    xywh_from_points,
+)
 
 from .workspace_backup import WorkspaceBackupManager
 
 log = getLogger('ocrd.workspace')
 
+
```
bertsky: Following up on the above comment on advertising the new functions, could we change the docstring of `resolve_image_as_pil` to point to them (and perhaps even warning of deprecation)?

bertsky:
> More tests would be wise but the ocrd_tesserocr test suite passes.

Yes, `ocrd_tesserocr` is very much in need of more test coverage now, and here `image_from_page`, `image_from_segment` and `save_image_file` should be controlled by tests. As for the latter, I believe we would need some real-life workspaces (as in assets) to cover this.

ocrd/ocrd/workspace.py
kba: Not a huge fan because I find the distinction unintuitive but I see the reasoning. Should `crop_image` then be renamed `cut_image` as well since image isn't necessarily a page?

ocrd/ocrd/workspace.py
kba: See 50f7621

ocrd/ocrd/workspace.py
```diff
@@ -247,12 +247,12 @@ def image_from_page(self, page, page_id):
 
         Given a PageType object, `page`, extract its PIL.Image from
         AlternativeImage if it exists. Otherwise extract the PIL.Image
-        from imageFilename and crop it if a Border exists. Otherwise
+        from imageFilename and cut it if a Border exists. Otherwise
         just return it.
 
-        When cropping, respect any orientation angle annotated for
+        When cuting, respect any orientation angle annotated for
```
bertsky:
```suggestion
        When cutting, respect any orientation angle annotated for
```

ocrd/ocrd/workspace.py
```diff
@@ -281,8 +281,8 @@ def image_from_page(self, page, page_id):
 
         alternative_image = page.get_AlternativeImage()
         if alternative_image:
-            # (e.g. from page-level cropping, binarization, deskewing or despeckling)
-            # assumes implicit cropping (i.e. page_xywh has been applied already)
+            # (e.g. from page-level cutting, binarization, deskewing or despeckling)
```
bertsky: This actually refers to `cropping` the preprocessing step, so
```suggestion
            # (e.g. from page-level cropping, binarization, deskewing or despeckling)
```

ocrd/ocrd/workspace.py
```diff
@@ -281,8 +281,8 @@ def image_from_page(self, page, page_id):
...
+            # assumes implicit cutting (i.e. page_xywh has been applied already)
```
bertsky: This actually refers to `cropping` the preprocessing step, so
```suggestion
            # assumes implicit cropping (i.e. page_xywh has been applied already)
```

ocrd/ocrd/workspace.py
```diff
@@ -293,7 +293,7 @@ def image_from_page(self, page, page_id):
             page_polygon = np.array(polygon_from_points(page_points))
             # create a mask from the page polygon:
             page_image = image_from_polygon(page_image, page_polygon)
-            # recrop into page rectangle:
+            # crop into page rectangle:
```
bertsky:
```suggestion
            # cut into page rectangle:
```

ocrd/ocrd/workspace.py
```diff
@@ -359,7 +359,7 @@ def image_from_segment(self, segment, parent_image, parent_xywh):
             segment_polygon = coordinates_of_segment(segment, parent_image, parent_xywh)
             # create a mask from the segment polygon:
             segment_image = image_from_polygon(parent_image, segment_polygon)
-            # recrop into segment rectangle:
+            # crop into segment rectangle:
```
bertsky:
```suggestion
            # cut into segment rectangle:
```

ocrd/ocrd/workspace.py
bertsky: That would be more consequential, but I tend to say no: `crop_image` is meant as replacement for `Image.crop` and should be memorable. I am becoming less enthusiastic about this terminological distinction by the minute... maybe this should be reverted (sorry).

wrznr:
Wrt. *cropping* vs. *cutting* (vs. *segmenting*?): Using the term *cropping* for localizing a page's border was a bad choice right from the start because it mixes the intellectual process of finding the borders and the physical process of separating the OCR-relevant from the irrelevant parts of the actual image. Using *cutting* does not improve things IMHO. The more I think about it, the more the more meaningful the use of the term (page-level) *segmentation* seems to me because this is what cropping right now does: It localizes the segment **page** on an image file. We could then use *cropping* as it is intended.

kba:
> We should not let the terminological discussion slow us down.

I second that, let's discuss in https://github.com/OCR-D/core/issues/289.

ocrd_utils/ocrd_utils/__init__.py
bertsky: @kba What of it now – you already merged, despite these unresolved – should I make a new PR with the docstrings fixed?

kba:
> @kba Do you want me to make the repairs myself and commit here?

I'll fix it right away, thanks for spotting.

ocrd_utils/ocrd_utils/__init__.py
kba: > @kba What of it now – you already merged, despite these unresolved – should I make a new PR with the docstrings fixed?

I thought I had resolved the issues in 677890b and ac95d38. Can you make a PR against master with the fixed docstrings or at least a WIP PR which shows what you want changed? Thanks!

ocrd/ocrd/workspace.py
kba: cf #289

ocrd_utils/ocrd_utils/__init__.py
bertsky: Yes, I will do that.

PR #266, opened 2019-07-19
resolver.workspace_from_url: cover relative paths, too

bertsky:
This (at least partially) fixes #96, i.e. `workspace_from_url` failing to download files that had relative paths. The reason was that `workspace.download_file` does not know anything about `baseurl`, i.e. it cannot concatenate the source path before copying locally. In contrast, `download_url` does wrap this case correctly.

Issue #227, opened 2019-01-04
Avoid superfluous file copies in workspace directory


wrznr:
Using https://raw.githubusercontent.com/OCR-D/assets/master/data/dietrich_fuehrer_1839/data/mets.xml, each processing steps copies the input files from the corresponding subdirectory into the workspace directory (i.e. files from the output directory `KRAKEN-BIN` are copied to `.` when the `KRAKEN-BIN` filegroup is used as input for e.g. region detection).

wrznr:
This can be closed, right?

mikegerber:
Seems to be fixed, using these versions:

ocrd (1.0.0b10)
ocrd-kraken (0.1.0)
ocrd-modelfactory (1.0.0b10)
ocrd-models (1.0.0b10)
ocrd-tesserocr (0.2.2)
ocrd-typegroups-classifier (0.0.1)
ocrd-utils (1.0.0b10)
ocrd-validators (1.0.0b10)

mikegerber:
It still happens with ocrd-ocropy-segment 0.0.3:

~~~
#!/bin/bash -x
cd `mktemp -d`

virtualenv -p /usr/bin/python3 venv
. venv/bin/activate
pip install ocrd-ocropy

wget https://ocr-d-repo.scc.kit.edu/api/v1/dataresources/1b0fb8b5-3397-4c99-9ec3-24d5954ac0fb/data/bernd_lebensbeschreibung_1738.ocrd.zip
dtrx bernd_lebensbeschreibung_1738.ocrd.zip

cd bernd_lebensbeschreibung_1738.ocrd/data
ocrd-ocropy-segment -m mets.xml -I OCR-D-IMG -O OCR-D-SEG-LINES

pip list | grep ocrd
ls -l
~~~

yields:

~~~
[...]
ocrd                   1.0.0b10 
ocrd-fork-ocropy       1.4.0a3  
ocrd-modelfactory      1.0.0b10 
ocrd-models            1.0.0b10 
ocrd-ocropy            0.0.3    
ocrd-utils             1.0.0b10 
ocrd-validators        1.0.0b10 
total 23348
-rw-rw-r--. 1 mike mike   13069 Jul  3 14:37 mets.xml
drwxrwxr-x. 2 mike mike     100 Feb 26 15:21 OCR-D-GT-SEG-BLOCK
drwxrwxr-x. 2 mike mike     100 Feb 26 15:21 OCR-D-GT-SEG-LINE
[...]

Issue #96, opened 2018-06-12
ocrd workspace clone -a -m mets.xml will not adapt links for files


VolkerHartmann:
After cloning mets.xml looks the same than before.
Although the contained image was copied to the new directory.

bertsky:
I am not sure if this is the same problem, but there still is a problem with `workspace clone`: 

```sh
cd assets/data
ocrd workspace clone -l kant_aufklaerung_1784/data/mets.xml ws1
```

After this `ws1/` will contain nothing but an `mets.xml` – pointing (as `LOCTYPE="URL"` with `file://` hrefs) not to the original directory, but to new local directories under ws1, which do not exist.

Am I doing something wrong or is that behaviour broken?

bertsky:
Also, by the [new rules for METS file URLs](https://ocr-d.github.io/mets#always-use-url-or-relative-filenames), I think core must now not produce absolute file paths at all when cloning.

bertsky:
I think at least one problem is in `Resolver.workspace_from_url`: If the source METS file uses relative paths, as is now allowed (and even required if local), then it will not be enough to just dump it in the new directory and instantiate a `Workspace` with `src_dir` set appropriately – because nothing will preserve that information without `download=True` or `download_local=True`. The new METS must always rebase its local file paths.

bertsky:
Please reopen.

kba:
The original bug was what is now https://github.com/OCR-D/core/issues/227, what  you describe is indeded a regression, thanks for investigating.

kba:
The `-l` option is gone now, but with #232

```shell
ocrd workspace clone -a kant_aufklaerung_1784/data/mets.xml ws1
```

should work as expected (files are copied).

bertsky:
I tried that with 1.0.0b5, but there I get a `FileNotFoundError` on the very first source file, which does exist in the filesystem. By a first glance, could it be that [the chdir](https://github.com/OCR-D/core/blob/35d26e421309cb17ed61ff3563692346e516338d/ocrd/ocrd/workspace.py#L81) in `download_file` is wrong, or at least not compensated for when entering `download_url`?

bertsky:
It also applies to all processors or tests that use `resolver.workspace_from_url` (with `download=True`) or `workspace.download_file` – when the input has relative paths, as do bags.

I think the cause is changeset 76e02944, which seems to be the wrong way of fixing #233.

bertsky:
[...]

PR #275, opened 2019-08-08
Delete files

kba:
Adds a command `ocrd workspace remove` that allows removing files, with the resp. `remove_files` commands on workspace and METS level.

Also a few fixes related to handling of local filenames for mets:file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Added:
+
+  * `ocrd workspace remove` to remove files, #275
+
 ## [1.0.0b11] - 2019-08-08
 
 Fixed:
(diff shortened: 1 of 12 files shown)


kba:
> b027faa: Is this related to #96 and my fix #266 by any chance?

Yes, it's related. The proper fix is to harmonize the URL handling throughout, b027faa is yet another band-aid I'm afraid.

bertsky:
> Yes, it's related. The proper fix is to harmonize the URL handling throughout, [b027faa](https://github.com/OCR-D/core/commit/b027faafafdfb29290435ed4f55c2c1a8cc85823) is yet another band-aid I'm afraid.

Um, does that mean b027faa is temporary? And does this mean #266 is going to be adopted?

kba:
Yeah, https://github.com/OCR-D/core/commit/b027faafafdfb29290435ed4f55c2c1a8cc85823 is temporary. I have another branch for fixing #96 and #227 that should hopefully render #266 unnecessary.

bertsky:
Understood – thanks for clarifying!

PR #284, opened 2019-08-14
improve bashlib ocrd__parse_argv:

### LINKED ISSUE #283: bashlib: mets_file param must be relative to workdir or absolute
In [lib.bash:ocrd__parse_argv](https://github.com/OCR-D/core/blob/1494e948ab9166b095e7855593df6048da4d1f68/ocrd/ocrd/lib.bash#L92), the `--mets` parameter gets passed to `ocrd_argv[mets_file]` as is. 

This does not work, if the path was relative when calling, and the processor decides to chdir (e.g. to the workdir) in between. The path should be made either relative to the working directory (i.e. stripped of any directory components) or absolute.

Moreoever, it should be allowed to omit that parameter to set the default name in the default path (i.e. `$PWD/mets.xml`). The error should only be thrown if the file does not exist or is not readable.

bertsky:
- quote all FS path parameters to prevent tokenization
- convert mets_file and working_dir to absolute paths
  (to ensure this still works after chdir; fixes #283)
- do not just set defaults, also validate values:
  - mets_file must be readable, otherwise fail
  - working_dir must be directory, otherwise fail
  - log_level must conform to spec, otherwise fail
  - input_file_grp must conform to spec, otherwise warn
  - output_file_grp must conform to spec, otherwise warn

### DIFF
--- a/ocrd/ocrd/cli/bashlib.py
+++ b/ocrd/ocrd/cli/bashlib.py
@@ -1,6 +1,10 @@
+from __future__ import print_function
+import sys
 import click
 
 from ocrd.constants import BASHLIB_FILENAME
+import ocrd_utils.constants
+import ocrd_models.constants
 
 # ----------------------------------------------------------------------
 # ocrd bashlib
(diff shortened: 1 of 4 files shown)


bertsky:
The last changeset is a proof-of-concept for exporting (constant) python data to bash. 

Functions (like logging or `OcrdExif` members) will be more tricky: for example with Click's `ignore_unknown_options` one could pass on all options/arguments to the requested function (available from a certain set).

ocrd/ocrd/cli/bashlib.py
```diff
@@ -22,3 +26,30 @@ def bashlib_filename():
     Dump the bash library filename for sourcing by shell scripts
     """
     print(BASHLIB_FILENAME)
+
+@bashlib_cli.command('constants')
+@click.argument('name')
+def bashlib_constants(name):
```
kba: Very clever, me like.

bertsky:
> LGTM Thanks! The bashlib code is generated from the sources in `ocrd/bashlib/src`, I'll adapt it.

Oh, I did not see that. Thanks.

bertsky:
Sorry, I made a typo. See https://github.com/OCR-D/core/pull/288/commits/ec39972c61009f30f89a9ba1e606b7e1e30a7e3d in (unrelated) PR #288.

PR #285, opened 2019-08-15
[WIP] atomic write of mets.xml, fix #278

### LINKED ISSUE #278: writing to METS must be atomic
I frequently get an empty METS file when interrupting (ctrl+c) a long run of `ocrd process`. This I cannot even restore with `ocrd workspace backup`. IMO the interrupt must be suspended as long as the file is open for writing.

kba:
Just a quick idea on how to do this. Should investigate where else mets.xml is opened for writing.

@bertsky


bertsky:
Thanks! This works very well for me.

Issue #254, opened 2019-06-26
Improve package docstrings


bertsky:
IMO the core packages (`ocrd`, `ocrd_models` etc) should really use package-level docstrings (i.e. top-level docstrings in `__init__.py` in each directory) to aid browsing the implementation from Python (or with a browser in the generated apidoc). Also, the main `README.md` should not call these "modules" but "packages".

kba:
Fixed by https://github.com/OCR-D/core/pull/292

PR #292, opened 2019-08-22
fix __init__ docstring order/markup

bertsky:
- re-instate item for coordinate conversion utils (not first any longer)
- improve RST indentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 <!-- BEGIN-MARKDOWN-TOC -->
 * [Introduction](#introduction)
-* [Modules](#modules)
+* [Packages](#packages)
 	* [ocrd_utils](#ocrd_utils)
 	* [ocrd_models](#ocrd_models)
 	* [ocrd_modelfactory](#ocrd_modelfactory)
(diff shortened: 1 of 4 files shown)


bertsky:
@kba would it be ok if I added more commits in accordance with #254?

bertsky:
From top-level (core) README.md:
> ### ocrd
> Contains all of the above and in addition decorators for creating OCR-D processors and CLI.

Shouldn't that read `installs all of the above`? (When importing, I still have to use the other packages, so they are not _contained_ in `ocrd`.) 

Also, shouldn't that read "decorators `and classes` for creating OCR-D processors and CLI"?

kba:
> Also, shouldn't that read "decorators and classes for creating OCR-D processors and CLI"?

Yes.

> Shouldn't that read installs all of the above? (When importing, I still have to use the other packages, so they are not contained in ocrd.)

Also true. Initially, I wanted to expose all the `ocrd_*` functionality directly from `ocrd/__init__.py` so you could do `from ocrd.utils import logging` install of `from ocrd_utils import logging` or even `from ocrd import logging`. But that got confusing and unwieldy.

kba:
> @kba would it be ok if I added more commits in accordance with #254?

Very okay that would be :) I just would not prioritize it before the dev workshop next week. But documentation is always welcome!

bertsky:
Okay, I will change that formulation in another commit then. 

And do you generally concur with `s/module/package/` in this scope?

kba:
> And do you generally concur with `s/module/package/` in this scope?

In usage, they are modules but they are distributed as separate packages. In the top-level README, "package" makes more sense, yes.

bertsky:
Exactly. I had this surprise effect when I first saw your restructuring that it's really "one repo, multiple packages". And since in Python, all packages are also modules...

bertsky:
Done. Fixes #254 – thankfully, you had already added most of the `__init__` docstrings by now.

kba:
Thanks!

PR #293, opened 2019-08-25
image_from_page: ensure to rotate image if orientation

bertsky:
Fixes a bug in which `image_from_page`: 
- returned a xywh dictionary without the 'angle' key and without rotating the image if no `AlternativeImage` was present already and if `Border` or `PrintSpace` where supplied,
- returned a xywh dictionary with the 'angle' key and without rotating the image if no `AlternativeImage` was present already and if no `Border` or `PrintSpace` where supplied

(This became relevant if page-level deskewing had been run earlier without producing `AlternativeImage`.)

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -293,18 +293,19 @@ class Workspace():
                       page_id)
             page_image = self._resolve_image_as_pil(
                 alternative_image[-1].get_filename())
-        elif border:
-            # get polygon outline of page border:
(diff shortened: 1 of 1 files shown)


bertsky:
@kba, can you please merge this quickly – I have got another patch on `AlternativeImage` waiting that should urgently be published before people start to adopt this. (This will be about monotonic `@comments` classification and the possibility to select among the derived images by operations via a mask into `@comments`.)

PR #294, opened 2019-08-28
Feature-based derived image selection/filtering

bertsky:
Based on ideas from @wrznr – thanks!!

This is really exciting, it lifts the whole thing to a new level.

The next steps could be:
- writing automated tests for this
- using these features in the processors
- moving from Exceptions to a workflow validation setting by incorporating the feature requirements into the static `ocrd-tool.json`
- modifying OCR-D/spec to reflect and stipulate this solution

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -247,25 +247,58 @@ class Workspace():
         ]
         return Image.fromarray(region_cut)
 
-    def image_from_page(self, page, page_id):
-        """Extract the Page image from the workspace.
-
-        Given a PageType object, `page`, extract its PIL.Image from
+    def image_from_page(self, page, page_id, feature_selector='', feature_filter=''):
+        """Extract a Page image from the workspace.
+        
+        Given a PageType object, ``page``, extract its PIL.Image from
         AlternativeImage if it exists. Otherwise extract the PIL.Image
-        from imageFilename and crop it if a Border exists. Otherwise
-        just return it.
+        from imageFilename, and crop it if a Border exists, and rotate
+        it if an orientation angle exists. Otherwise just return it.
+        
+        When an AlternativeImage exists, if ``feature_selector`` and/or
+        ``feature_filter`` is given, select/filter among the available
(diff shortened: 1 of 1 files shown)


bertsky:
@kba as you can see, the first 2 commits are already in #293.

ocrd/ocrd/workspace.py
```diff
@@ -247,150 +247,270 @@ def _resolve_image_as_pil(self, image_url, coords=None):
...
-        the page's bounding box / border (for passing down), and
-        an OcrdExif instance associated with the original image.
+        ``page``, then in the returned bounding box, reduce the offset by
+        half the width/height difference (so consumers being passed this
+        image and offset will still crop relative to the original center).
+        
+        Return a tuple:
+         * the extracted image,
+         * a dictionary with the absolute coordinates of the page's
+           bounding box / border (xywh), angle and the AlternativeImage
```
wrznr: AlternativeImage → AlternativeImage's ???

ocrd/ocrd/workspace.py
```diff
@@ -247,150 +247,270 @@ def _resolve_image_as_pil(self, image_url, coords=None):
...
-        Return the extracted image, and the absolute coordinates of
-        the segment's bounding box (for passing down).
+        ``segment``, then in the returned bounding box, reduce the offset by
+        half the width/height difference (so consumers being passed this
+        image and offset will still crop relative to the original center).
+        
+        Return a tuple:
+         * the extracted image,
+         * a dictionary with the absolute coordinates of the segment's
+           bounding box (xywh), angle and the AlternativeImage @comments
```
wrznr: See above.

ocrd/ocrd/workspace.py
bertsky: I deliberately avoided the possessive here, because that would carry an ambiguity: is it about the _input_ `AlternativeImage` or the _output_ image that can/should/will be added based on the returned image? The returned value actually is the latter, but with a possessive, it sounds more like the former (to me, anyway). Or should I make that more verbose?

PR #301, opened 2019-08-30
process: replace -m/--mets requirement by default

bertsky:
When running processors (core CLI or other), it should **not** be required to give `-m mets.xml` but default instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Changed:
+
+  * `-m/--mets` is not required anymore, #301
+
 Fixed:
 
   * typos in docstrings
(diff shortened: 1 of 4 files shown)


bertsky:
> I see the merit in making life easier for users. We should adapt the CLI spec accordingly since core should be faithful to the spec.

Agreed. Shouldn't be too difficult since `mets.xml` is also already required by the OCR-D-ZIP spec.

> bashlib already handles it similar, though there the default is `$PWD/mets.xml` not just `mets.xml`.

But isn't this equivalent here? `process_cli` calls `task_sequence.run_tasks` which uses `Resolver.workspace_from_url`, which effectively uses `os.path.abspath` to convert the METS file argument, and `ocrd_cli_wrap_processor` likewise. So both will convert relative paths w.r.t. the current working directory. Or do any of these functions chdir somewhere in between?

kba:
> So both will convert relative paths w.r.t. the current working directory.

That should be the case, yes. It's just that the bashlib version does it explicitly.

PR #303, opened 2019-09-03
Escape from file hell

### LINKED ISSUE #96: ocrd workspace clone -a -m mets.xml will not adapt links for files

### LINKED ISSUE #227: Avoid superfluous file copies in workspace directory

### LINKED ISSUE #248: validate imagines local images as remote
`ocrd workspace validate` imagines clearly local images as remote images:

~~~sh
cd `mktemp -d`
wget -q https://ocr-d-repo.scc.kit.edu/api/v1/dataresources/f15fb8c8-3842-4314-9a44-5e8b472d7bfc/data/buerger_gedichte_1778.ocrd.zip
dtrx buerger_gedichte_1778.ocrd.zip
cd buerger_gedichte_1778.ocrd/data
ocrd workspace validate mets.xml
~~~

yields:

~~~xml
<report valid="true">
  <notice>Won't download remote image <OCR-D-IMG/OCR-D-IMG_0001></notice>
  <notice>Won't download remote image <OCR-D-IMG/OCR-D-IMG_0002></notice>
</report>
~~~

### LINKED ISSUE #272: "prune-files" command
We have this use case:

  * Start with  fully manifested workspace with some files to be fixed
  * Delete all the files not related to this fix
  * Fix files
  * Reduce mets.xml to only the changed files, i.e.
    * For every mets:file: Does it exist on hdd?
    * If no: delete the mets:file and mets:fptr to that mets:file
  * Remove empty file groups and empty pages in mets.xml

@tboenig Did I get that right?

### LINKED ISSUE #302: Raise FileNotFoundErrors before trying to instantiate XML
Stacktraces like this:

```
  File "/data/monorepo/core/ocrd_modelfactory/ocrd_modelfactory/__init__.py", line 73, in page_from_file
    return parse(input_file.local_filename, silence=True)
  File "/data/monorepo/core/ocrd_models/ocrd_models/ocrd_page_generateds.py", line 9635, in parse
    doc = parsexml_(inFileName, parser)
  File "/data/monorepo/core/ocrd_models/ocrd_models/ocrd_page_generateds.py", line 50, in parsexml_
    doc = etree_.parse(infile, parser=parser, **kwargs)
  File "src/lxml/etree.pyx", line 3424, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1840, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1866, in lxml.etree._parseDocumentFromURL
  File "src/lxml/parser.pxi", line 1770, in lxml.etree._parseDocFromFile
  File "src/lxml/parser.pxi", line 1163, in lxml.etree._BaseParser._parseDocFromFile
  File "src/lxml/parser.pxi", line 601, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 711, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 638, in lxml.etree._raiseParseError
OSError: Error reading file 'OCR-D-GT-PAGE/PAGE_0017_PAGE.xml': failed to load external entity "OCR-D-GT-PAGE/PAGE_0017_PAGE.xml"
```

Are just confusing. It's just that a file does not exist, error message should make that clear.

kba:
I went through the code and tried to get rid of all the logic that was based on `file://` URLs and the complexities from supporting all sorts of URLs.

In particular, filenames will be resolved to relative paths from the workspace directory whereever possible. To achieve that, employ the `pushd_popd` context manager in all places where relative filenames can be expected.

Split workspace tests from resolver tests.

`Workspace.download_to_directory`'s  boolean `overwrite` argument replaced with an `if_exists` string, that can handle existing files as an exception (`"raise"`), skip them (`"skip"`, the default) or overwrite them (`"overwrite"`).

Fix #302 
Fix #272
Fix #248 
Fix #227
Fix #96 

Supersedes #266

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -242,10 +242,10 @@ def prune_files(ctx):
     with pushd_popd(workspace.directory):
         for f in workspace.mets.find_files():
             try:
-                if exists(f.local_filename):
-                    workspace.mets.remove_file(f)
+                if not f.local_filename or not exists(f.local_filename):
+                    workspace.mets.remove_file(f.ID)
             except Exception as e:
-                log.debug("Error removing %f: %s", f, e)
+                log.warning("Error removing %f: %s", f, e)
         workspace.save_mets()
 
 # ----------------------------------------------------------------------
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -135,16 +135,14 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, local_f
     #  log.error(local_filename)
     if not local_filename.startswith(ctx.directory):
         log.debug("File '%s' is not in workspace, copying", local_filename)
-        local_filename = ctx.resolver.download_to_directory(ctx.directory, "file://" + local_filename, subdir=file_grp)
+        local_filename = ctx.resolver.download_to_directory(ctx.directory, local_filename, subdir=file_grp)
     local_filename = relpath(local_filename, ctx.directory)
-
-    url = local_filename
 
     workspace.mets.add_file(
         fileGrp=file_grp,
         ID=file_id,
         mimetype=mimetype,
-        url=url,
+        url=local_filename,
         pageId=page_id,
         force=force,
         local_filename=local_filename
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -245,7 +245,7 @@ def prune_files(ctx):
                 if not exists(f.url):
                     workspace.mets.remove_file(f.ID)
             except Exception as e:
-                log.debug(e)
+                log.debug("Error removing %f: %s", f, e)
         workspace.save_mets()
 
 # ----------------------------------------------------------------------
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -242,8 +242,8 @@ def prune_files(ctx):
     with pushd_popd(workspace.directory):
         for f in workspace.mets.find_files():
             try:
-                if not exists(f.url):
-                    workspace.mets.remove_file(f.ID)
+                if exists(f.local_filename):
+                    workspace.mets.remove_file(f)
             except Exception as e:
                 log.debug("Error removing %f: %s", f, e)
         workspace.save_mets()
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -6,7 +6,7 @@ from tempfile import mkdtemp
 import click
 
 from ocrd import Resolver, Workspace, WorkspaceValidator, WorkspaceBackupManager
-from ocrd_utils import getLogger, pushd_popd, is_local_filename
+from ocrd_utils import getLogger, pushd_popd
 from ..constants import TMP_PREFIX
 
 log = getLogger('ocrd.cli.workspace')
@@ -245,7 +245,8 @@ def prune_files(ctx):
                 if not f.local_filename or not exists(f.local_filename):
                     workspace.mets.remove_file(f.ID)
             except Exception as e:
-                log.warning("Error removing %f: %s", f, e)
+                log.exception("Error removing %f: %s", f, e)
+                raise(e)
         workspace.save_mets()
 
 # ----------------------------------------------------------------------
@@ -306,7 +307,7 @@ def get_id(ctx):
 """)
 @click.argument('ID')
 @pass_workspace
-def set_id(ctx, id):
+def set_id(ctx, id):   # pylint: disable=redefined-builtin
     workspace = Workspace(ctx.resolver, directory=ctx.directory, mets_basename=ctx.mets_basename, automatic_backup=ctx.automatic_backup)
     workspace.mets.unique_identifier = id
     workspace.save_mets()
--- a/ocrd/ocrd/decorators.py
+++ b/ocrd/ocrd/decorators.py
@@ -1,8 +1,11 @@
 import os
+from os.path import isfile
 
 import click
 
 from ocrd_utils import (
+    is_local_filename,
+    get_local_filename,
     setOverrideLogLevel,
     VERSION as OCRD_VERSION
 )
@@ -29,9 +32,7 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
         print(msg)
         raise Exception(msg)
     else:
-        if mets.find('://') == -1:
-            mets = 'file://' + os.path.abspath(mets)
-        if mets.startswith('file://') and not os.path.exists(mets[len('file://'):]):
+        if is_local_filename(mets) and not isfile(get_local_filename(mets)):
             msg = "File does not exist: %s" % mets
             print(msg)
             raise Exception(msg)
--- a/ocrd/ocrd/decorators.py
+++ b/ocrd/ocrd/decorators.py
@@ -10,6 +10,7 @@ from ocrd_utils import (
     VERSION as OCRD_VERSION
 )
 
+from ocrd_utils import getLogger
 from .resolver import Resolver
 from .processor.base import run_processor
 
@@ -22,6 +23,7 @@ loglevel_option = click.option('-l', '--log-level', help="Log level",
                                default=None, callback=_set_root_logger_version)
 
 def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_dir=None, dump_json=False, version=False, **kwargs):
+    LOG = getLogger('ocrd_cli_wrap_processor')
     if dump_json:
         processorClass(workspace=None, dump_json=True)
     elif version:
@@ -29,12 +31,12 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
         print("Version %s, ocrd/core %s" % (p.version, OCRD_VERSION))
     elif mets is None:
         msg = 'Error: Missing option "-m" / "--mets".'
-        print(msg)
+        LOG.error(msg)
         raise Exception(msg)
     else:
         if is_local_filename(mets) and not isfile(get_local_filename(mets)):
             msg = "File does not exist: %s" % mets
-            print(msg)
+            LOG.error(msg)
             raise Exception(msg)
         resolver = Resolver()
         workspace = resolver.workspace_from_url(mets, working_dir)
--- a/ocrd/ocrd/decorators.py
+++ b/ocrd/ocrd/decorators.py
@@ -2,8 +2,10 @@ import os
 
 import click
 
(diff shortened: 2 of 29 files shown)


kba:
In Python 3.5, `pathlib.Path.resolve` has no `strict` argument, hence the test failures.

ocrd/ocrd/cli/workspace.py
```diff
@@ -244,10 +242,10 @@ def prune_files(ctx):
     with pushd_popd(workspace.directory):
         for f in workspace.mets.find_files():
             try:
-                if not exists(f.url):
+                if not f.local_filename or not exists(f.local_filename):
                     workspace.mets.remove_file(f.ID)
             except Exception as e:
-                log.debug(e)
+                log.warning("Error removing %f: %s", f, e)
```
wrznr: Maybe an error should be communicated as `ERROR`?

ocrd/ocrd/decorators.py
```diff
@@ -27,9 +32,7 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
         print(msg)
```
wrznr: Why not use logging here?

ocrd/ocrd/decorators.py
```diff
@@ -27,9 +32,7 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
...
         raise Exception(msg)
     else:
-        if mets.find('://') == -1:
-            mets = 'file://' + os.path.abspath(mets)
-        if mets.startswith('file://') and not os.path.exists(mets[len('file://'):]):
+        if is_local_filename(mets) and not isfile(get_local_filename(mets)):
             msg = "File does not exist: %s" % mets
             print(msg)
```
wrznr: And here?

ocrd/ocrd/resolver.py
```diff
@@ -30,62 +34,77 @@ def download_to_directory(self, directory, url, basename=None, overwrite=False,
             directory (string): Directory to download files to
             basename (string, None): basename part of the filename on disk.
             url (string): URL to download from
-            overwrite (boolean): Whether to overwrite existing files with that name
+            if_exists (string, "skip"): Waht to do if target file already exists. One of ``skip`` (default), ``overwrite`` or ``raise``
```
wrznr: Typo *Waht*

ocrd/ocrd/resolver.py
```diff
@@ -30,62 +34,77 @@ def download_to_directory(self, directory, url, basename=None, overwrite=False,
...
+        ret = str(Path(subdir_path, basename_path))
+        dst_path = Path(directory, ret)
+
+        #  log.info("\n\tdst_path='%s \n\turl=%s", dst_path, url)
+        #  print('url=%s', url)
+        #  print('directory=%s', directory)
+        #  print('subdir_path=%s', subdir_path)
+        #  print('basename_path=%s', basename_path)
+        #  print('ret=%s', ret)
+        #  print('dst_path=%s', dst_path)
```
wrznr: Maybe remove print debugging?

ocrd/ocrd/resolver.py
```diff
@@ -94,55 +113,40 @@ def workspace_from_url(self, mets_url, dst_dir=None, clobber_mets=False, mets_ba
...
-            log.debug("Target and source mets are identical")
-        else:
-            if exists(dst_mets) and not clobber_mets:
-                raise Exception("File '%s' already exists but clobber_mets is false" % dst_mets)
+            mets_basename = nth_url_segment(mets_url, -1)
+
+        # If src_baseurl wasn't given, determine from mets_url by removing last url segment
+        if not src_baseurl:
+            last_segment = nth_url_segment(mets_url)
+            src_baseurl = remove_non_path_from_url(remove_non_path_from_url(mets_url)[:-len(last_segment)])
```
wrznr: This looks like implicit magic. Do we need it?

ocrd/ocrd/workspace.py
```diff
@@ -387,21 +389,23 @@ def image_from_segment(self, segment, parent_image, parent_xywh):
...
+        segment_xywh['y'] -= round(0.5 * max(0,
+                                             segment_image.height - segment_xywh['h']))
         return segment_image, segment_xywh
 
     # pylint: disable=redefined-builtin
     def save_image_file(self, image,
                         file_id,
                         page_id=None,
-                        file_grp='OCR-D-IMG', # or -BIN?
+                        file_grp='OCR-D-IMG',  # or -BIN?
```
wrznr: This interesting since the bags already contain images in the directory `OCR-D-IMG`.

ocrd_modelfactory/ocrd_modelfactory/__init__.py
```diff
@@ -53,7 +56,7 @@ def page_from_image(input_file):
             imageWidth=exif.width,
             imageHeight=exif.height,
             # XXX brittle
```
wrznr: Still brittle?

ocrd_models/ocrd_models/ocrd_file.py
```diff
@@ -32,15 +34,12 @@ def __init__(self, el, mimetype=None, instance=None, local_filename=None, mets=N
...
-                self.local_filename = self.url
-            if self.url.startswith('file://'):
-                self.local_filename = self.url[len('file://'):]
+        if url:
+            self.url = url
 
-        #  if baseurl and not local_filename and '://' not in self.url:
-        #      self.local_filename = '%s/%s' % (baseurl, self.url)
+        if not(local_filename):
+            if self.url and is_local_filename(self.url):
```
wrznr: Is that for cases where the user incorrectly uses the URL field for local filenames? If yes do we really have to rescue them?

ocrd_utils/ocrd_utils/constants.py
```diff
@@ -21,3 +22,11 @@
     '.jpeg': 'image/jpg',
     '.xml': MIMETYPE_PAGE
 }
+
+MIME_TO_EXT = {
+    'image/tiff': '.tif',
+    'image/png': '.png',
+    'image/jpg': '.jpg',
```
wrznr: It may also be `image/jpeg` (https://wiki.selfhtml.org/wiki/MIME-Type/%C3%9Cbersicht#I) !

wrznr:
Vorsichtige Frage: What about #96?

ocrd/ocrd/cli/workspace.py
bertsky: I concur, so how about:
```python
log.exception("Cannot remove file '%s'", f.ID)
```

ocrd/ocrd/workspace.py
bertsky: Yes, by convention they always do. So maybe it would be better to use `None` as default here, to force users of this function to think about _useful_ file groups (hopefully set via their output fileGrp)?

ocrd/ocrd/cli/workspace.py
kba: Agreed, though I don't even remember when an exception can be raised here. But I'll let it throw and we'll see.

ocrd/ocrd/decorators.py
kba: The message is printed (so it appears in STDOUT) and an exception raised. What else would you expect?

ocrd/ocrd/resolver.py
kba: I thought those could be fixed by @stweil's next run with codespell ;-) Just joking, corrected, thanks.

ocrd/ocrd/decorators.py
wrznr: `LOG.error` ?

ocrd/ocrd/resolver.py
kba: It's commented out but I foresee more need for print-debugging in the future. Unless it hurts your sense of code aesthetics too much, I'd prefer to keep them there for the time being.

ocrd/ocrd/decorators.py
kba: I thought there was a reason but there really isn't. I'll log them as errors.

ocrd/ocrd/decorators.py
kba: done.

ocrd/ocrd/resolver.py
wrznr: You may ignore this question if it is dumb.

ocrd/ocrd/workspace.py
kba: > So maybe it would be better to use None as default here, to force users of this function to think about useful file groups 

Change the signature to 

```python
 def save_image_file(self, image,
                        file_id,
                        file_grp,
                        page_id=None,
                        format='PNG',
                        force=True)
```
?

ocrd_modelfactory/ocrd_modelfactory/__init__.py
kba: Yeah, fix is coming in another branch to solve the problem of imageFilename once and hopefully for all.

ocrd_utils/ocrd_utils/constants.py
kba: Added, thanks.

ocrd/ocrd/resolver.py
kba: It's not a dumb question, code is a bit convoluted but necessarily so.

The problem is that URLs can contain fragment identifiers and query strings and trailing slashes. The algorithm here ensures that `src_baseurl` of

`http://foo/path/to/mets.xml?yay-query-string#frag`

resolves to

`http://foo/path/to`

without trailing slash, query string or fragment identifier.

`

ocrd_models/ocrd_models/ocrd_file.py
kba: > incorrectly uses the URL field for local filenames

No, this is for `mets:file` with `mets:Locat` with `@xlink:href` pointing to relative files.

ocrd/ocrd/workspace.py
bertsky: Yes, at this moment we are most likely the only users of that function, so let's change the signature like that!

ocrd/ocrd/decorators.py
kba: The reason was that otherwise it couldn't be tested easily. But it wasn't the most significant test anyway.

ocrd/ocrd/workspace.py
kba: Done in https://github.com/OCR-D/core/pull/303/commits/fc1c0064aee35d5afe2e25d7c90401d563be84fe

kba:
> What about #96?

This should fix #96

PR #320, opened 2019-09-23
:bug: ocrd zip spill: directory -> dest

kba:
pretty sure @bertsky fixed this before but cannot find commit


kba:
It's a regression from https://github.com/OCR-D/core/pull/317

bertsky:
No, my fix (#317 which you already merged) was only partial – sorry.

PR #328, opened 2019-10-09
set LOCTYPE and OTHERLOCTYPE as necessary, fix #310

### LINKED ISSUE #310: mets:file/mets:FLocat requires @LOCTYPE
With current ocrd_mets, add_file does not write the `LOCTYPE` attribute anymore. But this is required by the METS XSD. I believe, `OTHER` is the only appropriate value for relative filesystem paths (with `OTHERLOCTYPE` suggested but not required to be set as well).

kba:
This will allow setting @LOCTYPE and @OTHERLOCTYPE.

Default for @LOCTYPE is `OTHER` and implies `@OTHERLOCTYPE = "file"` for now.

Setting @LOCTYPE to something different from `OTHER` will unset `@OTHERLOCTYPE`.

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_file.py
+++ b/ocrd_models/ocrd_models/ocrd_file.py
@@ -17,7 +17,7 @@ class OcrdFile():
     #  def create(mimetype, ID, url, local_filename):
     #      el_fileGrp.SubElement('file')
 
-    def __init__(self, el, mimetype=None, instance=None, local_filename=None, mets=None, url=None):
+    def __init__(self, el, mimetype=None, loctype='OTHER', instance=None, local_filename=None, mets=None, url=None):
         """
         Args:
             el (LxmlElement):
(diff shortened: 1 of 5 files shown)


ocrd_models/ocrd_models/ocrd_file.py
```diff
@@ -117,6 +118,47 @@ def pageId(self, pageId):
...
+        else:
+            self.otherloctype = None
+
+    @property
+    def otherloctype(self):
+        el_FLocat = self._el.find('mets:FLocat', NS)
+        return '' if el_FLocat is None else el_FLocat.get('OTHERLOCTYPE')
+
+    @otherloctype.setter
+    def otherloctype(self, otherloctype):
```
wrznr: I can set `OTHERLOCTYPE` without setting `LOCTYPE` to `OTHER`? Does this make sense? Especially considering that setting `LOCTYPE` to `OTHER` implicitly sets `OTHERLOCTYPE` to `FILE`...

kba: Good point. Fixed.

Issue #176, opened 2018-08-30
Matching PAGE imageFilename to mets:file when imageFilename is not a URL


kba:
Scenario:

1. Image files and PAGE referencing those image files by relative filepath:
    ```xml
    <Page imageFilename="foo.tif"/>
    ```
2. Create a METS file and run `workspace add`:

    ```xml
	<mets:file GROUPID="page0001" xlink:href="file://path/to/bla/foo.tif"
    ```

Now the PAGE `imageFilename` and `xlink:href` of the corresponding mets:file do not match anymore.

kba:
Solution 1) On `workspace add`, change the `imageFilename` of the PAGE.

Solution 2) Have a in place to map from `@imageFilename` to `@xlink:href` (like "match if fileName is suffix to a mets:file@xlink:href" or "match if fileName is GROUPID of a page and there is a mets:file with mimetype image/* with that GROUPID", this could be automated.)

Solution 3) Track the relation externally. A mechanism like this will be necessary anyway because of the problem of local URLs irreversibly replacing remote URLs when downloading files.

wrznr:
Without overseeing the technical consequences: Only a cosmetic nastiness?  I am not sure we ever touch the file refs in PAGE, do we?

kba:
See https://ocr-d.github.io/page#url-for-imagefilename--filename

The `imageFilename` is necessary to get from page to the `mets:file` that represents the image.

wrznr:
Ah, okay. In this case, :+1: for solution 3.

VolkerHartmann:
Solution 1 only works if workspace add is used which may be a drawback.
Solution 2 sounds complex. There may be several images for the same page (orig, binarized, cropped, deskewed,...)
Solution 3 works out of the box analyzing whole METS and referenced PAGEs in one step. This may be done each time an export/import is planned.
My vote for solution 3.

bertsky:
Pardon my being slow-witted, but what was the reason for https://ocr-d.github.io/page#url-for-imagefilename--filename (always requiring a URL, even if local) in the first place? Why not use relative paths (without `file://` scheme)?

I thought the workspace metaphor would work like a DVCS repository. But if we require URLs everywhere, I cannot move my workspaces around in the filesystem. Am I supposed to `clone -l` instead? 

(BTW, `pack` / `unpack` should also beware of file URLs.)

kba:
[...]

PR #333, opened 2019-10-16
workspace bagger: update PAGE imageFilenames, #176

kba:
After bagging the files, walk through the PAGE files and replace outdated `@imageFilename`

### DIFF
--- a/ocrd/ocrd/workspace_bagger.py
+++ b/ocrd/ocrd/workspace_bagger.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from os import makedirs, chdir, walk, getcwd
+from os import makedirs, chdir, walk
 from os.path import join, isdir, basename, exists, relpath
 from shutil import make_archive, rmtree, copyfile, move
 from tempfile import mkdtemp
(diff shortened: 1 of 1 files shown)


ocrd/ocrd/workspace_bagger.py
```diff
@@ -77,15 +82,40 @@ def _bag_mets_files(self, workspace, bagdir, ocrd_manifestation_depth, ocrd_mets
...
+            for page_file in bag_workspace.mets.find_files(mimetype=MIMETYPE_PAGE):
+                pcgts = page_from_file(page_file)
+                changed = False
+                #  page_doc.set(imageFileName
+            #  for old, new in changed_urls:
+                for old, new in changed_urls.items():
+                    if pcgts.get_Page().imageFilename == old:
+                        pcgts.get_Page().imageFilename = new
+                        changed = True
+                    # TODO replace AlternativeImage, recursively...
```
kba: TODO indeed. @tboenig @bertsky Do you have some good sample PAGE with non-trivial AlternativeImage we could add to assets so I can test this and the solution to the newest iteration of #176 ?

PR #336, opened 2019-10-24
Update Pillow to 6.2.0

kba:
I cannot reproduce the segfaults in 6.2.0 that were plaguing 6.0.0 and 6.1.0.

Tested by running `test_pil_version` a few thousand times, which consistently fails in >= 6.0.0 < 6.2.0 but not in 6.2.0

Can someone confirm this please?

It's also highly advisable to update because of https://nvd.nist.gov/vuln/detail/CVE-2019-16865

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -124,6 +124,7 @@ class TestUtils(TestCase):
             5.4.1 no
             6.0.0 yes
             6.1.0 yes
+            6.2.0 no
         """
         for _ in range(0, 10):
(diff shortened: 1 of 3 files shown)


tests/test_utils.py
```diff
@@ -124,6 +124,7 @@ def test_pil_version(self):
             5.4.1 no
```
wrznr: → yes?

kba: Test does **not** fail in 5.4.1

wrznr: Oh. I thought it reads “the test **should** fail” (in the sense of *you may not use this version*).

PR #337, opened 2019-10-25
spec/assets: update repos, and use cp -f

bertsky:
If you always build and test in a container that clones everything afresh, you would have been fine. Otherwise you need this.

BTW, I am still much in favour of using explicit submodules for this. See OCR-D/ocrd_olena#10 for example.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ uninstall:
 	for mod in $(BUILD_ORDER);do pip uninstall -y $$mod;done
 
 # Regenerate python code from PAGE XSD
-generate-page: repo/assets
+generate-page: assets
 	generateDS \
 		-f \
 		--root-element='PcGts' \
@@ -104,17 +104,20 @@ repo/spec:
 .PHONY: spec
 # Copy JSON Schema, OpenAPI from OCR-D/spec
 spec: repo/spec
-	cp repo/spec/ocrd_tool.schema.yml ocrd_validators/ocrd_validators/ocrd_tool.schema.yml
-	cp repo/spec/bagit-profile.yml ocrd_validators/ocrd_validators/bagit-profile.yml
+	git -C $< pull
+	cp -f repo/spec/ocrd_tool.schema.yml ocrd_validators/ocrd_validators/ocrd_tool.schema.yml
+	cp -f repo/spec/bagit-profile.yml ocrd_validators/ocrd_validators/bagit-profile.yml
 
 #
 # Assets
 #
 
+.PHONY: assets
 # Setup test assets
 assets: repo/assets
+	git -C $< pull
 	mkdir -p $(TESTDIR)/assets
-	cp -r -t $(TESTDIR)/assets repo/assets/data/*
+	cp -f -r -t $(TESTDIR)/assets repo/assets/data/*
 
 # Start asset server at http://localhost:5001
 assets-server:


kba:
We have a monorepo that symlinks `REPOROOT/core/repo/spec` to `REPOROOT/spec` and `REPOROOT/core/repo/assets` to `REPOROOT/assets` and so on. When testing in CI or containers, everything is cloned afresh as you say. The disadvantage of submodules is that they require constant updates across repositories to keep in sync.

I also don't understand why you have to pull `spec` everytime you do `make spec`. It doesn't hurt much since `make spec` is only required infrequently but why couldn't you manually update `spec` before `make spec`?

kba:
How about something like

```make
pull-repos:
    for repo in repo/*;do git -C $$repo pull;done
```

to update manually or combine this in your calls, e.g. `make pull-repos generate-page`?

bertsky:
> I also don't understand why you have to pull `spec` everytime you do `make spec`. It doesn't hurt much since `make spec` is only required infrequently but why couldn't you manually update `spec` before `make spec`?

Well exactly because whenever spec is required (e.g. for `make test`) and you **don't start off afresh**, but have a clone from a previous run already in place, then things will go wrong on the old version. This also goes for assets. Requiring to do that manually is tedious and unexpected.

> The disadvantage of submodules is that they require constant updates across repositories to keep in sync.

They **enable** you to keep things in sync. For example, some core branch/revision may _not_ work with the most recent spec/assets master. Making these dependencies explicit – especially when you already provide Makefile control – is a large benefit. The extra cost of having update rules in place is small IMHO. (You would then _manually_ have to checkout/pull the subrepos' versions as a commiter. All the user would do is _automatically_ update.)

bertsky:
> How about something like
> 
> ```makefile
> pull-repos:
>     for repo in repo/*;do git -C $$repo pull;done
> ```
> 
> to update manually or combine this in your calls, e.g. `make pull-repos generate-page`?

That still leaves me with the burden of having to know that and when I have to do this extra step.

kba:
> That still leaves me with the burden of having to know that and when I have to do this extra step.

I understand but having to pull twice for every `make test` seems extreme in the other direction. But if you want that behavior, why don't you replace `make test` with `make pull-repos test` consistently? It's a bit more to type but would keep repo management and building/testing separate?

kba:
> Making these dependencies explicit – especially when you already provide Makefile control – is a large benefit. 

Yes but with the symlink pattern we're using, we're absolutely in sync with the current state of assets/spec, even uncomitted changes for testing changes. Having multiple instances of these repos will lead (and has led before) to drift.

Compromise: I'll investigate whether our symlink pattern and submodules can be combined and in the meantime we make the `make pull-repos` change?

bertsky:
> > That still leaves me with the burden of having to know that and when I have to do this extra step.
> 
> I understand but having to pull twice for every `make test` seems extreme in the other direction. 

Why is that extreme? It's not costly, if that's what you mean. Having a regime in which you always clone afresh is. This is all about making other regimes work (better).

> But if you want that behavior, why don't you replace `make test` with `make pull-repos test` consistently? It's a bit more to type but would keep repo management and building/testing separate?

No, it would not. Submodule control would do that. (I'd always get a consistent checkout.)

This is not about me. I can remember `git -C repos/spec pull` just fine. But I already _know_ about this. Others don't. They just download and `make test`. Then they update and want to `make test` again.

kba:
As I said, I can investigate getting submodules and symlinks to work but calling make pull twice just to run tests is not acceptable for me. 

> Having a regime in which you always clone afresh.

As I said, we symlink and have to keep the linked repos updated. But not always clone? And even with submodules, you have to do an initial clone after checkout if that's what you mean.

I want a setup that doesn't require network requests for simple tasks like running unit tests. Submodules might work but this doesn't, sorry.

PR #339, opened 2019-11-01
reduce pixel_density problem severity: error -> warning,

kba:
OCR-D/spec#129
OCR-D/assets#33

### DIFF
--- a/ocrd_validators/ocrd_validators/workspace_validator.py
+++ b/ocrd_validators/ocrd_validators/workspace_validator.py
@@ -155,7 +155,7 @@ class WorkspaceValidator():
             for k in ['xResolution', 'yResolution']:
                 v = exif.__dict__.get(k)
                 if v is None or v <= 72:
-                    self.report.add_error("Image %s: %s (%s pixels per %s) is too low" % (f.ID, k, v, exif.resolutionUnit))
+                    self.report.add_warning("Image %s: %s (%s pixels per %s) is too low" % (f.ID, k, v, exif.resolutionUnit))
 
     def _validate_mets_file_group_names(self):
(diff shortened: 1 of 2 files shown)


ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -155,7 +155,7 @@ def _validate_pixel_density(self):
             for k in ['xResolution', 'yResolution']:
                 v = exif.__dict__.get(k)
                 if v is None or v <= 72:
-                    self.report.add_error("Image %s: %s (%s pixels per %s) is too low" % (f.ID, k, v, exif.resolutionUnit))
+                    self.report.add_warning("Image %s: %s (%s pixels per %s) is too low" % (f.ID, k, v, exif.resolutionUnit))
```
bertsky: I don't think this is the right thing to do here. This can be a significant problem for processors, so there should be a loud and clear error.

Moreover, core should at least provide a running definition of `suspiciously low` DPI – not just in the validator, but in OcrdExif too, and not just in isolation, but in relation to the pixel resolution as well.

bertsky: IMHO the best way to react in OcrdExif is to raise an error and set the density to `1` (as if nothing had been set) – all under the assumption of OCR-D/spec#129 that correct settings cannot be enforced in the first place.

kba:
I'll merge this now for the 2.0.0 release but as with OCR-D/spec#129 the discussion shall continue! @bertsky @wrznr feel free to open an issue here or in spec to discuss the rules of the heuristics and how we should handle bad input images.

PR #350, opened 2019-11-19
add python3-venv dependency to deps-ubuntu

bertsky:
perhaps also (for modules still using the old spec) `python3-virtualenv`?


kba:
> perhaps also (for modules still using the old spec) python3-virtualenv?

Probably best not to, we should rather update our documentation. `venv` is part of Python3 (though Debian for some reasone removes it from the base package).

PR #368, opened 2019-12-04
docker: add curl and software-properties-common for apt-add-repository

kba:
c.f. https://github.com/OCR-D/ocrd_tesserocr/pull/97


kba:
Yeah, I opened this before the discussion finished. At least we can get rid of the libglib-2.0.0 dependency.

PR #369, opened 2019-12-06
implement check for multi-page TIFF, fix #243

### LINKED ISSUE #243: Validator: Ensure single-page TIFF
HT @mikegerber:

```python
t = Image.open(self.tiff)
t.load()

if t.n_frames != 1:
    warnings.warn('{} is a multi-page TIFF'.format(self.tiff))
# XXX t.info['compression'] - Do the TIFFs really need to be uncompressed?
```

kba:
@mike-gerber @bertsky @wrznr

### DIFF
--- a/ocrd_validators/ocrd_validators/workspace_validator.py
+++ b/ocrd_validators/ocrd_validators/workspace_validator.py
@@ -85,6 +85,8 @@ class WorkspaceValidator():
                 self._validate_mets_files()
             if 'pixel_density' not in self.skip:
                 self._validate_pixel_density()
+            if 'multipage' not in self.skip:
+                self._validate_multipage()
             if 'dimension' not in self.skip:
                 self._validate_dimension()
             if 'imagefilename' not in self.skip:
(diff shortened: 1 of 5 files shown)


ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -141,13 +143,27 @@ def _validate_dimension(self):
...
+
+        See `spec <https://ocr-d.github.io/mets#no-multi-page-images>`_.
+        """
+        for f in [f for f in self.mets.find_files() if f.mimetype.startswith('image/') and f.url]:
+            if not is_local_filename(f.url) and not self.download:
+                self.report.add_notice("Won't download remote image <%s>" % f.url)
+                continue
+            exif = self.workspace.resolve_image_exif(f.url)
+            if exif.n_frames > 1:
+                self.report.add_warning("Image %s: More than 1 frame: %s" % (f.ID, exif.n_frames))
```
wrznr: Only a warning?

ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -141,13 +143,27 @@ def _validate_dimension(self):
             if page.imageWidth != exif.width:
                 self.report.add_error("PAGE '%s': @imageWidth != image's actual width (%s != %s)" % (f.ID, page.imageWidth, exif.width))
 
+    def _validate_multipage(self):
+        """
+        Validate the number of images per file is 1 (TIFF allows multi-page images)
+
+        See `spec <https://ocr-d.github.io/mets#no-multi-page-images>`_.
+        """
+        for f in [f for f in self.mets.find_files() if f.mimetype.startswith('image/') and f.url]:
```
wrznr: `f.url` test is only used for two validation methods. Intended?

ocrd_validators/ocrd_validators/workspace_validator.py
kba: Oh, right, we decided on making this an error in https://github.com/OCR-D/spec/pull/132, I'll update.

ocrd_validators/ocrd_validators/workspace_validator.py
kba: IIRC that was to guard against `<mets:file>` w/o FLocat. Which is actually allowed in METS, annoyingly. Better not to be lenient, I'll remove the check and see what breaks :)

kba: Or even better, make that a separate check in the validator.

kba: Is now checked in `_validate_mets_files`

ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -141,6 +147,20 @@ def _validate_dimension(self):
             if page.imageWidth != exif.width:
                 self.report.add_error("PAGE '%s': @imageWidth != image's actual width (%s != %s)" % (f.ID, page.imageWidth, exif.width))
 
+    def _validate_multipage(self):
+        """
+        Validate the number of images per file is 1 (TIFF allows multi-page images)
+
+        See `spec <https://ocr-d.github.io/mets#no-multi-page-images>`_.
+        """
+        for f in [f for f in self.mets.find_files() if f.mimetype.startswith('image/')]:
```
bertsky:
```suggestion
        for f in self.mets.find_files():
            if not f.mimetype.startswith('image/'):
                continue
```

kba: What's the difference?

bertsky: None – that's the point! It's more readable and more Pythonic IMHO. But as it says, it's only a suggestion.

kba: `for f in [f for f in` is quirky I agree but we use that pattern consistently in the tests and I don't want to change them all. Plus for tests I often find one-liners helpful to make it easier to copy/paste/replace all.

PR #370, opened 2019-12-09
Install latest Python packages pip, wheel

stweil:
wheel is required for builds.

pip is already available, but typically too old.
pip 19.0 or newer is required for manylinux2010 platform tag support.
pip 19.3 or newer is required for manylinux2014 platform tag support.
Installation of tensorflow==2.0.0 for example requires manylinux2010.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,7 @@ deps-test:
 
 # (Re)install the tool
 install: spec
+	$(PIP) install -U "pip>=19.0.0" wheel
 	for mod in $(BUILD_ORDER);do (cd $$mod ; $(PIP_INSTALL) .);done
 
 # Uninstall the tool
--- a/tests/model/test_ocrd_mets.py
+++ b/tests/model/test_ocrd_mets.py
@@ -28,7 +28,7 @@ class TestOcrdMets(TestCase):
         self.assertEqual(mets.unique_identifier, 'foo', 'Right identifier after change')
         as_string = mets.to_xml().decode('utf-8')
         self.assertIn('ocrd/core v%s' % VERSION, as_string)
-        self.assertIn('CREATEDATE="%d-%d-%02dT' % (
+        self.assertIn('CREATEDATE="%04u-%02u-%02uT' % (
             datetime.now().year,
             datetime.now().month,
             datetime.now().day,


stweil:
See also related discussion https://github.com/Calamari-OCR/calamari/issues/145.

stweil:
Why does Travis CI set `PIP_INSTALL="pip install -e"`? It was unused previously (and I think it would never have worked because `-e` expects an argument) and causes an error now.

stweil:
ci/circleci was broken because of a wrong ISO 8601 date format. This is fixed with the 2nd commit.

Makefile
```diff
@@ -69,6 +69,7 @@ deps-test:
 
 # (Re)install the tool
 install: spec
+	$(PIP_INSTALL) -U pip wheel
```
bertsky:
```suggestion
	$(PIP) install -U pip wheel
```

stweil: Done.

stweil: Adding wheel to `requirements.txt` did not work for me because that comes too late.

bertsky: > Adding wheel to `requirements.txt` did not work for me because that comes too late.

You mean `pip install -r requirements.txt` instead of `pip install -U pip wheel` does not work? I don't understand... (But it's probably too insignificant.)

Makefile
```diff
@@ -69,6 +69,7 @@ deps-test:
 
 # (Re)install the tool
 install: spec
+	$(PIP) install -U "pip>=9.0.0" wheel
```
bertsky: Your changelog says:
> pip 19.0 or newer is required for manylinux2010 platform tag support.
> pip 19.3 or newer is required for manylinux2014 platform tag support.

Yet here you now only require `pip>=9.0`? (And I believe you should use either `-U` or a minimal version.)

stweil: manylinux2014 seems to be still rarely required. For tensorflow only manylinux2010 is needed.

kba: But still, you meant `pip>=19.0.0` not `pip>=9.0.0`, right?

kba:
> Why does Travis CI set PIP_INSTALL="pip install -e"?

Not sure anymore but I think it was a workaround to Travis' old Ubuntu distro behaving diffeently.

Makefile
stweil: My mistake, sorry. I fixed it now.

PR #373, opened 2019-12-18
raise exceptions for pseudo-file URL (file:/path instead of file://path)

kba:


kba:
Thanks for the quick review

> We only have to make sure that all files in assets and in the GT “repo” are correct with respect to this modification.

Exactly, that's what I'm trying to ensure. Not silently correcting these faulty URLs is the first step.

Issue #259, opened 2019-07-11
PAGE validator: show physical pageId or METS file ID


bertsky:
The METS-related validators all show the `mets:file/@ID` of the incriminated page. But the PAGE validator does not: it just points to the ID of the element on the page. So there is no way in knowing what file it relates to (as PAGE identifiers need not be unique across files), and it is _very_ inconvenient.

Mitigation will not be simple: `PageValidator` does not necessarily know the `OcrdFile` which the page originated from, and `PageType` has no ID and `pcGtsId` is usually empty or unrelated.

IMO, at least when calling `PageValidator.validate` from the workspace validator the file id or page id should be passed in, and used in the error description.

kba:
As you say, this would be tricky for the standalone PAGE validator but when called in the context of the METS file referencing it, page ID can be passed to the constructor and output in the logs.

bertsky:
Is it just me, is no-one else constantly looking at these failed reports and wondering where the hell they come from? Probably everyone got used to running with most checks disabled anyway?

bertsky:
Fixed by #374

PR #374, opened 2019-12-19
PAGE validator: coordinate self-validity and mutual consistency

bertsky:
- improve recursive validation of multi-level textequiv consistency:
  * regions (e.g. Table) may contain regions (e.g. Text)
  * 'fix' recursively by going bottom-up
  * better reports, more logging
  * simplify
- add recursive validation of multi-level coordinate consistency:
  * add all region types, also consider Baseline
  * introduce 2 more error cases into ValidationReport:
    - CoordinateValidityException (self-intersecting, non-contiguous,
      too short or negative path)
    - CoordinateConsistencyException (child extends beyond parent)
  * introduce 2 new parameters:
    - check_coords (valid paths and consistent child/parent relations
      within the normal hierarchy)
    - check_baseline (the same only for TextLine.Coords vs Baseline)
  * expose them as 1 new CLI parameter `--page-discipline` (analoguous to
    `--page-strictness` for text) as enum:
    - poly (only check_coords)
    - baseline (only check_baseline)
    - both
    - off
  * use shapely.geometry for checks
  * handle PAGE ideosyncrasies (no Coords or ID at PageType / Baseline)
  * try to be efficient (re-use Polygon objects)
- remove unnecessary/redundant instantiation and method from PageValidator

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -48,16 +48,18 @@ def workspace_cli(ctx, directory, mets_basename, backup):
 @pass_workspace
 @click.option('-a', '--download', is_flag=True, help="Download all files")
 @click.option('-s', '--skip', help="Tests to skip", default=[], multiple=True, type=click.Choice(['imagefilename', 'dimension', 'mets_unique_identifier', 'mets_file_group_names', 'mets_files', 'pixel_density', 'page', 'url']))
-@click.option('--page-strictness', help="How strict to check PAGE consistency", type=click.Choice(['strict', 'lax', 'fix', 'off']), default='strict')
+@click.option('--page-strictness', help="How strict to check PAGE multi-level textequiv consistency", type=click.Choice(['strict', 'lax', 'fix', 'off']), default='strict')
+@click.option('--page-discipline', help="How fierce to check PAGE multi-level coordinate consistency", type=click.Choice(['poly', 'baseline', 'both', 'off']), default='poly')
 @click.argument('mets_url')
-def validate_workspace(ctx, mets_url, download, skip, page_strictness):
+def validate_workspace(ctx, mets_url, download, skip, page_strictness, page_discipline):
     report = WorkspaceValidator.validate(
         ctx.resolver,
         mets_url,
         src_dir=ctx.directory,
         skip=skip,
         download=download,
-        page_strictness=page_strictness
+        page_strictness=page_strictness,
+        page_discipline=page_discipline
     )
     print(report.to_xml())
     if not report.is_valid:
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -48,10 +48,10 @@ def workspace_cli(ctx, directory, mets_basename, backup):
 @pass_workspace
 @click.option('-a', '--download', is_flag=True, help="Download all files")
 @click.option('-s', '--skip', help="Tests to skip", default=[], multiple=True, type=click.Choice(['imagefilename', 'dimension', 'mets_unique_identifier', 'mets_file_group_names', 'mets_files', 'pixel_density', 'page', 'url']))
-@click.option('--page-strictness', help="How strict to check PAGE multi-level textequiv consistency", type=click.Choice(['strict', 'lax', 'fix', 'off']), default='strict')
-@click.option('--page-discipline', help="How fierce to check PAGE multi-level coordinate consistency", type=click.Choice(['poly', 'baseline', 'both', 'off']), default='poly')
+@click.option('--page-textequiv-consistency', '--page-strictness', help="How strict to check PAGE multi-level textequiv consistency", type=click.Choice(['strict', 'lax', 'fix', 'off']), default='strict')
+@click.option('--page-coordinate-consistency', help="How fierce to check PAGE multi-level coordinate consistency", type=click.Choice(['poly', 'baseline', 'both', 'off']), default='poly')
 @click.argument('mets_url')
-def validate_workspace(ctx, mets_url, download, skip, page_strictness, page_discipline):
+def validate_workspace(ctx, mets_url, download, skip, page_strictness, page_coordinate_consistency):
     report = WorkspaceValidator.validate(
         ctx.resolver,
         mets_url,
@@ -59,7 +59,7 @@ def validate_workspace(ctx, mets_url, download, skip, page_strictness, page_disc
         skip=skip,
         download=download,
         page_strictness=page_strictness,
-        page_discipline=page_discipline
+        page_coordinate_consistency=page_coordinate_consistency
     )
     print(report.to_xml())
     if not report.is_valid:
--- a/ocrd_validators/ocrd_validators/page_validator.py
+++ b/ocrd_validators/ocrd_validators/page_validator.py
@@ -58,7 +58,7 @@ _HIERARCHY = [
     (GlyphType,      None,             None), # pylint: disable=bad-whitespace
 ]
 
-class TextequivConsistencyError(Exception):
+class ConsistencyError(Exception):
     """
     Exception representing a consistency error in textual transcription across levels of a PAGE-XML.
     (Element text strings must be the concatenation of their children's text strings, joined by white space.)
@@ -66,7 +66,7 @@ class TextequivConsistencyError(Exception):
 
     def __init__(self, tag, ID, file_id, actual, expected):
         """
-        Construct a new TextequivConsistencyError.
+        Construct a new ConsistencyError.
 
         Arguments:
             tag (string): Level of the inconsistent element (parent)
@@ -81,7 +81,7 @@ class TextequivConsistencyError(Exception):
         self.file_id = file_id
         self.actual = actual
         self.expected = expected
-        super(TextequivConsistencyError, self).__init__(
+        super(ConsistencyError, self).__init__(
             "INCONSISTENCY in %s ID '%s' of file '%s': text results '%s' != concatenated '%s'" % (
                 tag, ID, file_id, actual, expected))
 
@@ -231,7 +231,7 @@ def validate_consistency(node, strictness, strategy, check_baseline, check_coord
                 elif (strictness == 'strict' # or 'lax' but...
                       or not compare_without_whitespace(concatenated, text_results)):
                     log.debug("Inconsistent text of %s %s", tag, node_id)
-                    report.add_error(TextequivConsistencyError(tag, node_id, file_id,
+                    report.add_error(ConsistencyError(tag, node_id, file_id,
                                                       text_results, concatenated))
     return consistent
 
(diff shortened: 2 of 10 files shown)


bertsky:
The second commit fixes #259. 

Note you might additionally want to set the default filter for coordinate checks to off, because users will see _massive_ consistency and validity problems with our current processors – as could be expected, since we did not challenge this right away. The current default is `page-discipline=poly`, but you may need `off`. Also, the default loglevel of `ocrd.page_validator` might have to be increased from `INFO` to `WARNING` now.

bertsky:
The tests obviously still have to be updated, as the number of failures increases and messages look different. @kba could you please do that yourself?

kba:
> @kba could you please do that yourself?

on it.

ocrd_validators/ocrd_validators/page_validator.py
```diff
@@ -2,99 +2,244 @@
...
     Compare two strings, ignoring all whitespace.
     """
     return re.sub('\\s+', '', a) == re.sub('\\s+', '', b)
 
-def handle_inconsistencies(node, strictness, strategy, report):
+def validate_consistency(node, strictness, strategy, check_baseline, check_coords, report, file_id):
     """
-    Check whether the text results on an element is consistent with its child element text results.
+    Check whether the text results on an element is consistent with its child element text results,
+    and whether the coordinates of an element are fully within its parent element coordinates.
```
kba:
```python
"""
Arguments:
   node : PAGE XML element to validate for consistency
   strictness (string) : Strictness of the textequiv consistency check.
  strategy (string) : Strategy to deal with textequiv inconsistencies
  check_baseline (boolean) : Whether to check baseline coordinate consistency
  check_coords (boolean) : Whether to check coordinate consistency of polygons
  report (ValidationReport) : Report to add to
  file_id (string): ID of the mets:file being checked
"""
```

Right?

ocrd_validators/ocrd_validators/page_validator.py
```diff
@@ -2,99 +2,244 @@
...
+                if (not child_poly.is_valid
+                    or child_poly.is_empty
+                    or child_poly.bounds[0] < 0
+                    or child_poly.bounds[1] < 0
+                    or child_poly.length < 4):
+                    report.add_error(CoordinateValidityError(child_tag, child.id, file_id, child_points))
+                    log.info("Invalid coords of %s %s", child_tag, child.id)
+                    consistent = False
+                elif not child_poly.within(node_poly):
+                    # TODO: automatic repair?
```
kba: Automatic repair if it's just one or two pixels, you mean? Probably better not to do here, I am not sure whether the "fixing-textequiv-consistency" functionality is wise here.

ocrd_validators/ocrd_validators/page_validator.py
```diff
@@ -146,39 +293,29 @@ def validate(filename=None, ocrd_page=None, ocrd_file=None, strictness='strict',
             ocrd_file (OcrdFile): OcrdFile instance wrapping OcrdPage
             strictness (string): 'strict', 'lax', 'fix' or 'off'
             strategy (string): Currently only 'index1'
+            check_baseline (bool): whether Baseline must be fully within TextLine/Coords
+            check_coords (bool): whether *Region/TextLine/Word/Glyph must each be fully
+                                 contained within Border/*Region/TextLine/Word, resp.
```
kba: Ah, forget the question above then.

ocrd_validators/ocrd_validators/page_validator.py
```diff
@@ -146,39 +293,29 @@ def validate(filename=None, ocrd_page=None, ocrd_file=None, strictness='strict',
...
 
         Returns:
             report (:class:`ValidationReport`) Report on the validity
         """
         if ocrd_page:
-            validator = PageValidator(ocrd_page, strictness, strategy)
+            page = ocrd_page
+            file_id = ocrd_page.get_pcGtsId()
```
kba: Do we require PAGE pcGtsId and mets:file ID to match? @wrznr @tboenig If we do not we should, right?

ocrd_validators/ocrd_validators/page_validator.py
kba: Disregard, documented down.

kba:
Please merge https://github.com/OCR-D/core/pull/375 and I will merge this.

ocrd_validators/ocrd_validators/page_validator.py
wrznr: Not sure. Might complicate the processing of existing PAGE.

bertsky: I'm also inclined to say we should, because it could help make logging output more usable. But I agree we also should not rely on that identity. And keep non-empty pcGtsId on import.

Anyway, here it's just "anything is better than nothing".

kba: Yeah, this is not related to this PR but we might want to check it in the workspace validation and at least notice users.

kba: https://github.com/OCR-D/core/issues/378

ocrd_validators/ocrd_validators/page_validator.py
bertsky: > Automatic repair if it's just one or two pixels, you mean?

I thought of really intersecting with the parent polygon – so naughty children will be restricted by their parents. (This is the opposite strategy of ocrd-segment-repair with sanitize=True.) Just as a simple counter-measure that comes with next to no cost and can be used readily.

> Probably better not to do here, I am not sure whether the "fixing-textequiv-consistency" functionality is wise here.

Why not? As long as it's not the default setting...

kba: > Why not? As long as it's not the default setting...

Because I would expect that kind of feature in e.g. `ocrd-segment-repair` but not the PAGE validator. Seems non-trivial to use this in a batch processing setting though I haven't tried. Nothing against fixing per se.

ocrd/ocrd/cli/workspace.py
```diff
@@ -48,16 +48,18 @@ def workspace_cli(ctx, directory, mets_basename, backup):
 @pass_workspace
 @click.option('-a', '--download', is_flag=True, help="Download all files")
 @click.option('-s', '--skip', help="Tests to skip", default=[], multiple=True, type=click.Choice(['imagefilename', 'dimension', 'mets_unique_identifier', 'mets_file_group_names', 'mets_files', 'pixel_density', 'page', 'url']))
-@click.option('--page-strictness', help="How strict to check PAGE consistency", type=click.Choice(['strict', 'lax', 'fix', 'off']), default='strict')
+@click.option('--page-strictness', help="How strict to check PAGE multi-level textequiv consistency", type=click.Choice(['strict', 'lax', 'fix', 'off']), default='strict')
+@click.option('--page-discipline', help="How fierce to check PAGE multi-level coordinate consistency", type=click.Choice(['poly', 'baseline', 'both', 'off']), default='poly')
```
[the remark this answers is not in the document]
bertsky: I think it's hopelessly ideosyncratic anyway, so better keep it short.

Or can we _rename_ the original parameter (and merely point the old option via an alias)? Then we could have, for example,
- `page-textequiv-consistency`
- `page-coordinate-consistency`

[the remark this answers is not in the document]
kba: > * page-textequiv-consistency
> * page-coordinate-consistency

:+1:

bertsky:
> Cannot push to https://github.com/bertsky/core/tree/validate-coords PR is just for CI.

Strange. It says `allow edits from maintainers` here...

> Please merge #375 and I will merge this.

Regarding `{,Textequiv}ConsistencyError` – I was hoping we could avoid renaming this, because it changes the existing API. I already use the old name [here](https://github.com/OCR-D/ocrd_keraslm/blob/da105a8a8b68844389cd3e08307c05c9c6123350/ocrd_keraslm/wrapper/rate.py#L323)...

kba:
> Strange. It says allow edits from maintainers here...

Yah, GitHub hates me today, not the only repo I cannot push to. temporary issue let's hope.

> Regarding {,Textequiv}ConsistencyError – I was hoping we could avoid renaming this, because it changes the existing API. I already use the old name here...

Wasn't aware of that, I'll revert.

ocrd_validators/ocrd_validators/page_validator.py
bertsky: I see. I thought of ocrd-segment-repair rather as a provider of non-trivial, logical repairs – whereas this (as well as the `strictness=fix` for textequiv) is a more automatic, pragmatic strategy. Besides, a dedicated processor can only add new annotation into the workspace, whereas the validator is responsible for making all existing annotations conform. What do you think?

kba: It makes sense to have these formulaic fixing close to the validation of the form, more so as the input data is changed not amended as you say. If it's simple coordinate fixes you meant with the TODO I'm fine having it in here (though we should change the method signatures accordingly, textequiv-strategy etc). If it's more complex, i.e. more than a few lines, it's better to have a dedicated processor

bertsky: >  If it's simple coordinate fixes you meant with the TODO

It is. Where we now check `child_poly.within(parent_poly)` we would then have to do `child_poly.intersection(parent_poly)`, and write that back to the element via `child.get_Coords().points = points_from_polygon(child_poly.coords)`.

> (though we should change the method signatures accordingly, textequiv-strategy etc)

Again I'd like to preserve backwards compatibility, so if `PageValidator.strictness` and `.strategy` are to be renamed, they at least need an alias kwarg.

For the new options, how about:
- CLI: `--page-coordinate-consistency=poly|fix-poly|baseline|fix-baseline|both|fix-both|off`
- API: `check_coords` `fix_coords` `check_baseline` `fix_baseline`

Or should the API more look like the CLI, or conversely?

kba: Point taken on backwards compatibility. I will rename the `page[-_]discipline` to `page-cordinate-consistency` but keep the old labels for textequiv checks for now.

kba: Please have a look at https://github.com/OCR-D/core/pull/375/commits/5ca8f7324b94cf0ea85d4afb9234d1f4296e7e96

bertsky: > Please have a look at [5ca8f73](https://github.com/OCR-D/core/commit/5ca8f7324b94cf0ea85d4afb9234d1f4296e7e96)

LGTM. I would probably not have elevated this to _deprecation_, but why not. I'll merge this now.

bertsky: So @kba do you want me to add the simple coordinate fixing as outlined above, or is this for a separate PR? (We'd still have to find a good mnemonic for that on the CLI and API.)

bertsky:
The partial CI failure on Travis is merely a temporary connectivity issue. How do I re-initiate the build without making an actual change?

kba:
>How do I re-initiate the build without making an actual change?

You can rerun jobs in travis ui, I'll do that now

ocrd_validators/ocrd_validators/page_validator.py
kba: I'd prefer a separate PR, I want to merge this and try it out on our GT asap

Issue #353, opened 2019-11-21
bashlib: expose validators


bertsky:
It would be very useful for OCR-D shell wrappers and non-OCR-D use-cases alike for core's bashlib to expose the various existing functions in `ocrd_validators` as separate CLI commands (i.e. not just on the workspace as a whole). For example:

```sh
ocrd validate parameters paramfile.json ocrd-tool.json
ocrd validate page [--strictness level] page.xml
```

This could also delegate to existing commands in the other action groups (for convenience):

```sh
ocrd validate ocrd-tool ocrd-tool.json
ocrd validate workspace [-s tests] [--page-strictness level] mets.xml
ocrd validate process [-m mets] command1 command2 ...
ocrd validate zip [options] source
```

bertsky:
Also: Offer restricting the workspace validator to a subset of file groups.

kba:
> Also: Offer restricting the workspace validator to a subset of file groups.

#383

kba:
A dedicated `validate` CLI is a useful addition and more intuitive than e.g. the infamous

`ocrd ocrd-tool ocrd-tool.json validate`

:)

Should be simple to implement.

kba:
Implemented and released in https://github.com/OCR-D/core/releases/tag/v2.3.0

PR #384, opened 2019-12-27
"ocrd validate" cli, #353

kba:
Yet untested, soliciting feedback on the interface before working further on this.

### DIFF
--- a/ocrd/ocrd/cli/validate.py
+++ b/ocrd/ocrd/cli/validate.py
@@ -55,9 +55,11 @@ def validate_parameters(ocrd_tool, executable, param_json):
     _inform_of_result(ParameterValidator(ocrd_tool['tools'][executable]).validate(parse_json_string_or_file(param_json)))
 
 @validate_cli.command('page')
-@click.argument('page')
+@click.argument('page', required=True, nargs=1)
 @click.option('--page-textequiv-consistency', help="How strict to check PAGE multi-level textequiv consistency", type=click.Choice(['strict', 'lax', 'fix', 'off']), default='strict')
-@click.option('--page-coordinate-consistency', help="How fierce to check PAGE multi-level coordinate consistency", type=click.Choice(['poly', 'baseline', 'both', 'off']), default='poly')
+@click.option('--page-textequiv-strategy', help="Strategy to determine the correct textequiv", type=click.Choice(['index1']), default='index1')
+@click.option('--check-baseline', help="Whether Baseline must be fully within TextLine/Coords", is_flag=True, default=False)
+@click.option('--check-coords', help="Whether *Region/TextLine/Word/Glyph must each be fully contained within Border/*Region/TextLine/Word, resp.", is_flag=True, default=False)
 def validate_page(page, **kwargs):
     '''
     Validate PAGE against OCR-D conventions
--- a/ocrd/ocrd/cli/validate.py
+++ b/ocrd/ocrd/cli/validate.py
@@ -31,11 +31,13 @@ def validate_cli():
     """
 
 @validate_cli.command('tool-json')
-@click.argument('ocrd_tool')
+@click.argument('ocrd_tool', required=False, nargs=1)
 def validate_ocrd_tool(ocrd_tool):
     '''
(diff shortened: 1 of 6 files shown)


ocrd/ocrd/cli/validate.py
```diff
@@ -0,0 +1,100 @@
...
+    OcrdToolValidator,
+    OcrdZipValidator,
+    PageValidator,
+    ParameterValidator,
+    WorkspaceValidator,
+)
+
+def _inform_of_result(report):
+    if not report.is_valid:
+        print(report.to_xml())
```
bertsky: Maybe this should be done via logging facilities?

kba: That report is multiline which isn't ideal for logging. Also, logging can be disabled by the user but if a validation check fails, we always want to show the report.

bertsky: > That report is multiline which isn't ideal for logging.

You could wrap the individual lines of the report into log messages.

> Also, logging can be disabled by the user but if a validation check fails, we always want to show the report.

But isn't this a case of logging-as-output? And you could even argue the actual result is the exit value.

Anyway, I leave that entirely to you – withdrawing my request.

kba: Yeah, I think the value of having a clean solution is outweighed by convenience. I'm thinking of users running this in batch on lots of files, generating lots of reports (@tboenig does that regularly for our GT). As-is the reports are crude but well-formed XML which can be used.

Needs more testing though, I'll try to write some unit tests before merging.

bertsky: > As-is the reports are crude but well-formed XML which can be used.

Oh, right, I did not think of this. Then maybe you want to allow writing the reports into files as well (like dinglehopper, perhaps even registered in METS)?

kba: > like dinglehopper, perhaps even registered in METS

But the point of this is to have all the invidiual validation methods in one place instead of workspace validation (for which I would consider such a parameter if I there is demand).

> Then maybe you want to allow writing the reports into files as well

Very little advantage over shell redirection I think. Basically I'm trying to avoid @tboenig having to change his setup at all :D

bertsky: > But the point of this is to have all the invidiual validation methods in one place instead of workspace validation (for which I would consider such a parameter if I there is demand).

Sorry, I forgot this is already about probing different parts individually. METS does not apply here.

> Very little advantage over shell redirection I think.

Even with the prospect of separating different XML reports into different files (which redirection cannot do easily)?

> Basically I'm trying to avoid @tboenig having to change his setup at all :D

But this is a new interface, isn't it? The existing workflows with validation should be unrelated...

kba: > Even with the prospect of separating different XML reports into different files (which redirection cannot do easily)?

You'd still have to execute once per subcommand so you could redirect everyone of them (not arguing for the sake of arguing here, it's not a problem to implement if I missed a use case).

> But this is a new interface, isn't it? 

Yeah that was just a joke

bertsky: > You'd still have to execute once per subcommand so you could redirect everyone of them

Sorry again, I was too confused to see that simple fact.

> > But this is a new interface, isn't it?
> 
> Yeah that was just a joke

Got it (eventually)!

wrznr:
@wrznr Push!

kba:
I have now tested most of the functionality.

`ocrd validate workspace` and `ocrd validate zip` have been disabled for now, since they are rather complex and are already easily accessible via their resp. subcommands.

`ocrd validate tasks` will now either check just the syntax/executable existence or do full input/output group validation if a workspace is provided.

`ocrd validate parameters` validates whether JSON parameters are valid according to the definition of the resp. tool in the resp. ocrd-tool.json

`ocrd validate tool-json` validates ocrd-tool.json. You can now say `ocrd validate tool-json` in the root of a MP repo instead of `ocrd ocrd-tool ocrd-tool.json validate` which is more efficient and will boost YOUR productivity with 15 characters less to type!

Issue #251, opened 2019-06-21
Using a non-existent file group as input should give an error message


mikegerber:
Using a non-existent file group as input should give an error message:

~~~
cd `mktemp -d`
wget https://ocr-d-repo.scc.kit.edu/api/v1/dataresources/61aeac32-ef0a-443d-820b-5db7bf8fc8a3/data/lohenstein_agrippina_1665.ocrd.zip
unzip lohenstein_agrippina_1665.ocrd.zip
cd lohenstein_agrippina_1665.ocrd/data
ocrd-tesserocr-segment-line   -l DEBUG -m mets.xml -I DOES-NOT-EXIST -O OCR-D-SEG-LINE
~~~

This is a NOP, but should give an error because DOES-NOT-EXIST does not exist!

See also https://github.com/OCR-D/ocrd_tesserocr/issues/47.

bertsky:
So how about it? Should be easy to fix, right?

bertsky:
In addition, already existing file groups as __output__ should be fatal. 

Both, input and output checks should happen **before** running the processor (taking up space and time, overwriting files).

I believe all this is already implemented in `ocrd process` via `task_sequence.run_tasks`, which raises exceptions in both cases. But we need this for the independent CLIs (`decorators.ocrd_cli_wrap_processor` which calls `processor.base.run_processor`) as well!

EEngl52:
@kba can this be closed?

kba:
```
2020-04-23 20:15:29.157 INFO ocrd.workspace_validator - input_file_grp=['DOES-NOT-EXIST'] output_file_grp=['OCR-D-SEG-LINE']                                                
Traceback (most recent call last):
  File "/home/kba/env/py3/bin/ocrd-tesserocr-segment-line", line 11, in <module>
    load_entry_point('ocrd-tesserocr', 'console_scripts', 'ocrd-tesserocr-segment-line')()                                                                                  
  File "/home/kba/env/py3/lib/python3.6/site-packages/click/core.py", line 829, in __call__                                                                                 
    return self.main(*args, **kwargs)
  File "/home/kba/env/py3/lib/python3.6/site-packages/click/core.py", line 782, in main                                                                                     
    rv = self.invoke(ctx)
  File "/home/kba/env/py3/lib/python3.6/site-packages/click/core.py", line 1066, in invoke                                                                                  
    return ctx.invoke(self.callback, **ctx.params)
[...]

Issue #245, opened 2019-06-19
Ability to delete a group


mikegerber:
`ocrd workspace delete-group` could be a command to delete a group/`fileGrp` in the workspace/METS file.

My use case for this:
--------------------------
Currently, I use xmlstarlet to remove a group of files so that I can repeat a script run, i.e. to enable repeated (and almost idempotent) runs of the following:

~~~sh
remove_filegrp() {
  filegrp_use=$1
  mets=$2

  xmlstarlet ed --inplace \
    -N mets=http://www.loc.gov/METS/ \
    -d "//mets:fileGrp[@USE='$filegrp_use']" $mets
}

remove_filegrp OCR-D-SEG-LINE mets.xml
ocrd-ocropy-segment -l DEBUG -m mets.xml -I OCR-D-IMG -O OCR-D-SEG-LINE
~~~

bertsky:
I see the need, too. 

It is also already possible to use the CLI's `workspace backup` facility for this, but you have to be careful to create a backup beforehand, and `undo` does not always pick the right version, and `restore` sometimes fails because the checksum is not unique (and then requires a full backup filename).

But I generally would prefer to have an option for __overwriting existing groups and files__ instead.

bertsky:
@mikegerber BTW you should also remove those file IDs in the `structMap/div/fptr` references. Here is an additional XPath expression to have xmlstarlet delete (_before_ the actual removal):
```XML
//mets:structMap//mets:fptr[@FILEID=//mets:fileGrp[@USE='$filegrp_use']/mets:file/@ID]
```

kba:
> I see the need, too.

OK then, `ocrd workspace rm-group` shouldn't be too hard to implement. Should this behave like `rmdir` and only delete empty fileGroups or like `rm -r` and also delete non-empty fileGroups?

Probably should also have the equivalent for removing files then, i.e. `ocrd workspace rm`. What do you think?

> But I generally would prefer to have an option for overwriting existing groups and files instead.

Like an option `--overwrite`/`--force` for `ocrd workspace add`?

kba:
> Like an option --overwrite/--force for ocrd workspace add?

We already have `--force` ("If file with ID already exists, replace it"). What behavior changes do you want here?

bertsky:
> > Like an option --overwrite/--force for ocrd workspace add?
> 
> We already have `--force` ("If file with ID already exists, replace it"). What behavior changes do you want here?

[...]

PR #392, opened 2020-01-07
Validate input group

kba:
Adds a static method `check_file_grp` to WorkspaceValidator which is called before run_processor when wrapping processors and will fail if
- input file groups are NOT in mets
- output file groups ARE in mets

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -95,7 +95,7 @@ def run_cli(
     log.debug("Running subprocess '%s'", ' '.join(args))
     return subprocess.call(args)
 
-class Processor(object):
+class Processor():
     """
     A processor runs an algorithm based on the workspace, the mets.xml in the
     workspace (and the input files defined therein) as well as optional
--- a/ocrd_validators/ocrd_validators/workspace_validator.py
+++ b/ocrd_validators/ocrd_validators/workspace_validator.py
@@ -26,18 +26,20 @@ class WorkspaceValidator():
 
     @staticmethod
     @contextmanager
-    def check_file_grp(workspace, input_file_grp, output_file_grp):
+    def check_file_grp(workspace, input_file_grp=None, output_file_grp=None):
         """
-        Return a report on whether input_file_grp is/are in workspace.mets
+        Return a report on whether input_file_grp is/are in workspace.mets and output_file_grp is/are not.
         To be run before processing
         """
         report = ValidationReport()
-        for grp in input_file_grp.split(','):
-            if grp not in workspace.mets.file_groups:
-                report.add_error("Input fileGrp[@USE='%s'] not in METS!" % grp)
-        for grp in output_file_grp.split(','):
-            if grp in workspace.mets.file_groups:
-                report.add_error("Output fileGrp[@USE='%s'] already in METS!" % grp)
+        if input_file_grp:
+            for grp in input_file_grp.split(','):
+                if grp not in workspace.mets.file_groups:
+                    report.add_error("Input fileGrp[@USE='%s'] not in METS!" % grp)
+        if output_file_grp:
+            for grp in output_file_grp.split(','):
+                if grp in workspace.mets.file_groups:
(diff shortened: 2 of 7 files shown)


bertsky:
I can see this fixes #251, but doesn't this also clash with #245 in that even if we will in some future version have an `--overwrite` option for processors (which makes `force=True` the default for `OcrdMets.add_file`), we would now run into the output filegrp exception?

kba:
> we would now run into the output filegrp exception?

True. I added the output file group restriction for consistency. Questionable how helpful this is if we don't offer a way to remedy the situation as of now (i.e. no clean "delete filegrp" feature).

What do you propose? Make the output_check configurable via param? Make it a notice? Omit it altogether?

bertsky:
> What do you propose? Make the output_check configurable via param? Make it a notice? Omit it altogether?

Yes, in the simplest case, you split your new subvalidator into 2 distinct checks (which can be skipped independently). This would give the user full control. Then in the task parser, when we have the `overwrite` option, you can disable the output check.

kba:
> Yes, in the simplest case, you split your new subvalidator into 2 distinct checks

OK, the `input_file_grp` and `output_file_grp` are now optional, if not provided or set to a False-ish value (`None`, empty string etc) the resp. checks are skipped. I've added a TODO to the cli decorator that once we have an `--overwrite` option, we should implement the output_file_grp conditionally.

I haven't changed anything in the task parser yet since these checks are not trivial (input/output filegroup defaults have to be taken into account e.g.)

kba:
> I haven't changed anything in the task parser yet since these checks are not trivial (input/output filegroup defaults have to be taken into account e.g.) 

I've improved the task sequence validation. Still doesn't take default input/output file groups into account, these matter mostly for the secondary outputs.

tests/test_task_sequence.py
```diff
@@ -116,5 +117,18 @@ def test_required_param(self):
...
+
+    def test_validate_sequence(self):
+        resolver = Resolver()
+        with TemporaryDirectory() as tempdir:
+            workspace = resolver.workspace_from_url(assets.path_to('kant_aufklaerung_1784/data/mets.xml'), dst_dir=tempdir)
+            params_path = Path(tempdir, 'params.json')
+            params_path.write_text('{"param1": true}')
+            with self.assertRaisesRegex(Exception, 'Input file group not contained in METS or produced by previous steps: FOO'):
+                validate_tasks([ProcessorTask.parse(x) for x in [
+                    'sample-processor-required-param -I IN -O OUT -p %s' % params_path,
```
kba: oops, this should also raise an exception as `IN` isn't in the METS

ocrd/ocrd/processor/base.py
```diff
@@ -95,7 +95,7 @@ def run_cli(
     log.debug("Running subprocess '%s'", ' '.join(args))
     return subprocess.call(args)
 
-class Processor(object):
+class Processor():
```
wrznr: Interesting. Is a change in some Python best practice causing the removal of the inheritance?

ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -23,6 +24,32 @@ class WorkspaceValidator():
...
+        To be run before processing
+        """
+        if not report:
+            report = ValidationReport()
+        if isinstance(input_file_grp, str):
+            input_file_grp = input_file_grp.split(',')
+        if isinstance(output_file_grp, str):
+            output_file_grp = output_file_grp.split(',')
+
+        print('hi')
```
wrznr: Pls. remove.

wrznr: Before merging!

ocrd/ocrd/processor/base.py
kba: It's the default in Python 3, no more old-style classes. It's a remnant of when we supported python 2.7

ocrd_validators/ocrd_validators/workspace_validator.py
kba: hooray for code review, thanks.

PR #396, opened 2020-01-10
PNG with DPI have no tag, fix #395

### LINKED ISSUE #395: OcrdExif: PNG has no tag for dpi unit!
When I open a PNG image file which includes DPI meta-data, (which is supported since 2014 by Pillow but still undocumented), core currently gives me:
```
  File "ocrd_modelfactory/__init__.py", line 77, in page_from_file
    return page_from_image(input_file)
  File "ocrd_modelfactory/__init__.py", line 49, in page_from_image
    exif = exif_from_filename(input_file.local_filename)
  File "ocrd_modelfactory/__init__.py", line 33, in exif_from_filename
    ocrd_exif = OcrdExif(pil_img)
  File "ocrd_models/ocrd_exif.py", line 42, in __init__
    self.resolutionUnit = 'cm' if img.tag.get(296) == 3 else 'inches'
AttributeError: 'PngImageFile' object has no attribute 'tag'
```

And that's exactly right: That tag is only valid for JPEG. Pillow's PNG plugin converts the fixed implicit _meter_ unit to a fixed implicit _inch_ [here](https://github.com/python-pillow/Pillow/blob/ef4a0b2f4c9346db37140f102e80068abc280167/src/PIL/PngImagePlugin.py#L430).

kba:
The fix is trivial, the complex thing was getting the test metadata right.

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_exif.py
+++ b/ocrd_models/ocrd_models/ocrd_exif.py
@@ -41,6 +41,8 @@ class OcrdExif():
             self.yResolution = int(img.info['dpi'][1])
             if img.format == 'TIFF':
                 self.resolutionUnit = 'cm' if img.tag.get(296) == 3 else 'inches'
+            else:    
+                self.resolutionUnit = 'inches'
         elif img.format == 'JPEG' and 'jfif_density' in img.info:
             self.xResolution = img.info['jfif_density'][0]
             self.yResolution = img.info['jfif_density'][1]
--- a/ocrd_models/ocrd_models/ocrd_exif.py
+++ b/ocrd_models/ocrd_models/ocrd_exif.py
@@ -39,7 +39,8 @@ class OcrdExif():
         if img.format in ('TIFF', 'PNG') and 'dpi' in img.info:
             self.xResolution = int(img.info['dpi'][0])
             self.yResolution = int(img.info['dpi'][1])
-            self.resolutionUnit = 'cm' if img.tag.get(296) == 3 else 'inches'
+            if img.format == 'TIFF':
+                self.resolutionUnit = 'cm' if img.tag.get(296) == 3 else 'inches'
         elif img.format == 'JPEG' and 'jfif_density' in img.info:
             self.xResolution = img.info['jfif_density'][0]
             self.yResolution = img.info['jfif_density'][1]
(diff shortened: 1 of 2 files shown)


ocrd_models/ocrd_models/ocrd_exif.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, img):
         if img.format in ('TIFF', 'PNG') and 'dpi' in img.info:
             self.xResolution = int(img.info['dpi'][0])
             self.yResolution = int(img.info['dpi'][1])
-            self.resolutionUnit = 'cm' if img.tag.get(296) == 3 else 'inches'
+            if img.format == 'TIFF':
+                self.resolutionUnit = 'cm' if img.tag.get(296) == 3 else 'inches'
```
bertsky:
```suggestion
                self.resolutionUnit = 'cm' if img.tag.get(296) == 3 else 'inches'
            else:    
                self.resolutionUnit = 'inches'
```

bertsky: Also, you still have to set `self.resolutionUnit = 'inches'` down below in the other `PNG` case. (Sorry, cannot make suggestions outside of hunk context.

bertsky: Another (minor) thing: the docstring wrongly describes that attribute as `resolutionUnits` (plural)

tests/model/test_exif.py
```diff
@@ -27,17 +29,19 @@ def test_tiff(self):
...
+        with Image.open(assets.path_to('kant_aufklaerung_1784-binarized/data/OCR-D-IMG-BIN/BIN_0020.png')) as img:
+            exif = OcrdExif(img)
         self.assertEqual(exif.width, 1457)
         self.assertEqual(exif.height, 2084)
-        self.assertEqual(exif.xResolution, 300)
-        self.assertEqual(exif.yResolution, 300)
-        self.assertEqual(exif.resolution, 300)
+        self.assertEqual(exif.xResolution, 295)
+        self.assertEqual(exif.yResolution, 295)
+        self.assertEqual(exif.resolution, 295)
```
bertsky:
```suggestion
        self.assertEqual(exif.resolution, 295)
        self.assertEqual(exif.resolutionUnit, 'inches')
```

bertsky: (And maybe the check for the unit attribute can also be extended to the other test cases.)

PR #398, opened 2020-01-11
LGTM: Update configuration because of renamed generated file

stweil:

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,6 +1,6 @@
 path_classifiers:
   generated:
-    - "ocrd/model/ocrd_page_generateds.py"
+    - "ocrd_models/ocrd_models/ocrd_page_generateds.py"
 
 queries:
   - exclude: py/import-and-import-from


kba:
What is the advantage of LGTM.com over pylint? I tend to cleanup unused imports, trailing whitespace and the like from time to time with pylint/flake8/black. They are easy to integrate and Open Source. Nothing against static code analysis but LGTM.com, at least for python, is just not worth the effort of having to sift through the documentation, learn their prorietary query language.

stweil:
Personally I use LGTM often, but pylint rarely, so I cannot say much about the pros and cons of both. LGTM is very easy to use: registration of some Open Source software is easy, in many cases no configuration is needed, nor do I need the documentation.

As far as I have seen, most or even all remaining issues are related to the generated code, so after this PR there remains no noise in the LGTM report.

stweil:
You could also add LGTM as CI tool which checks each pull request.

kba:
> You could also add LGTM as CI tool which checks each pull request.

Again, this might work or it might not and you have to read the docs and debug the setup. Scrutinizer also has CI/testing capability but getting this to work and work right is where stuff gets tricky.

> I use LGTM often, but pylint rarely

For me it's the opposite. It would be easy to run pylint in our CI setup either with travis or circle. If you want to get started, try [our configuration file](https://github.com/OCR-D/core/blob/master/.pylintrc). I don't see the point in advertising their services with badge and in PR if they don't offer any advantage over pylint.

stweil:
@kba, this PR does not add a badge or CI, but fixes the existing LGTM configuration which was necessary because the generated code was moved.

It helps me if you merge it.

If you don't want to merge it, I suggest to remove the old configuration in .lgtm.yml or even the whole file.

kba:
> LGTM will nevertheless continue to analyse this project, but show lot of issues in the generated code.

No point in LGTM analyzing the generateDS code. I'll merge it.

stweil:
LGTM now rates the code with A+: https://lgtm.com/projects/g/OCR-D/core/.

Maybe we should add the badge again. :-)

kba:
I especially like how well our Javascript scores ;-)

PR #399, opened 2020-01-13
remove image caching in workspace, fix #377

### LINKED ISSUE #377: Disable caching in resolver
With `AlternativeImage`, line images etc. the trivial caching mechanism in the resolver is getting overwhelmed leading to potential out-of-memory or out-of-open-filehandles issues.

Probably best to disable any caching and make sure files are closed properly.

kba:
Fix out-of-filehandles/memory leaks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Fixed:
+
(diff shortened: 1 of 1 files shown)


kba:
Happy to hear 😌 

> The use case Konzilsprotokolle

I've updated the gist here https://gist.github.com/kba/c568137a01cae7dffa9beffb24ee3845 Hopefully we can simplify this further soon with the fixed remove/add fileGrp feature.

PR #419, opened 2020-01-22
Default to "."/"mets.xml" for ocrd workspace init/validate/clone, fix #401

### LINKED ISSUE #401: "ocrd workspace init" should have current working directory as default
I think it would make sense to make

```
ocrd workspace init
```

equivalent to 

```
ocrd workspace init .
```

kba:


bertsky:
I must say I never understood why these commands required a `METS_URL` at all – they all get passed the workspace context, and I have to use `-d directory` and `-M basename` for the other workspace commands anyway.

kba:
> I must say I never understood why these commands required a DIRECTORY / METS_URL

Organic growth of the software according to shifting requirements. Once crunch time is over we should consider cleaning this up. But currently I would not want to break anything if I can avoid it.

PR #420, opened 2020-01-22
:fire: ocrd-tool CLI: simplify interface

kba:
Low priority and needs more work to stay compatible with bashlib processors. But since it's related to #419, I'm opening this PR lest we forget.


kba:
Obsolete. The interface is a bit unintuitive but breaking existing deployments because of it is unreasonable. Also there is the `ocrd validate tool-json` command for validation.

Issue #324, opened 2019-09-24
workspace validate creates a. TEMP directory with files in it b. copies of workspace files


mikegerber:
Before:

~~~
% ls
mets.xml  OCR-D-GT-SEG-BLOCK  OCR-D-GT-SEG-LINE  OCR-D-GT-SEG-PAGE  OCR-D-IMG  OCR-D-IMG-BIN  OCR-D-OCR-CALAMARI  OCR-D-OCR-TESS  OCR-D-SEG-LINE  OCR-D-SEG-REGION
~~~

Then:
~~~
% ocrd workspace validate mets.xml
16:42:08.319 INFO ocrd.resolver.download_to_directory - directory=|/srv/data/OCR-D/justi_abhandlung01_1758.ocrd/data| url=|mets.xml| basename=|mets.xml| if_exists=|raise| subdir=|None|
16:42:08.325 INFO ocrd.resolver.download_to_directory - directory=|/srv/data/OCR-D/justi_abhandlung01_1758.ocrd/data| url=|OCR-D-IMG/OCR-D-IMG_0001.tif| basename=|OCR-D-IMG_0001.tif| if_exists=|skip| subdir=|OCR-D-IMG|
16:42:08.330 INFO ocrd.resolver.download_to_directory - directory=|/srv/data/OCR-D/justi_abhandlung01_1758.ocrd/data| url=|OCR-D-IMG/OCR-D-IMG_0002.tif| basename=|OCR-D-IMG_0002.tif| if_exists=|skip| subdir=|OCR-D-IMG|
16:42:08.331 INFO ocrd.resolver.download_to_directory - directory=|/srv/data/OCR-D/justi_abhandlung01_1758.ocrd/data| url=|OCR-D-IMG/OCR-D-IMG_0003.tif| basename=|OCR-D-IMG_0003.tif| if_exists=|skip| subdir=|OCR-D-IMG|
16:42:08.332 INFO ocrd.resolver.download_to_directory - directory=|/srv/data/OCR-D/justi_abhandlung01_1758.ocrd/data| url=|OCR-D-IMG/OCR-D-IMG_0004.tif| basename=|OCR-D-IMG_0004.tif| if_exists=|skip| subdir=|OCR-D-IMG|
[...]

PR #423, opened 2020-01-22
workspace validation: clarify why _validate_dimension requires --download

kba:
Debugging #324 I understand now that the `TEMP/*` files are being created for the AlternativeImages in PAGE. Even if the PAGE is available locally, the images might not be. Even without `--download`, these images were loaded.

Instead of further complicating the logic for image resolving, this PR requires `--download` for the `_validate_dimensions` check. This is too strict for PAGE files with only local image references or w/o any image references.

### DIFF
--- a/ocrd_validators/ocrd_validators/workspace_validator.py
+++ b/ocrd_validators/ocrd_validators/workspace_validator.py
@@ -166,8 +166,8 @@ class WorkspaceValidator():
         Validate image height and PAGE imageHeight match
         """
         for f in self.mets.find_files(mimetype=MIMETYPE_PAGE):
-            if not f.local_filename and not self.download:
-                self.report.add_notice("Won't download remote PAGE XML <%s>" % f.url)
+            if not self.download:
+                self.report.add_notice("_validate_dimension: Not executed because --download wasn't set and PAGE might reference remote (Alternatve)Images <%s>" % f.url)
                 continue
             page = page_from_file(f).get_Page()
             _, _, exif = self.workspace.image_from_page(page, f.pageId)
--- a/tests/validator/test_workspace_validator.py
+++ b/tests/validator/test_workspace_validator.py
@@ -168,11 +168,15 @@ class TestWorkspaceValidator(TestCase):
             copytree(assets.path_to('kant_aufklaerung_1784/data'), wsdir)
             with pushd_popd(wsdir):
                 os.system("""sed -i 's,imageHeight="2083",imageHeight="1234",' OCR-D-GT-PAGE/PAGE_0017_PAGE.xml""")
-                report = WorkspaceValidator.validate(self.resolver, join(wsdir, 'mets.xml'), src_dir=wsdir, skip=[
-                    'page', 'mets_unique_identifier', 'mets_file_group_names', 'mets_files', 'pixel_density', 'imagefilename'
-                    ], download=False)
+                report = WorkspaceValidator.validate(
+                    self.resolver,
+                    join(wsdir, 'mets.xml'),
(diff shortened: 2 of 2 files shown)


cneud:
> Even if the PAGE is available locally, the images might not be.

When/why could this happen? AFAICT no one ever besides Aletheia and OCR-D uses `AlternativeImage`, so any images referenced from PAGE should always be available locally - either from OCR-D GT or as a direct result of OCR-D processing? 😕

kba:
> When/why could this happen?

E.g. https://github.com/OCR-D/assets/blob/master/data/kant_aufklaerung_1784-binarized/data/OCR-D-GT-WORD/INPUT_0017.xml

cneud:
Thanks @kba. I assume we have to eat our own dog food then.

kba:
> eat our own dog food then.

💩 ⍩⍩⍩⍩

bertsky:
> this PR requires `--download` for the `_validate_dimensions` check.

The change itself looks correct: it's not just the PAGE file itself that is need (and therefore must be local), but also its images. So instead of comparing the original and derived image references to the METS and verifying they are all local, you don't (want to) bother with the file status anymore and just require `--download` altogether.

> This is too strict for PAGE files with only local image references or w/o any image references.

True, now we chicken out of the test too easily. This leaves the user with but 2 options: using `--download` for everything or having no size validation even for local files.

Alternatives:
1. Don't verify size dimensions of derived images at all, only of the original image: Take `resolve_image_as_pil(page.imageFilename)` instead of `image_from_page(page)`. This one image file is probably worth the effort of METS lookup.
2. Add some kwarg `download=True` to `resolve_image_as_pil()`, so the download can be disabled. Then either use it with 1. or expose it in `image_from_page()` as well.
3. Live with the recursive download for files the user is interested in, but fix the `TEMP` issue independently. (I still don't understand why missing files go there and not under their fileGrp, and why existing file are downloaded _again_.)

kba:
> Don't verify size dimensions of derived images at all, only of the original image: Take resolve_image_as_pil(page.imageFilename) instead of image_from_page(page). This one image file is probably worth the effort of METS lookup.

That wouldn't be too much effort but do we really want to not verify dimensions of derived images? I mean it's a pretty trivial test so it's not that meaningful to extend it to derived images but it still feels wrong.

> Add some kwarg download=True to resolve_image_as_pil(), so the download can be disabled. Then either use it with 1. or expose it in image_from_page() as well.

That's what I wanted to avoid, the workspace methods are complicated enough as-is without additional if/else clauses.

> I still don't understand why missing files go there and not under their fileGrp

That's a bug I'm still investigating. `TEMP` is chosen *as the fileGrp* for files where we don't know the file group. It's a flaw in the logic where an OcrdFile is created from url/local_filename but not looked up in METS before downloading.

> why existing file are downloaded again

Aside from the `TEMP` bug, this should not happen anymore since #404. If it does, please open an issue with test data and I'll investigate.

bertsky:
> That wouldn't be too much effort but do we really want to not verify dimensions of derived images? I mean it's a pretty trivial test so it's not that meaningful to extend it to derived images but it still feels wrong.

Yes, you are right.

> That's what I wanted to avoid, the workspace methods are complicated enough as-is without additional if/else clauses.

I concur.

> That's a bug I'm still investigating. `TEMP` is chosen _as the fileGrp_ for files where we don't know the file group. It's a flaw in the logic where an OcrdFile is created from url/local_filename but not looked up in METS before downloading.

Oh, I see.

> Aside from the `TEMP` bug, this should not happen anymore since #404.

Now I got it. Thanks for clarifying!

PR #425, opened 2020-01-22
Clobber mets download to directory

kba:
This changes the behavior: if `clobber_mets` is `False` and METS is to be downloaded but already exists in the destination: **skip downloading** instead of **rasing an exception**.

Translating between `clobber_mets` and `if_exists` is not a good solution. The former should be replaced with the latter.

### DIFF
--- a/ocrd/ocrd/resolver.py
+++ b/ocrd/ocrd/resolver.py
@@ -151,7 +151,7 @@ class Resolver():
         log.debug("workspace_from_url\nmets_basename='%s'\nmets_url='%s'\nsrc_baseurl='%s'\ndst_dir='%s'",
             mets_basename, mets_url, src_baseurl, dst_dir)
 
-        self.download_to_directory(dst_dir, mets_url, basename=mets_basename, if_exists='overwrite' if clobber_mets else 'raise')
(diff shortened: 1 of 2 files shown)


kba:
> I do not understand the use-cases for clobber_mets=False, and I know of no example of it

The original idea was that you pass a METS URL which is only downloaded the first time it is to be used. Subsequent processors should not overwrite the METS every time they process a workspace.

bertsky:
> > I do not understand the use-cases for clobber_mets=False, and I know of no example of it
> 
> The original idea was that you pass a METS URL which is only downloaded the first time it is to be used. Subsequent processors should not overwrite the METS every time they process a workspace.

This explanation puzzles me. For all I know, processors must _always_ modify the METS (unless they have no output). And this _is_ about overwriting, not about downloading, isn't it?

kba:
> For all I know, processors must always modify the METS

Yes but the METS might not be downloaded yet. That's why we introduced the idea of a workspace which was initially just a crutch to make it easier for processors and we were careful not to use it as a first-class concept e.g. in our specs. Instead of handling the intricacies of downloading full/partial representations of the work, they now just had to act on a directory and files. Since then however, workspaces have become the standard way to interact with METS/Images/PAGE and most if not all the facilities to handle non-local data outside `ocrd workspace` has become obsolete.

After the final workshop we'll have to generously delete and straighten these remnants.

PR #428, opened 2020-01-23
ocrd workspace add-group to add folder of files in bulk, fix #416

### LINKED ISSUE #416: Bulk mode for ocrd workspace add
@stweil asked in #409 

> Is it already possible to add all files of a directory as a new group, or would that require an add-group command?

It is not but it's a desirable feature.

kba:
Allows adding files in bulk. Very rigid convention:

```sh
$ find IMAGE-FOLDER
IMAGE-FOLDER/page1.png
$ ocrd workspace add-group -m foo/bar IMAGE-FOLDER
```

will create
```xml
[...]
<mets:div TYPE="page" ID="page1">` <!-- in the physical structmap -->
[...]
<mets:fileGrp USE="IMAGE-FOLDER">
    <mets:file MIMETYPE="foo/bar" ID="IMAGE-FOLDER_page1">
        <mets:FLocat LOCTYPE="OTHER" OTHERLOCTYPE="FILE" xlink:href="IMAGE-FOLDER/page1.png"/>
    </mets:file>
</mets:fileGrp>
```


ocrd/ocrd/cli/workspace.py
```diff
@@ -159,6 +167,54 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, local_f
...
+    """
+    workspace = Workspace(ctx.resolver, directory=ctx.directory, mets_basename=ctx.mets_basename, automatic_backup=ctx.automatic_backup)
+
+    file_grp_dir = Path(file_grp_dir).resolve()
+
+    ws_dir = Path(ctx.directory).resolve()
+    print(file_grp_dir)
+
+    # XXX will raise ValueError if file_grp_dir isn't relative to ctx.directory
+    file_grp_dir.relative_to(ws_dir)
```
bertsky: Why `XXX`? Isn't that exception a good thing? If you think otherwise, maybe you can check explicitly and first copy the files into the workspace.

ocrd/ocrd/cli/workspace.py
```diff
@@ -11,6 +13,12 @@
 
 log = getLogger('ocrd.cli.workspace')
 
+def clean_id(dirty):
+    if re.match('^[0-9]', dirty):
+        raise ValueError("Make sure files and directories do not begin with a numeral which will lead to invalid XML identifiers")
+    return re.sub('[^A-Za-z0-9-_]+', '_', dirty)
```
bertsky: I didn't know this restriction of XML ids. (And you sure colons and dots are out?)

How about moving this from ocrd.cli.workspace to ocrd_utils?

ocrd/ocrd/cli/workspace.py
```diff
@@ -159,6 +167,54 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, local_f
...
+    print(file_grp_dir)
+
+    # XXX will raise ValueError if file_grp_dir isn't relative to ctx.directory
+    file_grp_dir.relative_to(ws_dir)
+    file_grp = clean_id(file_grp_dir.name)
+
+    for f in file_grp_dir.iterdir():
+        page_id = f.name
+        if f.suffix:
+            page_id = page_id[0:-len(f.suffix)]
```
bertsky: Shouldn't we at least try to map these onto existing pages in the structMap? You could try to (remove a common prefix and) find substrings shared by existing pageIds systematically.

Scenario: `add-group` for images followed by `add-group` for PAGE-XML files.

ocrd/ocrd/cli/workspace.py
```diff
@@ -159,6 +167,54 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, local_f
...
+        fileGrp ID = last folder name component
+        page ID = filename sans extension
+        file ID = sanitized concatenation of fileGrp ID and page ID
+    """
+    workspace = Workspace(ctx.resolver, directory=ctx.directory, mets_basename=ctx.mets_basename, automatic_backup=ctx.automatic_backup)
+
+    file_grp_dir = Path(file_grp_dir).resolve()
+
+    ws_dir = Path(ctx.directory).resolve()
+    print(file_grp_dir)
```
bertsky: no logging? (Also recommend logging file ID and pageId of every file to be added in the loop.)

ocrd/ocrd/cli/workspace.py
```diff
@@ -159,6 +167,54 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, local_f
...
+
+    # XXX will raise ValueError if file_grp_dir isn't relative to ctx.directory
+    file_grp_dir.relative_to(ws_dir)
+    file_grp = clean_id(file_grp_dir.name)
```
bertsky: maybe validate that fileGrp does not exist yet?

ocrd/ocrd/cli/workspace.py
```diff
@@ -159,6 +167,54 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, local_f
...
     workspace.save_mets()
 
+# ----------------------------------------------------------------------
+# ocrd workspace add-group
+# ----------------------------------------------------------------------
+
+@workspace_cli.command('add-group')
+@click.option('-m', '--mimetype', help="Media type of the file", required=True)
+@click.option('-f', '--force', help="Continue even if mets:file same ID already exists", default=False, is_flag=True)
+@click.argument('file_grp_dir', type=click.Path(dir_okay=True, file_okay=False, readable=True, resolve_path=True), required=True)
```
bertsky: What if the directory contains other kinds of files (e.g. thumbnails for images, or backup copies, or a mix of different file types)? Wouldn't it be a better interface to call with a glob pattern which the shell expands to a list of files?

ocrd/ocrd/cli/workspace.py
kba: The `XXX comment` is just there to make future developers aware that this might raise an exception and the statement is only there for this side effect.

> first copy the files into the workspace.

I would prefer to avoid this. Previous attempts to smartly handle files have backfired. Do you have a use case where it would simplify your life if `ocrd` took care of copying the files instead of you doing it explicitly before?

ocrd/ocrd/cli/workspace.py
kba: > I didn't know this restriction of XML ids. (And you sure colons and dots are out?)

METS defines it's `ID` as `xsd:int` which is a restriction on `xsd:NCName` which requires

* first character to be `[A-Za-z]`
* no colons

We ran into this before and it's pretty confounding if you don't know what to look for (in the cases I ran into this, culprit was always purely numeric file names).

Dots would be possible, I tend to avoid them because they're often part of the syntax of query languages like CSS selectors.

ocrd/ocrd/cli/workspace.py
kba: > Scenario: add-group for images followed by add-group for PAGE-XML files.

The assumption here is that file (base) names would be the same for different file groups. But I agree that this assumption is too restrictive. I think the convention of many libraries seems to be to have a file-type specific suffix to the file ID (e.g. `FILE_0009_THUMBS`) and page IDs that begin with `PHYS`, such as `PHYS_0009`. So it gets tricky fast with a common convention.

How about allowing users to define regular expressions for searching and template strings for creating ID on the command line. Something like

```
ocrd workspace add-group --file-id="s/.*(\d{4}).*/FILE_\1_DEFAULT" --page-id="s/.*(\d{4}).*/PHYS_\1/" ...
```
```

File: ocrd/ocrd/cli/workspace.py (hunk 4 above)
Comment (kba) [reply ocrd/ocrd/cli/workspace.py]: That was just a debug stmt. Logging is indeed missing, I'll add some.

File: ocrd/ocrd/cli/workspace.py (hunk 6 above)
Comment (kba) [reply ocrd/ocrd/cli/workspace.py]: I fear that users will fail to quote glob calls properly. But if I'm arguing for regexes, I can't argue against globs.

File: ocrd/ocrd/cli/workspace.py (hunk 2 above)
Comment (bertsky) [reply ocrd/ocrd/cli/workspace.py]: > How about moving this from ocrd.cli.workspace to ocrd_utils?

This should be checked in regular `add_file()` as well IMO.

File: ocrd/ocrd/cli/workspace.py (hunk 3 above)
Comment (bertsky) [reply ocrd/ocrd/cli/workspace.py]: Yes, that would be a good start. Regexps are complicated but easy and safe to parse.

But I think we should _also_ offer some kind of default which guesses these as best as we can.

File: ocrd/ocrd/cli/workspace.py [hunk 7]
```diff
@@ -159,6 +167,54 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, local_f
...
+    file_grp = clean_id(file_grp_dir.name)
+
+    for f in file_grp_dir.iterdir():
+        page_id = f.name
+        if f.suffix:
+            page_id = page_id[0:-len(f.suffix)]
+        page_id = clean_id(page_id)
+        file_id = clean_id('_'.join([file_grp, page_id]))
+        local_filename = str(f.relative_to(ws_dir))
+        workspace.mets.add_file(
```
bertsky: Another point (from #446): If the directory is large, then we should not use the regular `add_file()` here, which tries to find the ID and then adds it to `workspace.mets._file_by_id` dict: this creates considerable memory overhead and slows down. 

Instead, here we should check in bulk that the file_ids we want to add do not yet exist, and then add to the workspace without the dict mechanism.

ocrd/ocrd/cli/workspace.py
kba: https://github.com/OCR-D/core/issues/447

ocrd/ocrd/cli/workspace.py
kba: Probably best to fix #446 first then.

bertsky: Agreed. But still, avoiding the `self.find_files(ID=ID)` test in each iteration would make this more efficient. Perhaps by adding an option `ignore=False` complementing `force=False`?

ocrd/ocrd/cli/workspace.py
bertsky: I think I actually meant globbing by the shell, so you could also just pass a space-delimited list of path names.

IMO we have 2 principle routes:
- pattern for both file names and page IDs controlled by regex argument
- shell globbing or other pathname expansion, and guesswork/heuristics on our side

ocrd/ocrd/cli/workspace.py
kba: #500

ocrd/ocrd/cli/workspace.py
kba: Copying has now been implemented. The `--url` option (after expansion) is used as the relative filename.

ocrd/ocrd/cli/workspace.py
kba: logging added.

ocrd/ocrd/cli/workspace.py
kba: `--overwrite` parameter will remove file groups before adding files. For performance reasons, no checks on the actual METS currently implemented. But `--force` will be passsed on to `OcrdMets.add_file`.

ocrd/ocrd/cli/workspace.py
kba: `ignore` is a good idea, implemented in 35289a13

ocrd/ocrd/cli/workspace.py
kba: files can be passed directly or as a glob expression expanded by python's `glob` module.

the pattern to match files against is passed as `--regex`. I'll post an example forthwith.

ocrd/ocrd/cli/workspace.py
kba: I've rewritten the implementation. Here's an example how this can be used:

Let's assume you have directories `path/to/files/PAGE/` containing PAGE files and `path/to/files/IMG` with images. The files have a basename `page_0001.xml`, ``page_0001.tif'` etc.

```sh
ocrd workspace bulk-add \
    --regex '^.*/(?P<fileGrp>[^/]+)/page_(?P<pageid>.*)\.(?P<ext>[^\.]*)$' \
    --file-id 'FILE_{{ fileGrp }}_{{ pageid }}' \
    --page-id 'PHYS_{{ pageid }}' \
    --file-grp "{{ fileGrp }}" \
    --url '{{ fileGrp }}/FILE_{{ pageid }}.{{ ext }}' \
    'path/to/files/*/*.*'
```

This will first expand the glob to get filenames and resolve them to absolute paths.

Every path is then matched against `--regex` with `re.match`, yielding template variables derived from the syntax of the path. These template variables can be used in all file-specific options. `--url` after expansion is used as the filename relative to the workspace directory and copied into the workspace if not already present. After expanding all template variables, the file is added with `Workspace.add_file`.

In this case:
  * `path/to/files/PAGE/page_0001.xml` ->
    * `url`: `PAGE/FILE_0001.xml` (which will be copied unless `path/to/files` **is** the workspace dir)
    * `fileGrp`: `PAGE`
    * `ID`: `FILE_0001`
    * `pageId`: `PHYS_0001`

`--mimetype`, if not provided, is mapped from the file extension.

`--ignore` will disable the check for existing files with the same `@ID` and is a huge performance boost.

It may not be the prettiest solution but it is very versatile, as long as all the information can be derived directly from the file path. If the process is more involved, e.g. using mappings or string transformation methods, we could add a proper template engine like jinja. On the other hand, bulk renaming files with unix tools and fine-tuning the regex is probably easier.

We could also offer or at least document ready-made regex/template combinations for common use cases.

bertsky: This is an excellent idea and solution. It has maximum flexibility but has a clear interface and simple implementation, and is thus most reliable.

>     * (which will be copied unless `path/to/files` **is** the workspace dir)

Not quite: IIUC, files will be copied unless they exist exactly _under (expanded) `url`_ in the workspace directory. So in your example, because `regex` differs from `url`, they will be copied even if `path/to/files` is the workdir.

> On the other hand, bulk renaming files with unix tools and fine-tuning the regex is probably easier.

I agree: more elaborate schemes can easily be constructed with a simple (and fast) shell loop. `bulk-add` only needs to be the last step. 

But there's a catch: I believe one still needs an additional option to allow **skipping** files that don't match the regex. This could be for another run of `bulk-add`, or just for files that we don't care about (presentation, logs, meta-data etc).

For example, to import the old (first-generation zip-file) OCR-D GT directories, one could then do:
```
ocrd workspace bulk-add --skip --regex '^.*/(?P<dispname>[^/]*)/(?P=dispname)_(?P<pageid>[0-9]*)\.tif$' --file-id 'FILE_ORIG_{{ pageid }}' --page-id 'PHYS_{{ pageid }}' --file-grp OCR-D-IMG --url '{{ dispname }}_{{ pageid }}.tif' # only images, no copying
ocrd workspace bulk-add --skip --regex '^.*/(?P<dispname>[^/]*)/page/(?P=dispname)_(?P<pageid>[0-9]*)\.xml$' --file-id 'FILE_GT_{{ pageid }}' --page-id 'PHYS_{{ pageid }}' --file-grp OCR-D-GT-SEG-PAGE --url 'page/{{ dispname }}_{{ pageid }}.xml' # only PAGE, no copying
ocrd workspace bulk-add --skip --regex '^.*/(?P<dispname>[^/]*)/alto/(?P=dispname)_(?P<pageid>[0-9]*)\.xml$' --file-id 'FILE_GT-ALTO_{{ pageid }}' --page-id 'PHYS_{{ pageid }}' --file-grp OCR-D-GT-ALTO-SEG-PAGE --mimetype application/alto+xml --url 'alto/{{ dispname }}_{{ pageid }}.xml' # only ALTO, no copying
```
(You cannot match the non-existing image subdirectory as fileGrp in this convention directly, and breaking it up allows a basic form of string transformation.)

ocrd/ocrd/cli/workspace.py
```diff
@@ -151,10 +162,115 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, fname):
...
+            if not dry_run and not urlpath.exists():
+                if not urlpath.parent.is_dir():
+                    urlpath.parent.mkdir()
+                urlpath.write_bytes(file_path.read_bytes())
+
+        # Honor --overwrite
+        fileGrp = file_dict.pop('fileGrp')
+        if not dry_run and overwrite and fileGrp not in _groups_seen:
+            workspace.remove_file_group(fileGrp, recursive=True, force=True, keep_files=False)
+            _groups_seen.add(fileGrp)
```
bertsky, line 262: I am having trouble with this option. If the user really wanted to delete everything from these groups, she could easily do it with an extra `remove-group` before. But IMHO it would be a really useful extra feature, if instead of removing the complete group we would just overwrite that one file ID on `--overwrite`.

Generally, it would be great if we could think of `bulk-add --skip` as being **incremental**: ignore non-matching files, but also ignore already-existing entries.

ocrd/ocrd/cli/workspace.py
```diff
@@ -151,10 +162,115 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, force, fname):
...
+
+        # Add to workspace (or not)
+        if dry_run:
+            log.info('workspace.add_file(%s)' % file_dict)
+        else:
+            workspace.add_file(fileGrp, ignore=ignore, force=force, **file_dict)
+
+    # save changes to disk
+    workspace.save_mets()
```
bertsky, line 271: This makes me think of another problem: what if at some point bulk-add exits with non-zero status, after already having copied a lot of files to their new location? Then the next time I try, these files already exist in the filesystem, perhaps getting in the way. (It's covered with the proposed `--skip` extension, but maybe there's a better solution which makes METS and FS actions atomic?)

ocrd/ocrd/cli/workspace.py
kba: Since `overwrite_mode` on the workspace now only overrides `force` globally, I don't see the need for `--overwrite` at all anymore. `--force --skip` should be equivalent to `--overwrite --skip`.

ocrd/ocrd/cli/workspace.py
kba: `--skip` does that now, thx to your suggestion in dac4b9f5

bertsky: Where would be a good place to document recipes like that – maybe in the setup guide or workflow guide?

kba: Both make sense, though the information would get a bit lost in the workflow guide. For now, I've [created a page on bulk-add in the ocrd-website wiki](https://github.com/OCR-D/ocrd-website/wiki/Workspace-Bulk-Add) with the recipes we've discussed here.

PR #432, opened 2020-02-04
PAGE validator TextEquiv consistency: first instead of index1

### LINKED ISSUE #430: page_validator.py produces wrong concatenated text
In [get_text()](https://github.com/OCR-D/core/blob/master/ocrd_validators/ocrd_validators/page_validator.py#L260), the `TextEquiv` with `index=1` is used if it exists. The way I read the documentation of the `index` attribute in the [PAGE schema](https://www.primaresearch.org/schema/PAGE/gts/pagecontent/2019-07-15/pagecontent.xsd), it should use the one with the lowest index:

>  Used for sort order in case multiple TextEquivs are defined. The text content with the lowest index should be interpreted as the main text content. 

The lowest possible value for `index` is `0`, according to the schema.

bertsky:
Fixes #430.

Note: renaming `index1` to `first` in the CLI option for the `validate` command is backwards incompatible. But it's only the CLI (not the API), and was just introduced recently, and there were no other choices anyway, so I doubt this is significant.


kba:
Superseded by #432

mikegerber:
Is this unmerged?

Issue #269, opened 2019-08-02
parse fails to validate result of to_xml


bertsky:
I get a regression with 1.0.0b11: The call to `page_from_file` fails at `ocrd_models_generateds.parse` on a file previously generated by `ocrd_models.ocrd_page.to_xml`. (It mocks in `validate_ConfSimpleType` that the value is a `str` instead of a number.)

This is what I did:
```shell
ocrd-asv-ann-evaluate -m $mets -I OCR-D-GT-SEG-LINE,OCR-D-OCR-OCRO-fraktur-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP,OCR-D-OCR-OCRO-frakturjze-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP,OCR-D-OCR-TESS-Fraktur-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP,OCR-D-OCR-TESS-frk-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP,OCR-D-OCR-TESS-frk+deu-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP
```
where all the OCR file grps are from a previous recognize processor in a long chain that runs through ok. See [here](https://github.com/ASVLeipzig/cor-asv-ann/blob/a460bd5a95bd6fa092c40259f99355c3af02f01b/ocrd_cor_asv_ann/wrapper/evaluate.py#L16) for what the processor does.

This is what happens:
```log
16:05:16.373 DEBUG processor.EvaluateLines - adding input file group OCR-D-GT-SEG-LINE to page phys_0001
16:05:16.375 DEBUG processor.EvaluateLines - adding input file group OCR-D-OCR-OCRO-fraktur-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP to page phys_0001
16:05:16.378 DEBUG processor.EvaluateLines - adding input file group OCR-D-OCR-OCRO-frakturjze-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP to page phys_0001
16:05:16.381 DEBUG processor.EvaluateLines - adding input file group OCR-D-OCR-TESS-Fraktur-BINPAGE-wolf-DESKEW-tesserocr-CLIP-RESEG-DEWARP to page phys_0001
[...]

PR #437, opened 2020-02-13
[wip] update generateDS PAGE API

kba:
Fixing #269.

Just a first test to make sure runtime dependencies are met.

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_page.py
+++ b/ocrd_models/ocrd_models/ocrd_page.py
@@ -67,7 +67,7 @@ def to_xml(el):
     """
     sio = StringIO()
     el.export(sio, 0, name_='PcGts',
-            namespacedef_='xmlns:pc="%s"  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="%s %s/pagecontent.xsd"' % (
+            namespacedef_='xmlns="%s" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="%s %s/pagecontent.xsd"' % (
                 NAMESPACES['page'],
                 NAMESPACES['page'],
                 NAMESPACES['page']
--- a/ocrd_models/ocrd_models/ocrd_page_generateds.py
+++ b/ocrd_models/ocrd_models/ocrd_page_generateds.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 #
-# Generated Mon Aug  5 20:18:48 2019 by generateDS.py version 2.30.11.
+# Generated Thu Feb 13 16:47:27 2020 by generateDS.py version 2.35.13.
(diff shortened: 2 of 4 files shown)


kba:
>  the typing of the conf attribute

Good point, I could test that conversion also works as expected when using the `set_*` methods.

> why abandon the pc: namespace prefix now?

No reason TBH, generateDS has yet again changed in this regard and it seemed simpler to use no namespace for PAGE (which also reduces noise) than to find out how to reenable the `pc:` namespace. This was the original behavior IIRC. Do you depend on the namespace prefix anywhere?

kba:
> Good point, I could test that conversion also works as expected when using the set_* methods.

It does **not**, e.g. 

```python
el = pcgts.get_Page().get_TextRegion()[0].get_TextLine()[0].get_Word()[0].get_TextEquiv()[0]
el.set_conf('2.0987')
self.assertTrue(isinstance(el.conf, float))
```

fails. :(

bertsky:
> It does **not**, e.g.
> 
> ```python
> el = pcgts.get_Page().get_TextRegion()[0].get_TextLine()[0].get_Word()[0].get_TextEquiv()[0]
> el.set_conf('2.0987')
> self.assertTrue(isinstance(el.conf, float))
> ```
> 
> fails. :(

Okay, so the setter does not convert, but that only affects the in-memory DOM, not its serialization. Badly written processors which do this might also want to see strings there. As long as the getter does convert, I am still satisfied with it.

bertsky:
> > why abandon the pc: namespace prefix now?
> 
> No reason TBH, generateDS has yet again changed in this regard and it seemed simpler to use no namespace for PAGE (which also reduces noise) than to find out how to reenable the `pc:` namespace. 

I see. 

> This was the original behavior IIRC. Do you depend on the namespace prefix anywhere?

No, not anymore: our xmlstarlet/bashlib-based implementations in ocrd_olena and ocrd_im6convert all make an extra effort to determine the namespace prefix and URI of the input and use it.

kba:
> As long as the getter does convert, I am still satisfied with it.

I don't like that it is possible to serialize invalid data. One advantage of an API over direct XML access is that such errors should lead to instant errors.

bertsky:
> Other modules relying on it, like ocrd-cis-align, ocrd-cor-asv-fst-process and ocrd-cor-asv-ann-process might have to be adapted.

Oh wow, you _did_ merge this rather quickly. Now new installations of the above mentioned processors will fail until we adapt them. (I would have preferred making them robust first...)

kba:
> Now new installations of the above mentioned processors will fail

How so? What has changed implementation-wise? You mean your defensive type-casting?

bertsky:
> How so? What has changed implementation-wise? You mean your defensive type-casting?

That `get_conf()` will now deliver floats instead of strings. IIRC only ocrd_keraslm was cautious enough to allow both possibilities.

kba:
I thought one point of https://github.com/OCR-D/core/issues/269 was to fix this. I had an idea for systematically testing all module projects in ocrd_all, I'll implement it and see what this might have broken.

PR #443, opened 2020-02-19
PAGE API: add __hash__ to generated code

kba:
... so instances of generateDS-generated class can be used as keys in dict etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,13 +77,18 @@ uninstall:
 	for mod in $(BUILD_ORDER);do pip uninstall -y $$mod;done
 
 # Regenerate python code from PAGE XSD
+generate-page: GDS_PAGE = ocrd_models/ocrd_models/ocrd_page_generateds.py
+generate-page: GDS_PAGE_USER = ocrd_models/ocrd_page_user_methods.py
 generate-page: repo/assets
 	generateDS \
 		-f \
 		--root-element='PcGts' \
-		--user-methods='ocrd_models/ocrd_page_user_methods.py' \
-		-o ocrd_models/ocrd_models/ocrd_page_generateds.py \
+		-o $(GDS_PAGE) \
+		--disable-generatedssuper-lookup \
+		--user-methods=$(GDS_PAGE_USER) \
 		repo/assets/data/schema/data/$(PAGE_VERSION).xsd
+	# hack to prevent #451
+	sed -i 's,(Enum):$$,(object):,' $(GDS_PAGE)
 
 #
 # Repos
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,8 @@ generate-page: repo/assets
 		--disable-generatedssuper-lookup \
 		--user-methods=$(GDS_PAGE_USER) \
 		repo/assets/data/schema/data/$(PAGE_VERSION).xsd
-	# hack to prevent #451
-	sed -i 's,(Enum):$$,(object):,' $(GDS_PAGE)
+	# hack to prevent #451: enum keys will be strings
+	sed -i 's/(Enum):$$/(str, Enum):/' $(GDS_PAGE)
 
 #
 # Repos
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ generate-page: repo/assets
 	generateDS \
 		-f \
 		--root-element='PcGts' \
+		--user-methods='ocrd_models/ocrd_page_user_methods.py' \
 		-o ocrd_models/ocrd_models/ocrd_page_generateds.py \
 		repo/assets/data/schema/data/$(PAGE_VERSION).xsd
 
--- a/ocrd_models/ocrd_models/ocrd_page_generateds.py
+++ b/ocrd_models/ocrd_models/ocrd_page_generateds.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 #
-# Generated Thu Apr 23 14:47:51 2020 by generateDS.py version 2.35.20.
+# Generated Thu Apr 23 15:06:47 2020 by generateDS.py version 2.35.20.
 # Python 3.6.9 (default, Apr 18 2020, 01:56:04)  [GCC 8.4.0]
 #
 # Command line options:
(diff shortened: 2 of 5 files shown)


wrznr:
# [Codecov](https://codecov.io/gh/OCR-D/core/pull/443?src=pr&el=h1) Report
> Merging [#443](https://codecov.io/gh/OCR-D/core/pull/443?src=pr&el=desc) into [master](https://codecov.io/gh/OCR-D/core/commit/b063737a5cc4701fb507328b5940919848934ef1&el=desc) will **decrease** coverage by `1.66%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/OCR-D/core/pull/443?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #443      +/-   ##
==========================================
- Coverage   81.93%   80.27%   -1.67%     
==========================================
  Files          39       40       +1     
  Lines        2314     2362      +48     
  Branches      426      430       +4     
==========================================
  Hits         1896     1896              
- Misses        346      394      +48     
  Partials       72       72              
```


| [Impacted Files](https://codecov.io/gh/OCR-D/core/pull/443?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [ocrd\_models/ocrd\_page\_user\_methods.py](https://codecov.io/gh/OCR-D/core/pull/443/diff?src=pr&el=tree#diff-b2NyZF9tb2RlbHMvb2NyZF9wYWdlX3VzZXJfbWV0aG9kcy5weQ==) | `0.00% <0.00%> (ø)` | |

------

[Continue to review full report at Codecov](https://codecov.io/gh/OCR-D/core/pull/443?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/OCR-D/core/pull/443?src=pr&el=footer). Last update [b063737...c8ae6af](https://codecov.io/gh/OCR-D/core/pull/443?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

kba:
Cannot recreate the PAGE XML with generateDS 2.30.11.  With 2.35.18 it works but now the namespaces are missing again :/ @bertsky Can you please check whether this works as expected besides the missing PAGE nsprefix?

bertsky:
> Cannot recreate the PAGE XML with generateDS 2.30.11. With 2.35.18 it works but now the namespaces are missing again :/ @bertsky Can you please check whether this works as expected besides the missing PAGE nsprefix?

This immediately falls into the same trap as https://github.com/OCR-D/core/issues/451.

kba:
> This immediately falls into the same trap as #451.

I think the issue is that generateDS tries to use [`enum.Enum`](https://docs.python.org/3/library/enum.html#enum.Enum) for simpleType enumerations which is ia standard lib since 3.4. Without a type mixin, enum constants only inherit from `Enum`, hence the issues when serializing these enum constants.

IMHO this needs to be fixed up-stream. In the meantime, I found that [a sed one-liner after schema generation](https://github.com/OCR-D/core/pull/443/commits/52b1a37d302371f1e621cfef20152a1c213ddf0b#diff-b67911656ef5d18c4ae36cb6741b7965R91) can fix the enum types.

Test case to verify:

```python
    def test_simple_types(self):
        regions = self.pcgts.get_Page().get_TextRegion()
        reg = regions[0]
        self.assertTrue(isinstance(reg.get_type(), str))
        self.assertEqual(reg.get_type(), TextTypeSimpleType.CREDIT)
        self.assertTrue(isinstance(TextTypeSimpleType.CREDIT, str))
        self.assertEqual(reg.get_type(), 'credit')
        self.assertTrue(isinstance(TextTypeSimpleType.CREDIT, str))
        reg.set_type(TextTypeSimpleType.PAGENUMBER)
        self.assertEqual(reg.get_type(), 'page-number')
        self.assertTrue(isinstance(reg.get_type(), str))
```

bertsky:
> I think the issue is that generateDS tries to use [`enum.Enum`](https://docs.python.org/3/library/enum.html#enum.Enum) for simpleType enumerations which is ia standard lib since 3.4. Without a type mixin, enum constants only inherit from `Enum`, hence the issues when serializing these enum constants.
> 
> IMHO this needs to be fixed up-stream. In the meantime, I found that [a sed one-liner after schema generation](https://github.com/OCR-D/core/pull/443/commits/52b1a37d302371f1e621cfef20152a1c213ddf0b#diff-b67911656ef5d18c4ae36cb6741b7965R91) can fix the enum types.

Great! So the generateDS problem is fairly limited and we can hope to live with it.

Just of curiosity, because the diff of the generated source does not show it: What happens to non-string `simpleType` with your sed expression? (AFAIK we only have `ConfSimpleType` in PAGE as an example, but an important one no less.)

kba:
> Just of curiosity, because the diff of the generated source does not show it: What happens to non-string simpleType with your sed expression? (AFAIK we only have ConfSimpleType in PAGE as an example, but an important one no less.)

`ConfSimpleType` is not implemented as an Enum. Here's all the classes affected by this change:

```sh
grep -r 'str, Enum' ocrd_models
ocrd_models/ocrd_models/ocrd_page_generateds.py:class AlignSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class ChartTypeSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class ColourDepthSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class ColourSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class GraphicsTypeSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class GroupTypeSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class PageTypeSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class ProductionSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class ReadingDirectionSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class TextDataTypeSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class TextLineOrderSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class TextTypeSimpleType(str, Enum):
ocrd_models/ocrd_models/ocrd_page_generateds.py:class underlineStyleType(str, Enum):
```

kba:
Upstream issue: https://bitbucket.org/dkuhlman/generateds/issues/30/enum-for-simpletypes-should-mixin-str

PR #444, opened 2020-02-20
Parsing a json string should also work with long json strings.

j23d:
Run into this in a real world example on my Linux Ubuntu:

```bash
[...]
  File "/home/j23d/projects/ocrd/ocrd_all/.venv/lib/python3.6/site-packages/click/core.py", line 96, in invoke_param_callback
    return callback(ctx, param, value)
  File "/home/j23d/projects/ocrd/ocrd_all/core/ocrd/ocrd/decorators.py", line 30, in <lambda>
    callback=lambda ctx, param, value: parse_json_string_or_file(value))
  File "/home/j23d/projects/ocrd/ocrd_all/core/ocrd_utils/ocrd_utils/__init__.py", line 799, in parse_json_string_or_file
    with open(value, 'r') as f:
OSError: [Errno 36] File name too long: '{"dpi": -1, "textequiv_level": "word", "overwrite_words": false, "raw_lines": false, "char_whitelist": "", "char_blacklist": "", "char_unblacklist": ""}'
```

### DIFF
--- a/ocrd_utils/ocrd_utils/__init__.py
+++ b/ocrd_utils/ocrd_utils/__init__.py
@@ -194,7 +194,7 @@ def coordinates_for_segment(polygon, parent_image, parent_coords):
     - ``parent_coords``, its corresponding affine transformation,
 
     ...calculate the absolute coordinates within the page.
-    
+
     That is, apply the given transform inversely to ``polygon``
     The transform encodes (recursively):
 
@@ -409,12 +409,12 @@ def transpose_image(image, method):
       columns become rows (but counted from the bottom),
       i.e. all pixels get mirrored at the opposite diagonal;
       width becomes height and vice versa
- 
+
     Return a new PIL.Image.
     """
     LOG.debug('transposing image with %s', membername(Image, method))
     return image.transpose(method)
-    
+
 def get_local_filename(url, start=None):
     """
     Return local filename, optionally relative to ``start``
@@ -455,7 +455,7 @@ def image_from_polygon(image, polygon, fill='background', transparency=False):
     Images which already have an alpha channel will have it shrunk
     from the polygon mask (i.e. everything outside the polygon will
     be transparent, in addition to existing transparent pixels).
-    
+
     Return a new PIL.Image.
     """
     mask = polygon_mask(image, polygon)
@@ -591,12 +591,12 @@ def polygon_mask(image, coordinates):
 
 def adjust_canvas_to_rotation(size, angle):
     """Calculate the enlarged image size after rotation.
-    
+
     Given a numpy array ``size`` of an original canvas (width and height),
     and a rotation angle in degrees counter-clockwise ``angle``,
     calculate the new size which is necessary to encompass the full
     image after rotation.
-    
+
     Return a numpy array of the enlarged width and height.
     """
     angle = np.deg2rad(angle)
@@ -608,11 +608,11 @@ def adjust_canvas_to_rotation(size, angle):
 
 def adjust_canvas_to_transposition(size, method):
     """Calculate the flipped image size after transposition.
-    
+
     Given a numpy array ``size`` of an original canvas (width and height),
     and a transposition mode ``method`` (see ``transpose_image``),
     calculate the new size after transposition.
-    
+
     Return a numpy array of the enlarged width and height.
     """
     if method in [Image.ROTATE_90,
@@ -634,7 +634,7 @@ def rotate_coordinates(transform, angle, orig=np.array([0, 0])):
     by pure rotation, and subsequent translation back. However, since
     rotation necessarily increases the bounding box, and thus image size,
     do not translate back the same amount, but to the enlarged offset.)
-    
+
     Return a numpy array of the resulting affine transformation matrix.
     """
     LOG.debug('rotating coordinates by %.2f° around %s', angle, str(orig))
@@ -662,7 +662,7 @@ def shift_coordinates(transform, offset):
     ``offset`` of the translation vector, calculate the affine
     coordinate transform corresponding to the composition of both
     transformations.
-    
+
     Return a numpy array of the resulting affine transformation matrix.
     """
     LOG.debug('shifting coordinates by %s', str(offset))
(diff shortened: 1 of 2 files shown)


bertsky:
I don't think there is much we can do about this restriction – it comes from the kernel or FS.

You can always use actual parameter _files_ instead of strings, though. 

And if you don't want to have a permanent file in your filesystem, you can create a temporary one via process substitution:

    ocrd-tesserocr-recognize -I OCR-D-SEG-LINE -O OCR-D-OCR-TESS -p <(echo '{"dpi": -1, "textequiv_level": "word", "overwrite_words": false, "raw_lines": false, "char_whitelist": "", "char_blacklist": "", "char_unblacklist": ""}')

Regardless, I think you should change the title: _Parsing a json string ~~or file~~ should also work with long json strings._

j23d:
> I don't think there is much we can do about this restriction – it comes from the kernel or FS.
> You can always use actual parameter _files_ instead of strings, though.
> [...]

yes, I know. Just thought that it shouldn't fail based on the size of the string. But if its outside of the scope or to OS specific, I'll use it as you described.

> Regardless, I think you should change the title: _Parsing a json string ~or file~ should also work with long json strings._

done.

kba:
> fail based on the size of the strin

We can add an upper limit on file names in `parse_json_string_or_file` of 4kB (which is the limit in Linux AFAIK).

bertsky:
> Did not realize this was a PR :D @j23d's solution is more pythonic than my idea. I'm fine with this.

Neither did I! Sorry @j23d, you were right of course.

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -194,7 +194,7 @@ def coordinates_for_segment(polygon, parent_image, parent_coords):
     - ``parent_coords``, its corresponding affine transformation,
 
     ...calculate the absolute coordinates within the page.
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -409,12 +409,12 @@ def transpose_image(image, method):
       columns become rows (but counted from the bottom),
       i.e. all pixels get mirrored at the opposite diagonal;
       width becomes height and vice versa
- 
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -409,12 +409,12 @@ def transpose_image(image, method):
...
     Return a new PIL.Image.
     """
     LOG.debug('transposing image with %s', membername(Image, method))
     return image.transpose(method)
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -455,7 +455,7 @@ def image_from_polygon(image, polygon, fill='background', transparency=False):
     Images which already have an alpha channel will have it shrunk
     from the polygon mask (i.e. everything outside the polygon will
     be transparent, in addition to existing transparent pixels).
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -591,12 +591,12 @@ def polygon_mask(image, coordinates):
 
 def adjust_canvas_to_rotation(size, angle):
     """Calculate the enlarged image size after rotation.
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -591,12 +591,12 @@ def polygon_mask(image, coordinates):
...
     Given a numpy array ``size`` of an original canvas (width and height),
     and a rotation angle in degrees counter-clockwise ``angle``,
     calculate the new size which is necessary to encompass the full
     image after rotation.
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -608,11 +608,11 @@ def adjust_canvas_to_rotation(size, angle):
 
 def adjust_canvas_to_transposition(size, method):
     """Calculate the flipped image size after transposition.
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -608,11 +608,11 @@ def adjust_canvas_to_rotation(size, angle):
...
     Given a numpy array ``size`` of an original canvas (width and height),
     and a transposition mode ``method`` (see ``transpose_image``),
     calculate the new size after transposition.
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -634,7 +634,7 @@ def rotate_coordinates(transform, angle, orig=np.array([0, 0])):
     by pure rotation, and subsequent translation back. However, since
     rotation necessarily increases the bounding box, and thus image size,
     do not translate back the same amount, but to the enlarged offset.)
-    
+
```
bertsky:
```suggestion
    
```

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -662,7 +662,7 @@ def shift_coordinates(transform, offset):
     ``offset`` of the translation vector, calculate the affine
     coordinate transform corresponding to the composition of both
     transformations.
-    
+
```
bertsky:
```suggestion
    
```

kba:
Thanks @j23d

j23d:
> Did not realize this was a PR :D

ah, ok, was wondering a bit ;)

> As for the whitespace changes, [...]

good to know, I'll check next time.

PR #456, opened 2020-03-12
bashlib parse_argv: avoid errexit (catch+raise)

### LINKED ISSUE #455: bashlib: parse_argv not catching error when errexit in effect
bash scripts often use `errexit` (`set -e`) to prevent errors from going unnoticed and causing unintended follow-ups – without the need to catch each and every command's return status. This is also the case for `ocrd-olena-binarize` and `ocrd-im6convert`.

But core bashlib's `ocrd__parse_argv` (which is part of `ocrd__wrap`) cannot cope with this: when it tries to validate the JSON parameter file [here](https://github.com/OCR-D/core/blob/126d0d8de58336abe88b514e298dc204cb390bdc/ocrd/bashlib/src/parse_argv.bash#L60) via command substitution, errexit will cause the program to exit immediately, and **silently** (without showing why).

bertsky:
Fixes #455

### DIFF
--- a/ocrd/bashlib/src/parse_argv.bash
+++ b/ocrd/bashlib/src/parse_argv.bash
@@ -57,13 +57,11 @@ ocrd__parse_argv () {
 
 
     local params_parsed retval
-    params_parsed="$(ocrd ocrd-tool "$OCRD_TOOL_JSON" tool $OCRD_TOOL_NAME parse-params -p "${ocrd__argv[parameter]:-{\}}")"
-    retval=$?
-    if [[ $retval != 0 ]];then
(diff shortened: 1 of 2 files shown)


bertsky:
Example:
```
ocrd-olena-binarize -I OCR-D-IMG -O BIN -p '{"impl":"sauvola-ms-split", "k": "0.05"}'
ERROR: Failed to parse parameters (retval 1):
<report valid="false">
  <error>[k] '0.05' is not of type 'number'</error>
</report>
```

(Was just silent failure before.)

kba:
Great, thanks.

PR #457, opened 2020-03-17
really revert 3a0a3a8 (generateDS PAGE API)

### LINKED ISSUE #451: simple types in PAGE model are broken
This is a regression from #437:

```python
>>> regions[0].get_type()
'page-number'
>>> isinstance(regions[0].get_type(), str)
True
>>> isinstance(TextTypeSimpleType.PAGENUMBER, str)
False
>>> TextTypeSimpleType.PAGENUMBER == 'page-number'
False
>>> regions[0].set_type(TextTypeSimpleType.PAGENUMBER)
>>> regions[0].get_type()
<TextTypeSimpleType.PAGENUMBER: 'page-number'>
>>> str(TextTypeSimpleType.PAGENUMBER)
'TextTypeSimpleType.PAGENUMBER'
```

The latter is also what is used for XML serialization. This in turn causes invalid PAGE output when using `ocrd-tesserocr-segment-region`, which afterwards even without validation causes tons of error messages of the following form:
```
Warning: Value "TextTypeSimpleType.HEADING" near line 56 does not match xsd enumeration restriction on TextTypeSimpleType
```

This is super-urgent. I recommend doing a revert release first and then start proper investigation.

bertsky:
This (temporarily) fixes #451 by (actually) reverting (this time) #437.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [2.4.4] - 2020-03-17
+
+Fixed:
+
+  * #437 broke simple types in PAGE model, (really) revert 3a0a3a8, #451
+
 ## [2.4.3] - 2020-03-12
 
 Fixed:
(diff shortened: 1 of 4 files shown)


bertsky:
Had to revert the tests accordingly. Also making a new version here (although I don't know whether @kba will be able to upload the new release on PyPI any time soon).

Issue #366, opened 2019-12-04
bashlib: Pass log level to ocrd calls and add command to log spec-conformant


kba:
E.g.

```
$ocrd-olena-binarize -l WARN -I MAX -O WOLF,WOLF-IMG -m mets.xml -p '{"impl":"wolf"}'
12:32:31.776 INFO ocrd.workspace - Saving mets '/data/2019-12-04/mets.xml'
```

That should not log. Whereas `-L DEBUG` should log more verbose.

bertsky:
Is this really something to be fixed in core? AFAICS currently simply `ocrd-olena-binarize` is ignorant of the runtime loglevel (as can already be accessed by `${ocrd__argv[log_level]}`). 

But maybe you want to offer a bashlib CLI command to do the logging, which encapsulates everything (message formatting included)?

kba:
You're right, it's more of a documentation issue, i.e. calls to `ocrd workspace` etc. in ocrd_olena should always include `-l ${ocrd__argv[log_level}`.

> But maybe you want to offer a bashlib CLI command to do the logging, which encapsulates everything (message formatting included)?

Indeed we should. I'm surprised we don't already.

bertsky:
> You're right, it's more of a documentation issue, i.e. calls to `ocrd workspace` etc. in ocrd_olena should always include `-l ${ocrd__argv[log_level}`.

Or, maybe I was not right at all: maybe we should support setting a loglevel override via _environment variable_ in ocrd_utils, and then export to that variable in bashlib's `ocrd__wrap`?

bertsky:
> > But maybe you want to offer a bashlib CLI command to do the logging, which encapsulates everything (message formatting included)?
> 
> Indeed we should. I'm surprised we don't already.

In case you want to do this with bash code alone (not delegating to the Python implementation): I have a spec-conforming implementation in [ocrd-import](https://github.com/bertsky/workflow-configuration/blob/c907f022051a8d1df53011dab9aab7e907376671/ocrd-import#L36) already.

kba:
Nice. Maybe create a little `ocrd_utils.sh` lib with logging and maybe the coordinate arithmetics function in ocrd_olena? Could be reused in different scenarios then.

kba:
> support setting a loglevel override via environment variable in ocrd_utils

You're right being not right initially. Or something. Anyway, it's a good idea, let's do that.

bertsky:
[...]

PR #460, opened 2020-04-09
[app][feat] use config file for logging

M3ssman:
Enable python log files (see: https://docs.python.org/3.6/howto/logging.html#configuring-logging)

Advance: Configuration logic and behavior is separated from programming flow

Example Configuration: 
[ocrd_logging.zip](https://github.com/OCR-D/core/files/4455950/ocrd_logging.zip)

The example Configuration affects out-of-the the following ocr-modules:

* ocrd_utils
* ocrd.workspace
* processor.RepairSegmentation
* processor.OcropySegment
* processor.OcropyDewarp
* processor.TesserocrRecognize

The following ocrd-modules are not affected (due they are not pythonic or of other reasons)
* ocrd-im6convert
* ocrd-olena-binarize
* OCR-D-SEG-PAGE-anyocr

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ COPY ocrd ./ocrd
 COPY ocrd_modelfactory ./ocrd_modelfactory/
 COPY ocrd_models ./ocrd_models
 COPY ocrd_utils ./ocrd_utils
+RUN mv ./ocrd_utils/ocrd_logging.conf /etc
 COPY ocrd_validators/ ./ocrd_validators
 COPY Makefile .
 COPY README.md .
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,8 @@ assets-clean:
 .PHONY: test
 # Run all unit tests
 test: spec assets
-	$(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
+	HOME=$(CURDIR)/ocrd_utils $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR) -k TestLogging
+	HOME=$(CURDIR) $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
 
 test-profile:
 	$(PYTHON) -m cProfile -o profile $$(which pytest)
--- a/ocrd_utils/README.md
+++ b/ocrd_utils/README.md
@@ -3,3 +3,29 @@
 > OCR-D framework - shared code, helpers, constants
 
 See https://github.com/OCR-D/core
+
+
+## OCR-D Module Logging
+
+File-based control over logging facilities is done with standard [Python 3 logging module configuration files](https://docs.python.org/3.6/howto/logging.html#configuring-logging). This way, the level, format and destinations of log messages can be customized for all OCR-D modules individually and persistently, in the usual syntax.
+
+A template configuration file (with commented examples) is included in [ocrd_logging.conf](./ocrd_logging.conf). This is meant as an example, and should be **customized**. 
+
+To get into effect, you must put a copy (under the same name) into:
+1. your current working directory, 
+2. your user directory, or
+3. `/etc`. 
+These directories are searched in said order, and the first find wins. When no config file is found, the default logging configuration applies (which uses only stdout and the `INFO` loglevel for most loggers, cf. [here](./ocrd_logging.py)).
+
+Thus, a configuration file will override *all* settings from the default configuration, and from configuration files in lower-priority directories.
+
+For more information about logging, handlers and formats, see [Python documentation](https://docs.python.org/3/howto/logging.htm).
+
+#### Docker containers
+
+In the Dockerfiles used to build `ocrd/core` (and subsequently `ocrd/all`), the above mentioned template is directly copied to `/etc/ocrd_logging.conf` within the container image. This cofiguration is thereby also the default configuration for OCR-D containers. 
+
+Thus, if you want to customize logging rules in one of these Docker containers, you can create a custom configuration file and either:
+- place it into your local workspace directory when running the OCR-D container.
+- mount it under `/etc` when starting up the container, e.g. `docker run --mount type=bind,source=host/path/to/your-template.conf,destination=/etc/ocrd_logging.conf ocrd/all`
+- include a Dockerfile step (layer or stage) which copies this into `/etc/ocrd_logging.conf` at build time in your own Docker image.
--- a/ocrd_utils/ocrd_logging.conf
+++ b/ocrd_utils/ocrd_logging.conf
@@ -1,0 +1,84 @@
+# This is a template configuration file which allows customizing
+# format and destination of log messages with OCR-D.
+# It is meant as an example, and should be customized.
+# To get into effect, you must put a copy (under the same name)
+# into your CWD, HOME or /etc. These directories are searched
+# in said order, and the first find wins. When no config file
+# is found, the default logging configuration applies (cf. ocrd_logging.py).
+# 
+# mandatory loggers section
+# configure loggers with correspnding keys "root", ""
+# each logger requires a corresponding configuration section below
+#
+[loggers]
+#keys=root,ocrd_workspace
+keys=root
+#
+# mandatory handlers section
+# handle output for each logging "channel"
+# i.e. console, file, smtp, syslog, http, ...
+# each handler requires a corresponding configuration section below
+#
+[handlers]
+keys=consoleHandler,fileHandler
+
+#
+# optional custom formatters section
+# format message fields, to be used differently by logging handlers
+# each formatter requires a corresponding formatter section below
+#
+[formatters]
+keys=defaultFormatter,detailedFormatter
+
+#
+# default logger "root" using consoleHandler
+#
+[logger_root]
+level=INFO
+handlers=consoleHandler
+
+#
+# additional logger configurations can be added
+# as separate configuration sections like below
+#
+# example logger "ocrd_workspace" uses fileHandler and overrides
+# default log level "INFO" with custom level "DEBUG" 
+# "qualname" must match the logger label used in the corresponding 
+# ocrd modul
+# see in the modul-of-interrest (moi)
+#
+#[logger_ocrd_workspace]
+#level=DEBUG
+#handlers=fileHandler
+#qualname=ocrd.workspace
+
+#
+# handle stdout output
+#
+[handler_consoleHandler]
+class=StreamHandler
+formatter=defaultFormatter
+args=(sys.stdout,)
+
+#
+# example logfile handler
+# handle output with logfile
+#
+[handler_fileHandler]
+class=FileHandler
+formatter=detailedFormatter
+args=('ocrd.log','a+')
+
+#
+# default log format conforming to OCR-D (https://ocr-d.de/en/spec/cli#logging)
+#
+[formatter_defaultFormatter]
+format=%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s
+datefm=%H:%M:%S
+
+#
+# store more logging context information
+#
+[formatter_detailedFormatter]
+format=%(asctime)s.%(msecs)03d %(levelname)-8s (%(name)s)[%(filename)s:%(lineno)d] - %(message)s
+datefm=%H:%M:%S
--- a/ocrd_utils/ocrd_utils/logging.py
+++ b/ocrd_utils/ocrd_utils/logging.py
@@ -16,6 +16,7 @@ These files will be executed in the context of ocrd/ocrd_logging.py, with `loggi
 from __future__ import absolute_import
 
 import logging
+import logging.config
 import os
 
 __all__ = [
@@ -29,9 +30,9 @@ __all__ = [
 _overrideLogLevel = None
 
 _ocrdLevel2pythonLevel = {
-    'TRACE':    'DEBUG',
-    'OFF':      'CRITICAL',
-    'FATAL':    'ERROR',
+    'TRACE': 'DEBUG',
+    'OFF': 'CRITICAL',
+    'FATAL': 'ERROR',
 }
 
 def getLevelName(lvl):
@@ -72,14 +73,35 @@ def getLogger(*args, **kwargs):
     logger = logging.getLogger(*args, **kwargs)
     if _overrideLogLevel is not None:
         logger.setLevel(logging.NOTSET)
+    if logger.handlers:
+        # avoid duplicating messages
+        logger.propagate = False
     return logger
 
 # Default logging config
 
 def initLogging():
     """
-    Sets logging defaults
+    Reset root logger, read logging configuration if exists, otherwise use basicConfig
     """
+
+    for handler in logging.root.handlers[:]:
+        logging.root.removeHandler(handler)
+
+    global _overrideLogLevel # pylint: disable=global-statement
+    _overrideLogLevel = None
+    
+    CONFIG_PATHS = [
+        os.path.curdir,
+        os.path.join(os.path.expanduser('~')),
+        '/etc',
+    ]
+    for p in CONFIG_PATHS:
+        config_file = os.path.join(p, 'ocrd_logging.conf')
+        if os.path.exists(config_file):
+            logging.config.fileConfig(config_file)
+            return
+
     logging.basicConfig(
         level=logging.INFO,
         format='%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s',
@@ -91,22 +113,6 @@ def initLogging():
     logging.getLogger('PIL').setLevel(logging.INFO)
     # To cut back on the `Self-intersection at or near point` INFO messages
     logging.getLogger('shapely.geos').setLevel(logging.ERROR)
-
-    # Allow overriding
-
-    CONFIG_PATHS = [
-        os.path.curdir,
-        os.path.join(os.path.expanduser('~')),
-        '/etc',
-    ]
-
-
-    for p in CONFIG_PATHS:
-        config_file = os.path.join(p, 'ocrd_logging.py')
-        if os.path.exists(config_file):
-            logging.debug("Loading logging configuration from '%s'", config_file)
-            with open(config_file) as f:
-                code = compile(f.read(), config_file, 'exec')
-                exec(code, globals(), locals()) # pylint: disable=exec-used
+    
 
 initLogging()
--- a/ocrd_utils/ocrd_utils/logging.py
+++ b/ocrd_utils/ocrd_utils/logging.py
@@ -19,6 +19,8 @@ import logging
 import logging.config
 import os
 
+from .constants import LOG_FORMAT, LOG_DATEFMT
+
 __all__ = [
     'logging',
     'getLogger',
@@ -90,7 +92,7 @@ def initLogging():
 
     global _overrideLogLevel # pylint: disable=global-statement
     _overrideLogLevel = None
-    
+
     CONFIG_PATHS = [
         os.path.curdir,
         os.path.join(os.path.expanduser('~')),
@@ -104,8 +106,8 @@ def initLogging():
 
     logging.basicConfig(
         level=logging.INFO,
-        format='%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s',
-        datefmt='%H:%M:%S')
+        format=LOG_FORMAT,
+        datefmt=LOG_DATEFMT)
     logging.getLogger('').setLevel(logging.INFO)
     #  logging.getLogger('ocrd.resolver').setLevel(logging.INFO)
     #  logging.getLogger('ocrd.resolver.download_to_directory').setLevel(logging.INFO)
@@ -113,6 +115,6 @@ def initLogging():
     logging.getLogger('PIL').setLevel(logging.INFO)
     # To cut back on the `Self-intersection at or near point` INFO messages
     logging.getLogger('shapely.geos').setLevel(logging.ERROR)
-    
+
 
 initLogging()
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -10,7 +10,7 @@ from ocrd.decorators import (
     ocrd_loglevel,
     ocrd_cli_wrap_processor,
 )    # pylint: disable=protected-access
-from ocrd_utils.logging import setOverrideLogLevel, initLogging
+from ocrd_utils.logging import initLogging
 from ocrd_utils import pushd_popd, VERSION as OCRD_VERSION
 
 @click.command()
@@ -49,7 +49,6 @@ class DummyProcessor(Processor):
 def cli_dummy_processor(*args, **kwargs):
     return ocrd_cli_wrap_processor(DummyProcessor, *args, **kwargs)
 
-
 class TestDecorators(TestCase):
 
     def setUp(self):
@@ -72,7 +71,7 @@ class TestDecorators(TestCase):
         result = self.runner.invoke(cli_with_ocrd_loglevel, ['--log-level', 'DEBUG'])
         self.assertEqual(result.exit_code, 0)
         self.assertEqual(logging.getLogger('PIL').getEffectiveLevel(), logging.DEBUG)
-        setOverrideLogLevel('INFO')
+        initLogging()
 
     def test_processor_dump_json(self):
         result = self.runner.invoke(cli_dummy_processor, ['--dump-json'])
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -11,7 +11,9 @@ from ocrd_utils import (
     getLevelName,
     setOverrideLogLevel,
     initLogging,
-    getLogger
+    getLogger,
+    LOG_FORMAT,
+    LOG_DATEFMT
 )
 
 class TestLogging(TestCase):
@@ -38,8 +40,45 @@ class TestLogging(TestCase):
         self.assertEqual(getLevelName('FATAL'), logging.ERROR)
         self.assertEqual(getLevelName('OFF'), logging.CRITICAL)
 
-    def tearDown(self):
-        initLogging()
+    def test_logging_non_duplicate(self):
+        """
+        Verify that child loggers don't propagate a log message they handle
+        """
+
+        root_logger = logging.getLogger('')
+        self.assertTrue(root_logger.handlers, 'root logger has at least 1 handler')
+
+        parent_capture = FIFOIO(256)
+        parent_handler = logging.StreamHandler(parent_capture)
+        parent_handler.setFormatter(logging.Formatter(fmt=LOG_FORMAT, datefmt=LOG_DATEFMT))
+        parent_logger = getLogger('foo')
+        self.assertTrue(parent_logger.propagate, 'no handler on logger => propagate')
+        parent_logger.addHandler(parent_handler)
+        parent_logger = getLogger('foo')
+        self.assertFalse(parent_logger.propagate, 'should not propagate because StreamHandler has been attached')
+
+        child_capture = FIFOIO(256)
+        child_handler = logging.StreamHandler(child_capture)
+        child_handler.setFormatter(logging.Formatter(fmt=LOG_FORMAT, datefmt=LOG_DATEFMT))
+        child_logger = getLogger('foo.bar')
+        self.assertTrue(child_logger.propagate, 'no handler on logger => propagate')
+        child_logger.addHandler(child_handler)
+        child_logger = getLogger('foo.bar')
+        self.assertFalse(child_logger.propagate, 'should not propagate because StreamHandler has been attached')
+
+        child_logger.error('test')
+
+        parent_output = parent_capture.getvalue()
+        parent_capture.close()
+        child_output = child_capture.getvalue()
+        child_capture.close()
+        # print('parent', parent_output)
+        print('child', child_output)
+
+        # self.assertEqual(parent_output, '', 'parent logger should not receive msg')
+        # 0000-00-00 00:00:00,000.000 ERROR foo.bar - test\n
+        self.assertTrue(match(r'\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d(\.\d+)? ERROR foo.bar - test\n', child_output),
+                'regex match the log entry format')
 
 class TestLoggingConfiguration(TestCase):
 
@@ -74,13 +113,12 @@ class TestLoggingConfiguration(TestCase):
                 # ensure log level is set from temporary config file
                 self.assertEqual(logging.getLogger('').getEffectiveLevel(), logging.ERROR)
 
-
 class TestProfileLogging(TestCase):
 
     def testProcessorProfiling(self):
         log_capture_string = FIFOIO(256)
         ch = logging.StreamHandler(log_capture_string)
-        ch.setFormatter(logging.Formatter('%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s'))
+        ch.setFormatter(logging.Formatter(LOG_FORMAT))
         getLogger('ocrd.process.profile').setLevel('DEBUG')
         getLogger('ocrd.process.profile').addHandler(ch)
 
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -16,6 +16,9 @@ from ocrd_utils import (
     LOG_DATEFMT
 )
 
+# "0000-00-00 00:00:00.000 "
+DATEFMT_RE = r'\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d\.(\d+)? '
+
 class TestLogging(TestCase):
 
     def setUp(self):
@@ -57,28 +60,39 @@ class TestLogging(TestCase):
         parent_logger = getLogger('foo')
         self.assertFalse(parent_logger.propagate, 'should not propagate because StreamHandler has been attached')
 
+        child_logger = getLogger('foo.bar')
+        self.assertTrue(child_logger.propagate, 'no handler on logger => propagate')
+        child_logger.setLevel('DEBUG')
+
+        child_logger.error('first error')
+
         child_capture = FIFOIO(256)
         child_handler = logging.StreamHandler(child_capture)
         child_handler.setFormatter(logging.Formatter(fmt=LOG_FORMAT, datefmt=LOG_DATEFMT))
-        child_logger = getLogger('foo.bar')
-        self.assertTrue(child_logger.propagate, 'no handler on logger => propagate')
+
+        child_logger.debug('first debug')
+
         child_logger.addHandler(child_handler)
         child_logger = getLogger('foo.bar')
         self.assertFalse(child_logger.propagate, 'should not propagate because StreamHandler has been attached')
 
-        child_logger.error('test')
+        child_logger.debug('second debug')
+        child_logger.error('second error')
 
         parent_output = parent_capture.getvalue()
         parent_capture.close()
         child_output = child_capture.getvalue()
         child_capture.close()
         # print('parent', parent_output)
-        print('child', child_output)
+        # print('child', child_output)
 
-        # self.assertEqual(parent_output, '', 'parent logger should not receive msg')
-        # 0000-00-00 00:00:00,000.000 ERROR foo.bar - test\n
-        self.assertTrue(match(r'\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d(\.\d+)? ERROR foo.bar - test\n', child_output),
-                'regex match the log entry format')
+        self.assertTrue(match(DATEFMT_RE + 'ERROR foo.bar - first error\n', parent_output),
+                'parent received first error but not second error nor first debug')
+        self.assertTrue(match("\n".join([
+            DATEFMT_RE + 'DEBUG foo.bar - second debug',
+            DATEFMT_RE + 'ERROR foo.bar - second error',
+            ]), child_output),
+                'child received second error and debug but not first error and debug')
 
 class TestLoggingConfiguration(TestCase):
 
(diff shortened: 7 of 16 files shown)


bertsky:
> Example Configuration:
> [ocrd_logging.zip](https://github.com/OCR-D/core/files/4455950/ocrd_logging.zip)

```
formatter=ocrdFormatter
...

[formatter_ocrdFormatter]
format=%(asctime)s.%(msecs)03d %(levelname)-8s (%(name)s)[%(filename)s:%(lineno)d] - %(message)s
datefm=%H:%M:%S
```

Is it necessary to have this in the file, or can the default OCR-D formatter be used implicitly, too?



> The following ocrd-modules are not affected (due they are not pythonic or of other reasons)
> 
>     * ocrd-im6convert
>     * ocrd-olena-binarize

Yes, a known short-coming of bashlib, see https://github.com/OCR-D/core/issues/366

>     * OCR-D-SEG-PAGE-anyocr

That's strange. ocrd_anybaseocr uses the same Python facilities (`ocrd_utils.getLogger`) as the others, consistently. (Except that it does not prefix its logger names with `processor.`.

Are you quite sure?

M3ssman:
@anybase_ocr these are my observations. 
@default Yes, it only takes action in one of the locations mention in the file a config is found. If no `ocrd_logging.conf`is present, everything is like usual.

M3ssman:
@bertsky @kba 
Currently two test fail. I had to alter behavior of logging, since otherwise with the current implementation you get actually two loggers, one programmatically per default, the second from configuration. The second unfortunately interferes with the first one. The modifications in this PR skip the programmatically logger in favor of the configured one, as long there's a config file present at one of the three locations (`/etc`, `$HOME`, `$PWD`).

Let me if this is okay and I will tune the testcases to reflect this.

ocrd_utils/README.md
```diff
@@ -1,5 +1,21 @@
...
+Configuration of Logging-Facilities is done with standard Python3 Logging Library from modul `logging`. This way, logging configuration can be customized for all OCR-D modules using the Python logging framework.  
+A sample configuration file is included in the OCR-D container image at `/etc/ocrd_logging.conf`. This cofiguration is also meant to be the default configuration for OCR-D container images.  
+If this file doesn't exist or gets removed, logging mechanics use as fallback the `logging.basicConfig` functionalities which use stdout as message channel. 
+
+For more information about logging, handlers and formats, please see https://docs.python.org/3/howto/logging.htm.
+
+### Customization
+
+Create a custom Configuration and include this as part of your local workspace when running the OCR-D container.  
+In this case, it will override *all* settings from default configuration, which is included at `/etc/ocrd_logging.conf`.
```
bertsky, line 21: I don't think we should make this about Docker containers specifically. First we need a documentation of the general mechanism here (as in my suggestion for the preamble of the template file, but more user-oriented). After that, some specifics for Docker setup can be given.

```suggestion
File-based control over logging facilities is done with standard [Python 3 logging module configuration files](https://docs.python.org/3.6/howto/logging.html#configuring-logging). This way, the level, format and destinations of log messages can be customized for all OCR-D modules individually and persistently, in the usual syntax.

A template configuration file (with commented examples) is included in [ocrd_logging.conf](./ocrd_logging.conf). This is meant as an example, and should be **customized**. 

To get into effect, you must put a copy (under the same name) into:
1. your current working directory, 
2. your user directory, or
3. `/etc`. 
These directories are searched in said order, and the first find wins. When no config file is found, the default logging configuration applies (which uses only stdout and the `INFO` loglevel for most loggers, cf. [here](./ocrd_logging.py)).

Thus, a configuration file will override *all* settings from the default configuration, and from configuration files in lower-priority directories.

For more information about logging, handlers and formats, see [Python documentation](https://docs.python.org/3/howto/logging.htm).

#### Docker containers

In the Dockerfiles used to build `ocrd/core` (and subsequently `ocrd/all`), the above mentioned template is directly copied to `/etc/ocrd_logging.conf` within the container image. This cofiguration is thereby also the default configuration for OCR-D containers. 

Thus, if you want to customize logging rules in one of these Docker containers, you can create a custom configuration file and either:
- place it into your local workspace directory when running the OCR-D container.
- include a Dockerfile step (layer or stage) which copies this into `/etc/ocrd_logging.conf` at build time in your own Docker image.
```

ocrd_utils/ocrd_utils/logging.py
```diff
@@ -79,34 +79,26 @@ def getLogger(*args, **kwargs):
...
-            logging.debug("Loading logging configuration from '%s'", config_file)
             logging.config.fileConfig(config_file)
             return
 
+    logging.basicConfig(
+        level=logging.INFO,
+        format='%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s',
+        datefmt='%H:%M:%S')
+    logging.getLogger('PIL').setLevel(logging.INFO)
+    logging.getLogger('shapely.geos').setLevel(logging.ERROR)
```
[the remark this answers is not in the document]
M3ssman: Personally, I think it is a source of coming confusions. If the configuration style is to be appreciated, why to keep programmatic relics? They smell like common quick-and-dirty solutions that shouldn't be part of an official and professional piece of software. Do you want to keep this flavor?
Remember, they will never be taken into account if a config file is present. A future user might wonder what those lines stand for and why they seem to have zero effects. 
Nevertheless, the information itself is not lost when line-drop is committed, since the repository keeps them in it's history.

Dockerfile
```diff
@@ -10,6 +10,7 @@ COPY ocrd ./ocrd
 COPY ocrd_modelfactory ./ocrd_modelfactory/
 COPY ocrd_models ./ocrd_models
 COPY ocrd_utils ./ocrd_utils
+RUN mv ./ocrd_utils/ocrd_logging.conf /etc
```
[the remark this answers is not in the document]
M3ssman: What kind of template do you have in mind? Really basic? I'd like to see at least a console and a file handler, since this is a common requirement in prod environments.

ocrd_utils/README.md
M3ssman: Well, a really smart suggestion!

ocrd_utils/ocrd_utils/logging.py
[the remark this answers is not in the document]
bertsky: > Personally, I think it is a source of coming confusions. If the configuration style is to be appreciated, why to keep programmatic relics? They smell like common quick-and-dirty solutions that shouldn't be part of an official and professional piece of software. Do you want to keep this flavor?

You definitely need some fallback when no config file is present in the search directories. (You _could_ make that impossible by adding the source directory to the bottom of the list and adding the template config to the `package_data` directive. But that's not any less obscure IMO.)

> A future user might wonder what those lines stand for and why they seem to have zero effects.

Unlikely. A user capable of browsing the source code will also notice that it _says right here_ that this is only the fallback position.

> Nevertheless, the information itself is not lost when line-drop is committed, since the repository keeps them in it's history.

There's a huge (functional) difference between information preserved in a history and in comments: History can give you the vertical evolution of a piece of code _when you are already_ prepared to dig into it. Comments on the other hand readily avail themselves for information that is helpful to understand the intention of the present state and usually does not enter the history (for being too low-threshold / specific / contingent).

Dockerfile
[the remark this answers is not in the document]
bertsky: Foremost, one that reproduces the default you found in the code:
- `INFO` level for most loggers, not just `ocrd.workspace` – your `ERROR` for the root logger would be inherited by processor loggers
- stdout handler only

> I'd like to see at least a console and a file handler, since this is a common requirement in prod environments.

There's nothing wrong with merely showing how to add one by example. But I don't see the point in having a file logger with a fixed file path (`ocrd.log`) by default. This will inevitably cause problems (e.g. multiple processes conflicting in that file, either synchronously or successively).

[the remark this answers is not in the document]
kba: Yes, let's keep the default as minimal as it was before: [log to stdout with a simple pattern](https://ocr-d.de/en/spec/cli#logging) and log level at INFO (which is why the tests fail I think).

ocrd_utils/README.md
```diff
@@ -1,5 +1,30 @@
...
+
+For more information about logging, handlers and formats, see [Python documentation](https://docs.python.org/3/howto/logging.htm).
+
+#### Docker containers
+
+In the Dockerfiles used to build `ocrd/core` (and subsequently `ocrd/all`), the above mentioned template is directly copied to `/etc/ocrd_logging.conf` within the container image. This cofiguration is thereby also the default configuration for OCR-D containers. 
+
+Thus, if you want to customize logging rules in one of these Docker containers, you can create a custom configuration file and either:
+- place it into your local workspace directory when running the OCR-D container.
+- include a Dockerfile step (layer or stage) which copies this into `/etc/ocrd_logging.conf` at build time in your own Docker image.
```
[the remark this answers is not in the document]
bertsky: > Couldn't you mount the file into a container at runtime as well?

Theoretically yes. But unlikely with our search directories (except if the mount point is the workspace directory itself, which is already covered by the description): `$HOME` will almost always resolve to `/` because we must keep the external user for FS access, and `/etc` would probably break most tools running in the container.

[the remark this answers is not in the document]
kba: You can mount individual files into a docker container, at least it was possible in the past. But we could also extend the search path to include `/etc/ocrd/ocrd_logging.conf` so a directory with the logging configuration file (and possibly other configuration files in the future) could be mounted.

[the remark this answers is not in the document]
M3ssman: I guess the point is, when you copy it into a running container, it can cause strange issues since for the first modules the default config will be used immediately, but further ones might use the new copy. So we might end up with two loggers.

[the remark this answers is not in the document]
bertsky: > when you copy it into a running container, it can cause strange issues since for the first modules the default config will be used immediately, but further ones might use the new copy. So we might end up with two loggers.

What do you mean by _running container_ – a running workflow, or just that the Docker has been spawned from the image? The latter case would not be any different from copying the file at build time, and the former case is just bad luck/management (and still would _not_ yield two loggers at the same time)...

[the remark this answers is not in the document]
kba: I did not mean "running container", that would be strange indeed. I mean at the point when `docker run` is executed, you can define volumes/binds. All the files have to be mounted before the `ENTRYPOINT` is executed.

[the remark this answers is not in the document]
M3ssman: A running workflow (inside a container), and yes, I agree, this is an indication of organizational failure ;)

Dockerfile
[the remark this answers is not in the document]
M3ssman: The test at expects it to be `logging.INFO`but actual it is `logging.DEBUG`. I've got no idea where this is coming from ...

I found just this at `tests/test_workspace.py`, which could(?) address this:
```
from ocrd_utils import setOverrideLogLevel
setOverrideLogLevel('DEBUG')
```
Is this modification set for the whole testing range?

ocrd_utils/README.md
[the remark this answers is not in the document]
bertsky: > I mean at the point when `docker run` is executed, you can define volumes/binds. All the files have to be mounted before the `ENTRYPOINT` is executed.

I know. (That's why thumbs are up.) So IMO this realistic option should be added as a second item in the description.

Dockerfile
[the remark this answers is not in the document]
bertsky: > I found just this at `tests/test_workspace.py`, which could(?) address this:
> 
> ```
> from ocrd_utils import setOverrideLogLevel
> setOverrideLogLevel('DEBUG')
> ```
> 
> Is this modification set for the whole testing range?

No just for that test set (`test_workspace`). It simulates a `--log-level DEBUG` run-time parameter.

ocrd_utils/README.md
```diff
@@ -1,5 +1,21 @@
 # ocrd_utils
 
-> OCR-D framework - shared code, helpers, constants
+Additional functionalities for OCR-D framework
```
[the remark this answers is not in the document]
bertsky:
```suggestion
> OCR-D framework - shared code, helpers, constants
```

Besides, this quotation is identical to the setuptools description.

ocrd_utils/README.md
```diff
@@ -1,5 +1,30 @@
...
+Thus, a configuration file will override *all* settings from the default configuration, and from configuration files in lower-priority directories.
+
+For more information about logging, handlers and formats, see [Python documentation](https://docs.python.org/3/howto/logging.htm).
+
+#### Docker containers
+
+In the Dockerfiles used to build `ocrd/core` (and subsequently `ocrd/all`), the above mentioned template is directly copied to `/etc/ocrd_logging.conf` within the container image. This cofiguration is thereby also the default configuration for OCR-D containers. 
+
+Thus, if you want to customize logging rules in one of these Docker containers, you can create a custom configuration file and either:
+- place it into your local workspace directory when running the OCR-D container.
```
bertsky, line 29:
```suggestion
- place it into your local workspace directory when running the OCR-D container.
- mount it under `/etc` when starting up the container, e.g. `docker run --mount type=bind,source=host/path/to/your-template.conf,destination=/etc/ocrd_logging.conf ocrd/all`
```

ocrd_utils/ocrd_logging.conf
```diff
@@ -0,0 +1,81 @@
...
+# into your CWD, HOME or /etc. These directories are searched
+# in said order, and the first find wins. When no config file
+# is found, the default logging configuration applies (cf. ocrd_logging.py).
+# 
+# mandatory loggers section
+# configure loggers with correspnding keys "root", ""
+# each logger requires a corresponding configuration section below
+#
+[loggers]
+keys=root
```
bertsky, line 14:
```suggestion
#keys=root,ocrd_workspace
keys=root
```

ocrd_utils/ocrd_logging.conf
```diff
@@ -0,0 +1,81 @@
...
+# mandatory handlers section
+# handle output for each logging "channel"
+# i.e. console, file, smtp, syslog, http, ...
+# each handler requires a corresponding configuration section below
+#
+[handlers]
+keys=consoleHandler,fileHandler
+
+#
+# optional custum formatters section
```
bertsky, line 25:
```suggestion
# optional custom formatters section
# format message fields, to be used differently by logging handlers
# each formatter requires a corresponding formatter section below
```

ocrd_utils/ocrd_logging.conf
```diff
@@ -0,0 +1,81 @@
...
+# example logfile handler
+# handle output with logfile
+#
+[handler_fileHandler]
+class=FileHandler
+formatter=detailedFormatter
+args=('ocrd.log','a+')
+
+#
+# default log format from ocrd_logging.py
```
bertsky, line 70:
```suggestion
# default log format conforming to OCR-D (https://ocr-d.de/en/spec/cli#logging)
```

ocrd_utils/ocrd_utils/logging.py
```diff
@@ -97,6 +97,7 @@ def initLogging():
         level=logging.INFO,
         format='%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s',
         datefmt='%H:%M:%S')
+    logging.getLogger('').setLevel(logging.INFO)
```
bertsky: This must have slipped through earlier – I guess this was the culprit for CI failures all along, right?

tests/test_decorators.py
```diff
@@ -49,7 +49,7 @@ def process(self):
 def cli_dummy_processor(*args, **kwargs):
     return ocrd_cli_wrap_processor(DummyProcessor, *args, **kwargs)
 
-
+setOverrideLogLevel('INFO')
```
bertsky, line 52: I don't see the point in this.

tests/test_logging.py
```diff
@@ -11,6 +11,7 @@
     getLogger
 )
 
+setOverrideLogLevel('INFO')
```
bertsky, line 14: Same here. Why do an override already? That way you cannot test overriding inside the function.

ocrd_utils/ocrd_utils/logging.py
M3ssman: Yep, that's it. I though it'd be sufficient to set the `level` parameter, but obviously it's not the case. We live, we learn.

tests/test_logging.py
M3ssman: Oops, your right! That's from my first shot to please the failing tests! Which turned out to be completely useless! 
I'll drop both asap!

bertsky:
One more request: We should also use the existing tests to automatically verify that the template config file actually does the same as the default settings. To that end, one could just add the following extra line to `Makefile`:

```make
test: spec assets
	HOME=$(CWD) $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
	HOME=$(CWD)/ocrd_utils/ocrd_utils $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR) -k TestLogging
```

This effectively first disables any config file present in the _host system's_ `$HOME`, and then repeats the (logging-related) subset of tests with the template config file.

(We should long ago have done that.)

M3ssman:
> One more request: We should also use the existing tests to automatically verify that the template config file actually does the same as the default settings. To that end, one could just add the following extra line to `Makefile`:
> 
> ```makefile
> test: spec assets
> 	HOME=$(CWD) $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
> 	HOME=$(CWD)/ocrd_utils/ocrd_utils $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR) -k TestLogging
> ```
> 
> This effectively first disables any config file present in the _host system's_ `$HOME`, and then repeats the (logging-related) subset of tests with the template config file.
> 

I think it is better to add a dedicated test class for this scenario. 
If we only re-arrange $HOME, the first assertion regarding the `rootLogger` references a logger coming straight outta python, not from the logging-module.

bertsky:
> I think it is better to add a dedicated test class for this scenario.

It depends on how this is done, but I surmise a dedicated test will always be weaker than that. We want to ensure config file equals defaults and both are correct. So let's simply test all desired behaviour _twice_: once with config file, and once without.

> 
> If we only re-arrange $HOME, the first assertion regarding the `rootLogger` references a logger coming straight outta python, not from the logging-module.

That `rootLogger` still sees the effects of `ocrd_utils.logging.initLogging`, because this is run _at import time_. There's nothing new or special here.

> Ah, there is already a testcase for this, `TestLogging2` which should be taken into account.

Oh, speaking of, you should rewrite `test_configfile` to test something meaningful under the configfile regime. (So there you have your dedicated test case, but whatever you are going to test there, it's less general than the above recipe.)

M3ssman:
I don't get the idea behind the current Implementation of `TestLoggin2.test_configfile`. 
Looks like some mocking-fixture-stuff? It's missing any assertions, so what gets actually tested?

Further, if the default test should somehow reflect the absence of the logging conf file, the only thing I can imagine would be to assert that the current rootLogger is not born of `fileConfig`. I've got no idea how to test this. 

Any suggestions or help appreciated!

bertsky:
> I don't get the idea behind the current Implementation of `TestLoggin2.test_configfile`.
> Looks like some mocking-fixture-stuff? It's missing any assertions, so what gets actually tested?

Agreed. It does not even make sense to call `initLogging` again, because that's too late after `setUp`. (But actually, even that routine should never work, because – as I said earlier – `initLogging` is already called when _importing_ `ocrd_utils.logging`, and since it uses `logging.basicConfig`, calling it a second time should have **no effect**. [Python 3.8](https://docs.python.org/3/library/logging.html) introduces the `force` kwarg against that problem, but we are stuck without it.) Maybe @kba can elaborate?

Now, since you brought in `logging.config.fileConfig`, things have gotten a little better: With `disable_existing_loggers=True` (the default) IIUC the effects of `basicConfig` do get reversed now. So we _could_ in principle re-run `initLogging` when we have a config file waiting.

So here's what I'd do:
1. Add a comment at the import that this already runs `initLogging/basicConfig`.
2. Remove the `setUp` implementation.
3. Change `test_configfile` like this:

```python
    def test_configfile(self):
        # from import-time initLogging (Python defaults)
        self.assertEqual(logging.getLogger('').getEffectiveLevel(), logging.INFO)
        with TemporaryDirectory() as tempdir:
            with pushd_popd(tempdir):
                
                with open('ocrd_logging.conf', 'w') as f:
                    f.write('''
[loggers]
keys=root
[logger_root]
level=ERROR
''')
                # this will call logging.config.fileConfig with disable_existing_loggers=True,
                # so the defaults from the import-time initLogging should be invalided
                initLogging()
                # from config file
                self.assertEqual(logging.getLogger('').getEffectiveLevel(), logging.ERROR)
```

M3ssman:
@bertsky 
Thanks for your suggestions. It was almost plain codable, with minor variations:
* I need to check first that any log level was set, i.e. it was not ` loggin.NOTSET` due the effect, that the call of `make test`wanted to put it as `logging.DEBUG`
*  some additions to create a temporary mwe config file to please the parser

tests/test_logging.py
```diff
@@ -26,25 +26,47 @@ def test_setOverrideLogLevel(self):
...
-class TestLogging2(TestCase):
-
-    def setUp(self):
-        initLogging()
 
     def test_getLevelName(self):
         self.assertEqual(getLevelName('ERROR'), logging.ERROR)
         self.assertEqual(getLevelName('FATAL'), logging.ERROR)
         self.assertEqual(getLevelName('OFF'), logging.CRITICAL)
+        setOverrideLogLevel('INFO')
```
bertsky, line 34: I was puzzled by this line before – now you moved it from one function to the other (because you moved that function from one class to the other).

@kba Please explain the reason for this in the first place!

Frankly, I don't know what unittest or pytest is supposed to do with our import-time state change. We have no explicit module or class or testcase setUp/tearDown (except for L34 before, which as I have argued should not have worked). But if pytest does in fact undo basicConfig between classes _somehow_, then at least this made _some_ sense.

tests/test_logging.py
```diff
@@ -26,25 +26,47 @@ def test_setOverrideLogLevel(self):
...
+
+
+class TestLoggingConfiguration(TestCase):
 
-    def test_configfile(self):
+    def test_tmpConfigfile(self):
+        self.assertNotEqual(logging.getLogger('').getEffectiveLevel(), logging.NOTSET)
```
bertsky, line 40: I don't get why it is `NOTSET` now. The module-level import-time `initLogging` (or any of `TestLogging`'s changes) _should_ still be in effect, right?

tests/test_logging.py
```diff
@@ -26,25 +26,47 @@ def test_setOverrideLogLevel(self):
...
+                        keys=
+                        
+                        [logger_root]
+                        level=ERROR
+                        handlers=consoleHandler
+                        
+                        [handler_consoleHandler]
+                        class=StreamHandler
+                        formatter=
+                        args=(sys.stdout,)
```
bertsky, line 62: Is it really necessary to have all these instructions? How does my minimal proposal fail?

tests/test_logging.py
bertsky: Can you please add the first part of my above proposal to avoid host-side interference _already_, namely adding `HOME=$(CWD) ` to the pytest call in the makefile? (Just to be on the safe side that no one's flying ocrd_logging.conf files get in the way.)

M3ssman: What's `$(CWD)`? You do not mean `$(PWD)`?

tests/test_logging.py
M3ssman: The parser fails with `KeyErrors` like 
```
tests/test_logging.py:64: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ocrd_utils/ocrd_utils/logging.py:93: in initLogging
    logging.config.fileConfig(config_file)
/usr/lib/python3.6/logging/config.py:76: in fileConfig
    formatters = _create_formatters(cp)
/usr/lib/python3.6/logging/config.py:109: in _create_formatters
    flist = cp["formatters"]["keys"]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <configparser.ConfigParser object at 0x7f7157cb60f0>, key = 'formatters'

    def __getitem__(self, key):
        if key != self.default_section and not self.has_section(key):
>           raise KeyError(key)
E           KeyError: 'formatters'

/usr/lib/python3.6/configparser.py:959: KeyError
```

But it's fine adding a rather empty section. Also it complains where there's no handler set for rootLogger, which in turn means that there must be a handler configured, which has to have attribute `class`set and so forth.

tests/test_logging.py
bertsky: > What's `$(CWD)`? You do not mean `$(PWD)`?

CWD is make's current working directory (i.e. the one it chdired to and actually ran from, including `-C` indirection and all). This is the safest bet.

PWD is a shell variable, and would in general be made available under that name within make, too. But it is not much use, because it only reflects the CWD at the place `make` was started from (without chdir since then).

tests/test_logging.py
bertsky: I see. So your version _is_ the minimum.

tests/test_logging.py
M3ssman: $(CWD) has not value set when I run `make test`on my Ubuntu 18.04 LTS ... ?

M3ssman: > I don't get why it is `NOTSET` now. The module-level import-time `initLogging` (or any of `TestLogging`'s changes) _should_ still be in effect, right?

The assertion is, that it is NOT logging.NOTSET so it must be set to something. 
In several local test runs it was almost everytime `logging.DEBUG` and one time `logging.INFO`. I don't why, so I handled it rather pragmatic.

bertsky: > $(CWD) has not value set when I run `make test`on my Ubuntu 18.04 LTS ... ?

Are you sure? Did you write `HOME=$(CWD) python -m pytest` etc in the recipe (and then checked for the value of `HOME` in Python)?

M3ssman: No, I just echo' d `$(CWD)` straight from the Makefile. It was empty, whereas `$(PYTHON)` yields `python`.
I'm no expert regarding make-stuff, therefore the probably absurd question.

bertsky: So sorry! I meant `$(CURDIR)`. A memory glitch...

tests/test_logging.py
kba: > @kba Please explain the reason for this in the first place!

I can't unfortunately. `git blame` didn't help. IIRC I used `tox` at the time to do cross-version testing and that behaved differently wrt to clearing module cache between tests or for python <= 2.7 or something like that.. The split into two test cases was also for that reason and is obsolete.

Makefile
```diff
@@ -132,7 +132,7 @@ assets-clean:
 .PHONY: test
 # Run all unit tests
 test: spec assets
-	$(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
+	HOME=$(CURDIR) $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
```
bertsky, line 135:
```suggestion
	HOME=$(CURDIR) $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
	HOME=$(CURDIR)/ocrd_utils/ocrd_utils $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR) -k TestLogging
```

M3ssman: This results on my local machine in

```
HOME=/home/m3ssman/Projects/github/ulb-sachsen-anhalt-core/ocrd_utils/ocrd_utils python -m pytest --continue-on-collection-errors tests -k TestLogging
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.6.9, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /home/m3ssman/Projects/github/ulb-sachsen-anhalt-core
collected 242 items / 239 deselected / 3 selected                                                                                                                                                          

tests/test_logging.py .F.                                                                                                                                                                            [100%]

================================================================================================= FAILURES =================================================================================================
___________________________________________________________________________________ TestLogging.test_setOverrideLogLevel ___________________________________________________________________________________

self = <tests.test_logging.TestLogging testMethod=test_setOverrideLogLevel>

    def test_setOverrideLogLevel(self):
        rootLogger = logging.getLogger('')
        somelogger = getLogger('foo.bar')
        somelogger.setLevel(getLevelName('ERROR'))
>       self.assertEqual(rootLogger.getEffectiveLevel(), logging.INFO)
E       AssertionError: 10 != 20

tests/test_logging.py:20: AssertionError
========================================================================================= short test summary info ==========================================================================================
FAILED tests/test_logging.py::TestLogging::test_setOverrideLogLevel - AssertionError: 10 != 20
=============================================================================== 1 failed, 2 passed, 239 deselected in 0.91s ================================================================================
Makefile:135: recipe for target 'test' failed
make: *** [test] Error 1

```

M3ssman: Something different ... I needed to add some python libs to `requirements_test.txt`, shall I propagate this?

kba: Feel free to add to `requirements_test.txt` but keep in mind that these are installed for every run of CI. Which libs do you want to add?

M3ssman: Before I could run `make test` locally without lib-errors, I needed to add `atomicwrites`, `deprecated` and `opencv-python`. The last library is quite large, though.

M3ssman: @kba Please nevermind my last remarks. Additionally, I also need to do install `ocrd_utils` because of the pushd_popd stuff. After all, I don't think it is desirable and I really like to come to an end with this ;)

kba: These libs are already in `requirements.txt` (as `Deprecated`, `atomicwrites` and `opencv-headless-python`). Are you sure, you've run `pip install -e .`?

kba: I get the same assertion error as @m3ssman with the proposed change to Makefile BTW.

M3ssman: @bertsky 
The second test locally only works if I add an explicit `initLogging`, say in a `setUp` method. Otherwise, it alerts like posted above and fails.

bertsky: > E       AssertionError: 10 != 20

That's interesting. It means that the effective level of the root logger is not `INFO` (20) but `DEBUG` (10; `TRACE`) here.

But we never set it to that level anywhere! (Except in `test_decorators.TestDecorators.test_loglevel_override`.)

Next observation: this happens regardless of what `HOME` points to. (Also, in my first suggestion, I mistakenly assumed the template was under `ocrd_utils/ocrd_utils/ocrd_logging.conf`, whereas in fact it resides one directory level upwards. So this had no effect anyway.)

Obviously, just picking one testcase/class via `-k` invalidates our setup.

bertsky: > Obviously, just picking one testcase/class via `-k` already invalidates our setup.
> But we never set it to that level anywhere! (Except in `test_decorators.TestDecorators.test_loglevel_override`.)

We're lucky to have stumbled over this at last.

Wow. The problem is that we have a `setOverrideLogLevel('DEBUG')` in `test_workspace` – on the **module level** (i.e. without even starting any test). This is also in effect when filtering test classes/cases via `-k`, because obviously pytest needs to first discover what's there before it can filter.

This was introduced by 53ca390a – and looks like an accident.

But that reminds us that we also need a much more strict test case isolation.

bertsky: > But that reminds us that we also need a much more strict test case isolation.

So how do we undo the effects of `setOverrideLogLevel` – ever? We cannot access `ocrd_utils.logging._overrideLogLevel`, because it is not exported. But `setOverrideLogLevel` does not allow to set it back to `None`.

kba: > This was introduced by 53ca390 – and looks like an accident.

Indeed it was, should be reverted.

> So how do we undo the effects of setOverrideLogLevel

Accept `None` as argument and if so, set root logger to `NOTSET`?

bertsky: > So how do we undo the effects of `setOverrideLogLevel` – ever (or of `initLogging`'s `basicConfig`, for that matter)? We cannot access `ocrd_utils.logging._overrideLogLevel`, because it is not exported. But `setOverrideLogLevel` does not allow to set it back to `None`.

I suggest the following:
- `initLogging` should attempt to undo both any previous `basicConfig`, by [removing all root logger handlers](https://stackoverflow.com/a/12158233), and then set `_overrideLogLevel = None`. Now we can reliably re-init by simply calling it again.
- to get test case isolation, we should create a `setUp` implementation (or `setUpClass`) including `initLogging` for all test cases that need a certain logging state. Also, test cases that manipulate the logging state for their own purposes should probably have a `tearDown` implementation (or `tearDownClass`) likewise.

bertsky: > > This was introduced by [53ca390](https://github.com/OCR-D/core/commit/53ca390a301cdfae0e1b3a71a6426fac6dbbd3ce) – and looks like an accident.
> 
> Indeed it was, should be reverted.

The whole commit, including `test_download_to_directory_from_workspace_download_file`? Or just the `setOverrideLogLevel` (and probably the `system('find %s' % ws_dir)`)?

kba: Again, thanks for digging and proposing the fixes :) `initLogging` resetting all overrides makes sense to me.

Have a look at https://github.com/kba/ocrd-core/commit/bd232c0dfc7363538957e5db67e0221aef89ac3e That should revert the debug statements, changes the TestCase base class to do `initLogging()` per default on `tearDown` and change initLogging() to remove all handlers for root loggers and reset `_overrideLogLevel` to `None`.

@m3ssman Feel free to use that commit, I didn't want to push to your master branch in case you're working on this.

M3ssman:
@kba @bertsky I pushed the last recommendations. Is there another review stage necessary?

Makefile
```diff
@@ -133,6 +133,7 @@ assets-clean:
 # Run all unit tests
 test: spec assets
 	HOME=$(CURDIR) $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR)
+	HOME=$(CURDIR)/ocrd_utils/ocrd_utils $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR) -k TestLogging
```
bertsky, line 136:
```suggestion
	HOME=$(CURDIR)/ocrd_utils $(PYTHON) -m pytest --continue-on-collection-errors $(TESTDIR) -k TestLogging
```

bertsky: Sorry – my ultimate request, I promise!

bertsky: Promise broken already :roll_eyes: 

There's another problem which must be dealt with before this can be merged: When you set up multiple loggers via `logging.config.fileConfig` (not via `logging.getLogger`), then all [their propagate attributes](https://docs.python.org/3/library/logging.html#logging.Logger.propagate) default to true. But this means that any message they receive will _also_ go to their parents (where they usually will be rendered as a duplicate).

For example, if I set up a logger qualifying for `ocrd.processor` at debug level and a root logger at info level, then all processor messages will be rendered twice.

The remedy is simply to do `logger.propagate = False` in `ocrd_utils.logging.getLogger`. @M3ssman may I add this commit, or would you like to do it yourself?

M3ssman: @bertsky Added your suggesstion ;)

M3ssman:
Can please somebody explain what has happened to codedev?

M3ssman:
> Can please somebody explain what has happened to codedev?

Just for protocol: I flipped the test targets within the Makefile to re-please Codedev. 
It respect only the last call for reports.

bertsky:
Wait, disabling `propagate` was a terrible idea. The formatters don't work anymore. Investigating...

bertsky:
> Wait, disabling `propagate` was a terrible idea. The formatters don't work anymore. Investigating...

Found it. The problem was that we disabled `propagate` unconditionally. But it must only be disabled for loggers that already have their own handlers (via config). Otherwise those other handlers will neither have a handler (and formatter!) nor propagate to their parent.

I cannot push to https://github.com/ulb-sachsen-anhalt/core. See https://github.com/bertsky/core/commit/1f7658074c46e1b191ef5c55db4668c9f9a2dc18

bertsky:
@kba your idea for capturing and comparing log messages in #462 is necessary to actually see this problem (i.e. duplicate messages or missing formatting). So we should also include this in `test_logging`, perhaps even with regexps for the precise date format and number of messages sent.

kba:
@bertsky Thanks again for digging and fixing! I cannot push to ulb's master either, so @M3ssman will have to merge.

> So we should also include this in test_logging, perhaps even with regexps for the precise date format and number of messages sent.

Can we merge #462 into this one? Then I can reuse the code for testing non-duplicate/correctly-formatted log messages.

IIUC this is the last change blocking merge, right?

M3ssman:
@bertsky Pushed again. 
Well, this is the most epic PR I've ever heard of ;)

M3ssman:
@kba If it makes it easier, we should integrate.

bertsky:
> IIUC this is the last change blocking merge, right?

Yes. This is as good as it gets!

> Can we merge #462 into this one? Then I can reuse the code for testing non-duplicate/correctly-formatted log messages.

@kba Whatever suits you better. (But since this PR runs on master, which makes @M3ssman the only one allowed to add commits, and we already strained him quite a bit for that contribution, I suggest merging first and then rebasing and extending #462 instead.)

kba:
Sorry, I obviously missed commits in #462. Merging now.

PR #462, opened 2020-04-17
Execution of processors is tracked with logger "profile.orcd.process.<processor-name>", fix #461

### LINKED ISSUE #461: [Request] Track execution time for each processed module
For profiling purposes, it is important to track execution time for an ocr-d workflow step to identify potential performance bottlenecks.

Please provide a configurable functionality that can be utilized this way.

kba:
@m3ssmann @bertsky

### DIFF
--- a/ocrd_utils/ocrd_utils/logging.py
+++ b/ocrd_utils/ocrd_utils/logging.py
@@ -16,6 +16,7 @@ These files will be executed in the context of ocrd/ocrd_logging.py, with `loggi
 from __future__ import absolute_import
 
 import logging
+import logging.config
 import os
 
 __all__ = [
@@ -29,9 +30,9 @@ __all__ = [
 _overrideLogLevel = None
 
 _ocrdLevel2pythonLevel = {
-    'TRACE':    'DEBUG',
-    'OFF':      'CRITICAL',
-    'FATAL':    'ERROR',
+    'TRACE': 'DEBUG',
+    'OFF': 'CRITICAL',
+    'FATAL': 'ERROR',
 }
 
 def getLevelName(lvl):
@@ -72,14 +73,35 @@ def getLogger(*args, **kwargs):
     logger = logging.getLogger(*args, **kwargs)
     if _overrideLogLevel is not None:
         logger.setLevel(logging.NOTSET)
+    if logger.handlers:
+        # avoid duplicating messages
+        logger.propagate = False
     return logger
 
 # Default logging config
 
 def initLogging():
     """
-    Sets logging defaults
+    Reset root logger, read logging configuration if exists, otherwise use basicConfig
     """
+
+    for handler in logging.root.handlers[:]:
+        logging.root.removeHandler(handler)
+
+    global _overrideLogLevel # pylint: disable=global-statement
+    _overrideLogLevel = None
+    
+    CONFIG_PATHS = [
+        os.path.curdir,
+        os.path.join(os.path.expanduser('~')),
+        '/etc',
+    ]
+    for p in CONFIG_PATHS:
+        config_file = os.path.join(p, 'ocrd_logging.conf')
+        if os.path.exists(config_file):
+            logging.config.fileConfig(config_file)
+            return
+
     logging.basicConfig(
         level=logging.INFO,
         format='%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s',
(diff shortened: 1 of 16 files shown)


bertsky:
> Do I get this right - every ocrd module project will have it's own logger?

I also find this less favourable. Remember, we wanted to simply direct the "profiling" logger to another handler (e.g. file based) so it would never spam the normal handler. With a dedicated logger for each processor, you would have a hard time anticipating logger names in the redirect config.

I propose writing the name of the processor into the log message instead, probably even the input and output file groups. This way, statistics can still be aggregated together and then sliced into details ad lib.

Also, I don't quite understand why you go to such lengths to intercept the additional messages in the new test case. Wouldn't a temporary logging config with a temporary file logger be the best solution?

M3ssman:
@kba Could you please concentrate this into a single, dedicated logger with label `ocrd.process.profile`, where the log message itself contains the labels of the actual run processor?

kba:
I thought a logger per processor for profiling was clever because it allows "toggling" profiling for specific processors in the config. But I hear you and will change it.

> Also, I don't quite understand why you go to such lengths to intercept the additional messages in the new test case. Wouldn't a temporary logging config with a temporary file logger be the best solution?

That is also an option, I just wanted the logging output in a variable without reading files and this seemed the easiest solution, it didn't feel like going to great lengths, actually. If you feel this is detrimental to the test suite, I can change it.

bertsky:
> If you feel this is detrimental to the test suite, I can change it.

Not at all, please keep it.

bertsky:
Wow, I am surprised Github is handling the timeline so badly. Now after the push from #460 the commits appear at their commit date here, too.

Perhaps some of those commits need to be squashed anyway...

ocrd_utils/ocrd_utils/logging.py
```diff
@@ -104,15 +106,15 @@ def initLogging():
 
     logging.basicConfig(
         level=logging.INFO,
-        format='%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s',
-        datefmt='%H:%M:%S')
+        format=LOG_FORMAT,
```
kba, line 109: I am confused why this worked in the first place. The `datefmt` didn't include the `year-month-day` part and `%(asctime)` should be formatted according to `datefmt`, yet it was formated `%Y-%m-%d  %H:%m:%s` as it is supposed to. Weird.

tests/test_logging.py
```diff
@@ -38,8 +40,45 @@ def test_getLevelName(self):
...
+        child_logger.error('test')
+
+        parent_output = parent_capture.getvalue()
+        parent_capture.close()
+        child_output = child_capture.getvalue()
+        child_capture.close()
+        # print('parent', parent_output)
+        print('child', child_output)
+
+        # self.assertEqual(parent_output, '', 'parent logger should not receive msg')
```
bertsky, line 78: Why is this commented out?

Another configuration worth testing: before adding a handler to `child_logger`, send a message and validate that `parent_output` received it. (So, combining both, you could assert the same condition for both outputs – which now validates that `parent_handler` received only the first message.)

Also, how about log levels? You could set `child_handler` to `DEBUG` and send a debug message to `child_logger`, both before and after you attach `child_handler`. Then you assert that `child_output` received it only once. (This validates that `parent_handler` ignored this level.)

kba: > Why is this commented out?

Forgot to comment back in before commit, fixed.

> Another configuration worth testing: before adding a handler to child_logger, send a message and validate that parent_output received it. (So, combining both, you could assert the same condition for both outputs – which now validates that parent_handler received only the first message.)

> Also, how about log levels? You could set child_handler to DEBUG and send a debug message to child_logger, both before and after you attach child_handler. Then you assert that child_output received it only once. (This validates that parent_handler ignored this level.)

See https://github.com/OCR-D/core/pull/462/commits/43d905d4834c6ec90b1ca332da51fdd8953eff01 Sends two error and two debug messages to child_logger. parent handler receives only the first error (because at that point no handler on child_logger and first debug message below parent_logger's level) and child_handler receives only second error and second debug.

bertsky: Excellent!

ocrd_utils/ocrd_utils/logging.py
kba: The reason I got confused is that I didn't read the specs even though @VolkerHartmann reminded me that it was my idea in the first place NOT to have the (redundant) date in the log. So, another chapter in the logging PR saga ....

PR #465, opened 2020-04-23
logging: subclass Logger to disable propagation if handler is set

### LINKED ISSUE #464: logging: avoid duplicate messages
Despite our epic efforts in #460 and #462, we now do have the unfortunate situation that log messsages are sometimes received as duplicates.

Example to reproduce:
1. configure a logger with handler for `processor`
2. program a processor sending to `processor.MyProc`
3. watch messages coming in twice, because `processor.MyProc` has no handler (so `getLogger` keeps `propagate=True`) and `processor` has a handler but _also_ has the default `propagate=True` (which no `getLogger` was asked to disable) so root's handler gets it as well

Proposed solution: Invert the logic and set `propagate=False` in the _root_ logger in `initLogging`, but then set `propagate=True` in `getLogger` iff the requested logger has not been equipped with handlers qua config.

kba:
fixes #464

### DIFF
--- a/ocrd_utils/ocrd_utils/logging.py
+++ b/ocrd_utils/ocrd_utils/logging.py
@@ -37,6 +37,19 @@ _ocrdLevel2pythonLevel = {
     'FATAL': 'ERROR',
 }
 
+class PropagationShyLogger(logging.Logger):
+
+    def addHandler(self, hdlr):
+        super().addHandler(hdlr)
+        self.propagate = not self.handlers
+
+    def removeHandler(self, hdlr):
+        super().removeHandler(hdlr)
+        self.propagate = not self.handlers
+
+logging.setLoggerClass(PropagationShyLogger)
+logging.getLogger().propagate = False
+
 def getLevelName(lvl):
     """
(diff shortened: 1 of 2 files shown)


kba:
> cover the case that someone adds a handler after they did getLogger.

Exactly, I see no other solution than directly hooking the logger to change propagation when handlers are added/removed.

> Does this also mean that callers will have the correct `propagate` semantics even if they don't use our `ocrd_utils.logging.getLogger`, but the standard `logging.getLogger` (as long as someone does an `ocrd_utils.logging.initLogging` earlier)?

This sets the Logger class on the `logging` package, so yes, all calls to `logging.getLogger` after `logging.setLoggerClass(PropagationShyLogger)` will receive instances of PropagationShyLogger. Unless the calls happen before or they meddling with custom loggers themselves.

I think it's a neat hack but I hope this doesn't introduce any new regressions, the devil really is in the detail with logging...

bertsky:
> I think it's a neat hack but I hope this doesn't introduce any new regressions, the devil really is in the detail with logging...

Absolutely. Plus I think this recipe/pattern is really worth broad adoption for users of Python logging.

Issue #434, opened 2020-02-11
Allow overwriting output groups


mikegerber:
@bertsky wrote in #245:
>  (The --force option is not exposed to those CLIs.)

I often have a usage pattern of 1. force-removing an output group to 2. run a processor again, e.g.:

~~~
ocrd workspace remove-group -rf OCR-D-IMG-BINPAGE
ocrd workspace remove-group -rf OCR-D-IMG-BIN
ocrd-olena-binarize -I OCR-D-IMG -O OCR-D-IMG-BINPAGE,OCR-D-IMG-BIN \
    -p "$ocrd_olena_binarize_parameters"
~~~
or
~~~
ocrd workspace remove-group -rf OCR-D-SEG-LINE
ocrd-sbb-textline-detector -I OCR-D-IMG-BINPAGE -O OCR-D-SEG-LINE \
    -p "$ocrd_sbb_textline_detector_parameters"
~~~

It would be great if the processor CLI wrapper would provide `--force` to reduce this to:

~~~
ocrd-olena-binarize --force -I OCR-D-IMG -O OCR-D-IMG-BINPAGE,OCR-D-IMG-BIN \
    -p "$ocrd_olena_binarize_parameters"
~~~
and
~~~
ocrd-sbb-textline-detector --force -I OCR-D-IMG-BINPAGE -O OCR-D-SEG-LINE \
    -p "$ocrd_sbb_textline_detector_parameters"
~~~

bertsky:
Yes, this would be the simplest solution I think. @kba, this would only change the default `force` for `ocrd_mets.add_file`, but could thus be easily wrapped by workflow engines.

kba:
> @kba, this would only change the default force for ocrd_mets.add_file, but could thus be easily wrapped by workflow engines.

That would be easy but it's not the same as removing the whole file group before running the processors. I'd prefer removing the whole group because that makes it clearer which files were created with `--force`. But maybe I'm missing a use case here?

And probably not call it `--force` but `--overwrite` because the latter is more specific.

bertsky:
> That would be easy but it's not the same as removing the whole file group before running the processors. I'd prefer removing the whole group because that makes it clearer which files were created with `--force`.

You're absolutely right. This could introduce inconsistencies, the least of which would be having two pages with the same `pageId` in the same fileGrp.

> But maybe I'm missing a use case here?

Maybe. But the above use-case it the most pressing, and it does warrant removing the whole output filegrp.

[...]

PR #471, opened 2020-04-29
Add --overwrite parameter to processor base, #434, OCR-D/spec#151

kba:


ocrd/ocrd/decorators.py
```diff
@@ -46,9 +56,25 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
...
+            LOG.debug("Removing files because of --overwrite")
+            for grp in kwargs['output_file_grp'].split(','):
+                if 'page_id' in kwargs:
+                    for page_id in kwargs['page_id'].split(','):
+                        LOG.debug("Removing files in output file group %s with page ID %s", grp, page_id)
+                        for file in workspace.mets.find_files(pageId=page_id, fileGrp=grp):
+                            workspace.remove_file(file, force=True, keep_file=False)
+                else:
+                    LOG.debug("Removing all files in output file group %s ", grp)
+                    workspace.remove_file_group(grp, recursive=True, force=True, keep_files=False)
```
bertsky, line 71: As I have warned before, this is – unfortunately – not enough. PAGE files can reference derived image files in their respective file groups. You _must_ identify these. (It's probably worth doing this only once in a re-usable function. There's similar demand for derived image dependency tracking in other contexts.)

ocrd/ocrd/decorators.py
```diff
@@ -46,9 +56,25 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
...
+        if overwrite and 'page_id' in kwargs:
+            # If 'overwrite' and 'page_id' were given, it is possible the output
+            # file group is not completely empty
+            report = WorkspaceValidator.check_file_grp(workspace, kwargs['input_file_grp'], '')
+        else:
+            report = WorkspaceValidator.check_file_grp(workspace, kwargs['input_file_grp'], kwargs['output_file_grp'])
```
bertsky, line 77: Just a thought: Even _without_ `--overwrite`, if I merely want to process some page(s), I should not be prohibited to do so if the output file group already exists, but not for the page(s) I am interested in.

kba: We could check whether, provided `--page_id` is set, the output file group contains files representing those pages. It's doable but the effect would be the same as with `--overwrite` (if the pages are not in the output file group, nothing is changed). But I like that `--overwrite` conveys that the user is aware that the output file group already exists and that processing could result in data loss.

ocrd/ocrd/decorators.py
kba: Good point, I missed that completely. Can you pls provide a small-ish workspace with rich use of AlternativeImage for addition to assets so I can test? Thanks

bertsky: But how do I share it with you?

I recommend just running `gt-binarize-page-olena-sauvola-denoise-ocropy-deskew-page-ocropy-clip-deskew-region-tesseract-resegment-dewarp-ocr-ocropy-tesseract-extract-lines.mk` on `kant_aufklaerung` yourself.

ocrd/ocrd/decorators.py
bertsky: Exactly!

kba: "Exactly" as in 
  
  * "extend the functionality of the file group check to take `--page-id` into account and allow existing output file groups as long as they do not contain these specific pages", or
  * "force the user to acknowledge that his could potentially be a destructive action by providing the `--overwrite` flag, even if no files are actually overwritten, just an existing file group used as output"
?

bertsky: I don't see it as either-or. Yes, we should allow adding pages to existing file groups _without_ `--overwrite`. There's nothing destructive in that. The user should always have that safe option. By making the user always need `--overwrite`, we reduce its use to cases where the user _should_ pay attention.

bertsky:
One more thing: This also needs to be controlled by your new `overwrite` setting:

https://github.com/OCR-D/core/blob/5dcd52246aafd3ae13fb145bed5ec9f2a394a386/ocrd/ocrd/workspace.py#L784

And in the task validator, it would be great (if difficult) to also check whether any (explicit or implicit) image output file group already exists.

(Since I started this with `force=True` at a time when we had no output validators, we kept overlooking this, and practitioners often accidentally loose or invalidate their derived images by the next run.)

ocrd/ocrd/decorators.py
kba: I've added a PAGE API method `get_AllImagePaths` to get all image paths in a PAGE-XML document.

`Workspace.remove_file` now has two additional parameters `page_recursive` and `page_same_group`. If the former is `True` (the default), then on removal of a PAGE-XML file, it is also parsed for paths of `pc:Page/@imageFilename` and `pc:AlternativeImage/@filename` and those are recursively removed as well.

The idea behind the `page_same_region` parameter is to restrict deleting images to those that are in the same `pc:fileGrp` as the PAGE-XML document is, e.g. keeping the image referenced in `@imageFilename` if the PAGE-XML document is derived from another PAGE-XML document. However, I cannot test this, since I don't have any test data that follows this convention we agreed on in the last Tech Call (that AlternativeImages should be written to the fileGrp of the PAGE-XML that references them).

Also I'm interested in PAGE-XML samples with AlternativeImage on Word and Glyph level. If anybody can provide, I'd appreciate it!

ocrd/ocrd/decorators.py
kba: `WorkspaceValidator.check_file_grp` now accepts a `page_id` parameter. If provided, an exception is only raised if the output group already contains **AND** there are files in that output file group for that page/these pages.

kba:
> And in the task validator, it would be great (if difficult) to also check whether any (explicit or implicit) image output file group already exists.

This is not yet implemented, the algorithm is too complicated for me today. However, in the `Task.validate()` method, input and output file groups are now read from the `ocrd-tool.json` file and extend the provided input/output groups with defaults. So at least we now have all the information necessary to validate whether the task sequence is wired correctly.

`ocrd process` now also supports `--overwrite`, passing it on to processor calls.

ocrd/ocrd/decorators.py
bertsky: > The idea behind the `page_same_group` parameter is to restrict deleting images to those that are in the same `pc:fileGrp` as the PAGE-XML document is, e.g. keeping the image referenced in `@imageFilename` if the PAGE-XML document is derived from another PAGE-XML document. However, I cannot test this, since I don't have any test data that follows this convention we agreed on in the last Tech Call (that AlternativeImages should be written to the fileGrp of the PAGE-XML that references them).

Yes, this is a hen-or-egg problem. I think we could either start a feature branch in core and at least one preprocessing module to test this realistically, or add a pseudo-processor in the test section here (perhaps derived from the new dummy processor) which produces a dummy AlternativeImage via `image_from_segment` and subsequent `save_image_file`.

bertsky: > Also I'm interested in PAGE-XML samples with AlternativeImage on Word and Glyph level. If anybody can provide, I'd appreciate it!

Sorry, I don't know of any processor that creates derived images below the line level.

ocrd/ocrd/processor/base.py
```diff
@@ -104,6 +105,8 @@ def run_cli(
         args += ['--output-file-grp', output_file_grp]
     if parameter:
         args += ['--parameter', parameter]
+    if overwrite:
+        args += ['--overwrite', overwrite]
```
bertsky, line 109: IIUC this is not an option which itself takes an argument, but only a flag:
```suggestion
        args += ['--overwrite']
```

ocrd/ocrd/task_sequence.py
```diff
@@ -74,16 +83,18 @@ def __str__(self):
...
     report = ValidationReport()
     prev_output_file_grps = workspace.mets.file_groups
 
+    first_task = tasks.pop(0)
+    first_task.validate()
+
     # first task: check input/output file groups from METS
-    # TODO disable output_file_grps checks once CLI parameter 'overwrite' is implemented
-    WorkspaceValidator.check_file_grp(workspace, tasks[0].input_file_grps, tasks[0].output_file_grps, report)
+    WorkspaceValidator.check_file_grp(workspace, first_task.input_file_grps, first_task.output_file_grps, page_id, report)
```
bertsky, line 94: Since your `overwrite` solution is based on _removing_ existing output files right before processing (i.e. in `run_cli` / `ocrd_cli_wrap_processor`) but **after** task validation, you'll still have to exempt checking output file groups here:

```suggestion
    WorkspaceValidator.check_file_grp(workspace, first_task.input_file_grps, '' if overwrite else first_task.output_file_grps, page_id, report)
```

ocrd/ocrd/task_sequence.py
```diff
@@ -74,16 +83,18 @@ def __str__(self):
...
 
-    prev_output_file_grps += tasks[0].output_file_grps
-    for task in tasks[1:]:
+    prev_output_file_grps += first_task.output_file_grps
+    for task in tasks:
         task.validate()
         # check either existing fileGrp or output-file group of previous task matches current input_file_group
         for input_file_grp in task.input_file_grps:
```
bertsky, line 100: Cannot comment/suggest on non-context lines, but actually my focus is on L92 (now L103):
How about an extra check that output fileGrps do not exist yet (at least for relevant pages):
```python
        if not overwrite:
            WorkspaceValidator.check_file_grp(workspace, '', task.output_file_grps, page_id, report)
```

ocrd/ocrd/workspace.py
```diff
@@ -117,18 +119,31 @@ def download_file(self, f, _recursion_count=0):
...
             ocrd_file = self.mets.remove_file(ID)
+            if ocrd_file.mimetype == MIMETYPE_PAGE and page_recursive:
+                with pushd_popd(self.directory):
+                    ocrd_page = parse(self.download_file(ocrd_file).local_filename, silence=True)
+                    for img_url in ocrd_page.get_AllImagePaths():
+                        img_kwargs = {'url': img_url}
+                        if page_same_group:
+                            img_kwargs['fileGrp'] = ocrd_file.fileGrp
+                        for img_file in self.mets.find_files(**img_kwargs):
+                            self.remove_file(img_file, keep_file=keep_file, force=force)
```
bertsky, line 146: Very elegant!

bertsky: Wait, they way `get_AllImagePaths` is implemented (namely _both_ the original and all derived images), this would also remove the original image, right?

ocrd_models/ocrd_page_user_methods/get_AllImagePaths.py
```diff
@@ -0,0 +1,50 @@
...
+                ret += doc.xpath('/page:PcGts/page:Page/page:AlternativeImage/@filename', namespaces=NAMESPACES)
+            if region:
+                for class_ in PAGE_REGION_TYPES:
+                    ret += doc.xpath('//page:%sRegion/page:AlternativeImage/@filename' % class_, namespaces=NAMESPACES)
+            if line:
+                ret += doc.xpath('//page:TextLine/page:AlternativeImage/@filename', namespaces=NAMESPACES)
+            if word:
+                ret += doc.xpath('//page:Word/page:AlternativeImage/@filename', namespaces=NAMESPACES)
+            if glyph:
+                ret += doc.xpath('//page:Glyph/page:AlternativeImage/@filename', namespaces=NAMESPACES)
```
bertsky, line 48: Another excellent solution!

ocrd/ocrd/decorators.py
```diff
@@ -46,9 +56,21 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
...
+            for grp in kwargs['output_file_grp'].split(','):
+                if page_id:
+                    for one_page_id in kwargs['page_id'].split(','):
+                        LOG.debug("Removing files in output file group %s with page ID %s", grp, one_page_id)
+                        for file in workspace.mets.find_files(pageId=one_page_id, fileGrp=grp):
+                            workspace.remove_file(file, force=True, keep_file=False, page_recursive=True)
+                else:
+                    LOG.debug("Removing all files in output file group %s ", grp)
+                    workspace.remove_file_group(grp, recursive=True, force=True, keep_files=False, page_recursive=True, page_same_group=True)
+        report = WorkspaceValidator.check_file_grp(workspace, kwargs['input_file_grp'], kwargs['output_file_grp'], page_id)
```
bertsky, line 73: Great! (I always thought you were going to implement `overwrite` via some dynamic default for the `force` kwarg, but this solution is much better for validation and error handling.)

ocrd/ocrd/workspace.py
```diff
@@ -27,6 +28,7 @@
     pushd_popd,
     MIME_TO_EXT,
     MIME_TO_PIL,
+    MIMETYPE_PAGE
 )
```
bertsky, line 33: Cannot comment on non-context lines, but would like to refer to `save_image_file` at the end of the file:

Please don't forget to change its default `force=True` to `force=False`.

ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -25,27 +24,40 @@ class WorkspaceValidator():
...
         if output_file_grp:
             for grp in output_file_grp:
                 if grp in workspace.mets.file_groups:
-                    report.add_error("Output fileGrp[@USE='%s'] already in METS!" % grp)
+                    if page_id:
+                        for one_page_id in page_id:
+                            if workspace.mets.find_files(fileGrp=grp, pageId=one_page_id):
+                                report.add_error("Output fileGrp[@USE='%s'] already contains output for page %s" % (grp, one_page_id))
+                    else:
+                        report.add_error("Output fileGrp[@USE='%s'] already in METS!" % grp)
```
bertsky, line 60: Fantastic – it looks so simple now!

bertsky:
> > And in the task validator, it would be great (if difficult) to also check whether any (explicit or implicit) image output file group already exists.
> 
> This is not yet implemented, the algorithm is too complicated for me today.

Thinking about it, this might be next to impossible to do: You'd have to know what a processor wants to do with any of its output fileGrps, esp. which of them are going to end up as AlternativeImages, which could then get in conflict with existing METS files (but not necessarily under pageId). You'd also have to know the implicit output fileGrps.

So maybe we should just wait for the new "all-under-one-fileGrp" scheme? (But still, when using  `pageId` selectively, existing derived images which are below page level and thus have no pageId might get in conflict late during processing, after validation.)

ocrd/ocrd/task_sequence.py
```diff
@@ -57,12 +57,21 @@ def validate(self):
...
+        actual_input_grps = [*self.ocrd_tool_json['input_file_grp']]
+        for i, grp in enumerate(self.input_file_grps):
+            actual_input_grps[i] = grp
+        self.input_file_grps = actual_input_grps
         param_validator = ParameterValidator(self.ocrd_tool_json)
         report = param_validator.validate(parameters)
         if not report.is_valid:
             raise Exception(report.errors)
-        if 'output_file_grp' in self.ocrd_tool_json and not self.output_file_grps:
-            raise Exception("Processor requires output_file_grp but none was provided.")
```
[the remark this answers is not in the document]
kba: > existing semantics of **giving alternative examples** of how the input/output file group could by named.

Argh, you're right. We really need to fix this in the specs asap.

ocrd/ocrd/decorators.py
kba: Tracking in https://github.com/OCR-D/core/issues/505

ocrd/ocrd/task_sequence.py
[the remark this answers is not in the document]
kba: b5aaa3d0

ocrd/ocrd/task_sequence.py
kba: Makes sense, thx. 275a0a66

ocrd/ocrd/workspace.py
kba: Yes, should probably distinguish `page` into `page_filename` and `page_alternative`?

ocrd/ocrd/workspace.py
kba: Good point, thx. 0eaf5f82

kba:
> So maybe we should just wait for the new "all-under-one-fileGrp" scheme?

If we can make the information on input/outpup file groups in ocrd-tool.json more explicit and semantically richer, that would go a long way towards better introspection of workflow. It's still tricky to do and you cannot predict every behavior of a processor but wihtout proper information on input/output, it is just convention/heuristics.

ocrd/ocrd/workspace.py
bertsky: Actually, I don't think that `/PcGts/Page/@imageFilename` needs to be there at all. The original image is never "owned" by any PAGE fileGrp in our functional model. And you never need to access it outside of `image_from_page`.
But if you want to keep it for other use cases, then perhaps an additional boolean kwarg `include_original=False` (or similar) would be more intuitive.

kba: Getting the `@imageFilename` is trivial, so I changed the method to only return `AlternativeImage/@filename` where the load-xml-and-xpath approach really makes sense. ab91268e

ocrd/ocrd/decorators.py
```diff
@@ -46,9 +56,21 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
...
         resolver = Resolver()
         workspace = resolver.workspace_from_url(mets, working_dir)
-        # TODO once we implement 'overwrite' CLI option and mechanism, disable the
-        # `output_file_grp_ check by setting to False-y value if 'overwrite' is set
-        report = WorkspaceValidator.check_file_grp(workspace, kwargs['input_file_grp'], kwargs['output_file_grp'])
+        page_id = kwargs.get('page_id')
+        if overwrite:
+            if 'output_file_grp' not in kwargs or not kwargs['output_file_grp']:
+                raise Exception("--overwrite requires --output-file-grp")
+            LOG.debug("Removing files because of --overwrite")
```
bertsky, line 63: I am currently testing on real workflows. One thing that sticks out is that this log message is only _debug_ level, while the subsequent messages (up to 3 per file ID) are all _info_: 
- in `ocrd_models.ocrd_mets`: `remove_one_file` ...
- in `ocrd_models.ocrd_mets`: `Delete fptr element` ...
- in `ocrd.workspace`: `rm ` ...

So (to give these other messages at least some context) IMHO either this should be upgraded to _info_ as well, or the others downgraded to _debug_.

ocrd/ocrd/decorators.py
```diff
@@ -46,9 +56,21 @@ def ocrd_cli_wrap_processor(processorClass, ocrd_tool=None, mets=None, working_d
...
+            for grp in kwargs['output_file_grp'].split(','):
+                if page_id:
+                    for one_page_id in kwargs['page_id'].split(','):
+                        LOG.debug("Removing files in output file group %s with page ID %s", grp, one_page_id)
+                        for file in workspace.mets.find_files(pageId=one_page_id, fileGrp=grp):
+                            workspace.remove_file(file, force=True, keep_file=False, page_recursive=True)
+                else:
+                    LOG.debug("Removing all files in output file group %s ", grp)
+                    workspace.remove_file_group(grp, recursive=True, force=True, keep_files=False, page_recursive=True, page_same_group=True)
```
bertsky, line 72: As long as we don't have #505 in place, this must be set to false:

```suggestion
                    workspace.remove_file_group(grp, recursive=True, force=True, keep_files=False, page_recursive=True, page_same_group=False)
```

Otherwise processing **will** fail when derived images already exist:
```
Exception: File with ID='OCR-D-SEG-BLOCK-ocropus_00037190401554.0000_clip' already exists
```

ocrd/ocrd/decorators.py
bertsky:
```suggestion
            LOG.info("Removing files because of --overwrite")
```

ocrd/ocrd/decorators.py
bertsky, line 73: And there's yet **another problem**: When processing itself fails (for some reason), now the deletion is effective immediately in the filesystem, but the new METS does _not_ get written. Thus, stale fptrs remain which have to be removed with an extra `workspace prune-files`.

```suggestion
            workspace.save_mets()
        report = WorkspaceValidator.check_file_grp(workspace, kwargs['input_file_grp'], kwargs['output_file_grp'], page_id)
```

ocrd/ocrd/decorators.py
kba: That was accidentally left after testing. Even with #505 adapted by processors, it should not be the default, at least for the time being.

ocrd/ocrd/decorators.py
kba: Good catch. Failing processors is something we should test more.

bertsky:
I'd like to add another change along the lines of https://github.com/OCR-D/core/pull/507/commits/ae93f263204e08b74ec27dcb6a6564dcdae63f60, but cannot use the GH suggestions for that. I know this file/context is already in conflict with master now, so may I push directly to the PR and then we'll use GH's resolver took afterwards?

kba:
> I know this file/context is already in conflict with master now, so may I push directly to the PR and then we'll use GH's resolver tookl afterwards?

Sure.

bertsky:
> > > And in the task validator, it would be great (if difficult) to also check whether any (explicit or implicit) image output file group already exists.
> > 
> > 
> > This is not yet implemented, the algorithm is too complicated for me today.
> 
> Thinking about it, this might be next to impossible to do: You'd have to know what a processor wants to do with any of its output fileGrps, esp. which of them are going to end up as AlternativeImages, which could then get in conflict with existing METS files (but not necessarily under pageId). You'd also have to know the implicit output fileGrps.

This problem is still very real, and we **must** solve it (esp. since `save_image_file(force=False)`): As long as the existing output files contain derived images, we have to ensure to also delete those from the METS (under the "overwrite-as-pre-delete" regime). But here's the catch: we must _not delete_ all derived images there (which typically are also the results of previous annotation steps on the same or higher hierarchy levels), but _only the one to be overwritten_. However, only the processor knows for sure which derived images it wants to annotate. 

At the moment we do delete all derived images referenced in the output page – including results from binarization, clipping etc (which cannot be re-generated by core dynamically). And even worse, afterwards many other file groups won't be usable anymore.

We forgot to properly acknowledge that – regardless of #505 – derived images are never _owned_ by one file group alone. They will be referenced by multiple PAGE file groups at the same time. (Otherwise all derived images would have to be copied for each new output.)

Therefore, the whole "overwrite-as-pre-delete" principle should be re-examined. If we cannot know just what the processor wants to write, then we cannot use pre-deletion but have to use "overwrite-as-force-true".

kba:
After discussion this morning, I've disabled the delete-before-processing algorithm in `ocrd_cli_wrap_processor` and instead introduced an `overwrite_mode` parameter that globally sets `force=True` for all operations on the workspace where `force` can be set. The advantage of doing it this way is that processor implementations don't need to know about `force` or `overwrite`, if the user sets `--overwrite` it's a global override.

While not currently use, I left the `page_recursive`/`page_same_group` mechanics in place because I still think that the "delete-before-processing" approach is better in the long-run.

ocrd/ocrd/workspace.py
```diff
@@ -47,13 +47,15 @@ class Workspace():
         directory (string) : Folder to work in
         mets (:class:`OcrdMets`) : OcrdMets representing this workspace. Loaded from 'mets.xml' if ``None``.
         mets_basename (string) : Basename of the METS XML file. Default: Last URL segment of the mets_url.
+        overwrite_mode (boolean) : Whether to force add operations on this workspace globally
```
kba, line 50: and remove of course

bertsky:
Thank you! 

Perhaps you can make a small announcement on Gitter about `--overwrite` and `--page-id` and task validation, so others get to appreciate this as well?

kba:
[Done](https://gitter.im/OCR-D/Lobby?at=5ee2225a2cf2f36eae526dbf)

PR #472, opened 2020-04-29
bashlib: add logging methods, #366

kba:


kba:
I've added an `ocrd log` CLI like @bertsky suggested:

```sh
ocrd log info "hello world"
ocrd log error "This value is out of bounds: %s" "$value"
# the next two statements are equivalent
ocrd log --name foo.bar "yadda yadda"
OCRD_TOOL_NAME=foo.bar ocrd log  "yadda yadda"
```

Unfortunately, I'm stuck getting this to work:

```
ocrd -l DEBUG log debug "MSG"
```

Which should print out `MSG` but does not because the logger propagation voodoo I suspect. Therefore, I `setOverrideLogLevel` right before the actual logging, otherwise messages with level < INFO never appear. Need to investigate some more.

PR #474, opened 2020-04-29
💩 ocrd_page: export with pc: namespace

kba:
...by any means necessary

After exhausting all documented approaches to make the output of `OcrdPage.to_xml` use the `pc:` namespace prefix consistently, I resorted to another `sed` hack that seems to do the trick.

I will try to clear this up with upstream generateDS but for the time being this should at least support the use case https://github.com/OCR-D/core/issues/269#issuecomment-621338143 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,6 +92,8 @@ generate-page: repo/assets
 	# hack to ensure output has pc: prefix
 	@#sed -i "s/namespaceprefix_=''/namespaceprefix_='pc:'/" $(GDS_PAGE)
 	sed -i 's/_nsprefix_ = None/_nsprefix_ = "pc"/' $(GDS_PAGE)
+	# hack to ensure child nodes also have pc: prefix...
+	sed -i 's/.*_nsprefix_ = child_.prefix$$//' $(GDS_PAGE)
 
 #
 # Repos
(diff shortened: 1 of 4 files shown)


bertsky:
> Will this also work for parsing/deserialising and then re-generating/serialising PAGE that does _not_ have the prefix (or an older namespace altogether), like `kant_aufklaerung` in the assets repo?

No, not completely!!

I have tested several existing PAGE files. It can even convert older (2013, 2017) NS. But as soon as the input does _not_ have the `pc:` prefix, then it fails to generate the prefix – but only on the `*/TextEquiv/Unicode` element. (All other elements I searched are fine.)

kba:
> But as soon as the input does not have the pc: prefix, then it fails to generate the prefix – but only on the */TextEquiv/Unicode element. (All other elements I searched are fine.)

I think I've identified the reason for that particular behavior (namespace prefixes of child elements seem to be set to `child.prefix` even if that child is text content) and disabled it with another sed hack. See `test_to_xml_unicode_nsprefix` for loading a PAGE w/o namespace prefix and producing PAGE w/ namespace prefix.

bertsky:
> > But as soon as the input does not have the pc: prefix, then it fails to generate the prefix – but only on the */TextEquiv/Unicode element. (All other elements I searched are fine.)
> 
> I think I've identified the reason for that particular behavior (namespace prefixes of child elements seem to be set to `child.prefix` even if that child is text content) and disabled it with another sed hack. See `test_to_xml_unicode_nsprefix` for loading a PAGE w/o namespace prefix and producing PAGE w/ namespace prefix.

Ah, I see (I think). You are in for lots of generateDS bug reports!

Wonderful – this works for me as well. (But I'll keep testing just in case...)

kba:
Upstream issue: https://bitbucket.org/dkuhlman/generateds/issues/31/how-to-consistently-output-xml-with

PR #479, opened 2020-05-13
PageType.get_AllRegions to list all kinds of regions

kba:
First step to have a type-agnostic listing of all Regions, Text-, Table-, Graphics- etc.

Currently only available on the `Page` element, <del>ordering by reading order not yet implemeneted.</del> Is now implemented.

https://github.com/OCR-D/core/issues/240#issuecomment-493135797

Since we've been adding a few methods to the generateDS code now, I've refactored the code so that the methods can be defined in their own file with proper syntax highlighting etc..


ocrd_models/ocrd_page_user_methods.py
```diff
@@ -83,6 +83,26 @@ def show(self):
...
+        Returned in random order unless ``reading_order`` is set (NOT CURRENTLY IMPLEMENTED)
+        """
+        if reading_order:
+            raise NotImplementedError("Ordering of regions by Reading Order not currently Implemented :(")
+        if not regions:
+            regions = ['Advert', 'Chart', 'Chem', 'Custom', 'Graphic', 'Image', 'LineDrawing', 'Map', 'Maths', 'Music', 'Noise', 'Table', 'Text']
+        ret = []
+        for region in regions:
+            ret += getattr(self, 'get_{}Region'.format(region))()
+        return ret
```
bertsky, line 103: Have not tried it out, but how about this:

```suggestion
        if reading_order:
            reading_order = reading_order.get_OrderedGroup() or reading_order.get_UnorderedGroup()
        if reading_order:
            def get_recursive_reading_order(rogroup):
                if isinstance(rogroup, (OrderedGroupType, OrderedGroupIndexedType)):
                    elements = rogroup.get_AllIndexed()
                if isinstance(rogroup, (UnorderedGroupType, UnorderedGroupIndexedType)):
                    elements = (rogroup.get_RegionRef() + rogroup.get_OrderedGroup() + rogroup.get_UnorderedGroup())
                regionrefs = list()
                for elem in elements:
                    regionrefs.append(elem.get_regionRef())
                    if not isinstance(elem, (RegionRefType, RegionRefIndexedType)):
                        regionrefs.extend(get_recursive_reading_order(elem))
                return regionrefs
            reading_order = get_recursive_reading_order(reading_order)
        if reading_order:
            ret = dict([(region.id, region) for region in ret])
            return [ret[region_id] for region_id in reading_order if region_id in ret]
        else:
            return ret
```

Since the ReadingOrder tree is an injective mapping to the region tree, but we are only interested in the top-level regions here, we have to first get the full sequence and then "filter".

ocrd_models/ocrd_page_user_methods.py
```diff
@@ -117,9 +117,9 @@ def get_recursive_reading_order(rogroup):
             reading_order = get_recursive_reading_order(reading_order)
         if reading_order:
             ret = dict([(region.id, region) for region in ret])
-            return [ret[region_id] for region_id in reading_order if region_id in ret]
-        else:
-            return ret
+            ret = [ret[region_id] for region_id in reading_order if region_id in ret]
+        ret = [r in ret if r.__class__.__name__.replace('RegionType', '') in regions
```
bertsky, line 121: There's a problem with that line (missing closing bracket, no `for`).

kba: thx i am aware but had to leave the office now, will finish pr later from home, hence the "wip".

ocrd_models/ocrd_models/ocrd_page_generateds.py
```diff
@@ -2850,6 +2850,43 @@ def buildChildren(self, child_, node, nodeName_, fromsubclass_=False, gds_collec
...
+                    regionrefs = list()
+                    for elem in elements:
+                        regionrefs.append(elem.get_regionRef())
+                        if not isinstance(elem, (RegionRefType, RegionRefIndexedType)):
+                            regionrefs.extend(get_recursive_reading_order(elem))
+                    return regionrefs
+                reading_order = get_recursive_reading_order(reading_order)
+            if reading_order:
+                id2region = dict([(region.id, region) for region in ret])
+                ret = [id2region[region_id] for region_id in reading_order if region_id in id2region]
```
bertsky, line 2887: I see one further problem we did not think about before: What if not all regions are contained in `RegionOrder`? We could:
1. omit them from the result when running with `order='reading-order'`. (Current state of code), or
2. sort them at the beginning/end of the list, or
3. return them in an extra list (making the function two-valued), or
4. raise a (specialized) exception, returning both lists as members so this can be handled by the caller.

ocrd_models/ocrd_models/ocrd_page_generateds.py
```diff
@@ -5347,7 +5384,28 @@ def buildChildren(self, child_, node, nodeName_, fromsubclass_=False, gds_collec
...
+        ret = self.get_AllIndexed()
+        self.set_RegionRefIndexed([])
+        self.set_OrderedGroupIndexed([])
+        self.set_UnorderedGroupIndexed([])
+        return ret
+    
+    def add_AllIndexed(self, elements):
+        if not isinstance(elements, list):
+            elements = [elements]
+        for element in sorted(elements, key=lambda x : x.index):
```
bertsky, line 5400: Wouldn't it be prudent to do some `@index` consistency checking vodoo here (like no `elements` possess an already existing or mutually conflicting index)? If we have to raise an exception anyway, then the earlier (nearer to the cause) the better, wouldn't you agree?

ocrd_models/README.md
```diff
@@ -3,3 +3,30 @@
...
+def getFirstTextRegion(self)
+  return self.get_TextRegion[0]
+```
+
+2. Edit `ocrd_models/ocrd_page_user_methods.py` and append to the `METHOD_SPECS` list:
+
+```python
+METHOD_SPECS = (
+  # ...
+  _add_method(r'^PageType$', 'get_TextRegion')
```
bertsky, line 23:
```suggestion
  _add_method(r'^PageType$', 'get_FirstTextRegion')
```

ocrd_models/ocrd_page_user_methods.py
```diff
@@ -80,94 +82,27 @@ def show(self):
...
-        for element in sorted(elements, key=lambda x : x.index):
-            if isinstance(element, RegionRefIndexedType):
-                self.add_RegionRefIndexed(element)
-            elif isinstance(element, OrderedGroupIndexedType):
-                self.add_OrderedGroupIndexed(element)
-            elif isinstance(element, UnorderedGroupIndexedType):
-                self.add_UnorderedGroupIndexed(element)
-        return self.get_AllIndexed()
-''', class_names=r'^(OrderedGroupType|OrderedGroupIndexedType)$')
+def _add_method(class_re, method_name):
```
bertsky, line 85: Nice!

ocrd_models/README.md
```diff
@@ -3,3 +3,30 @@
...
 See https://github.com/OCR-D/core
+
+## Adding user methods to the generated PAGE API
+
+Let's say you want to add a method `get_FirstTextRegion` on the pc:Page element:
+
+1. Create a file `ocrd_models/ocrd_page_user_methods/get_FirstTextRegion.py`
+
+```python
+def getFirstTextRegion(self)
```
bertsky, line 14:
```suggestion
def get_FirstTextRegion(self):
```

ocrd_models/README.md
```diff
@@ -3,3 +3,30 @@
...
+  return self.get_TextRegion[0]
+```
+
```
bertsky, line 17:
```suggestion

   (Note that the method name and file name must be identical.)
```

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    if order not in ['document', 'reading-order']:
+        raise Exception("Argument 'order' must be either 'document' or 'reading-order', not '{}'".format(order))
+    if not classes:
+        classes = ['Advert', 'Chart', 'Chem', 'Custom', 'Graphic', 'Image', 'LineDrawing', 'Map', 'Maths', 'Music', 'Noise', 'Separator', 'Table', 'Text', 'Unknown']
+    def region_class(x):
+        return x.__class__.__name__.replace('RegionType', '')
+    ret = []
+    for region in classes:
+        ret += getattr(self, 'get_{}Region'.format(region))()
```
bertsky, line 17:
```suggestion
    def region_class(x):
        return x.__class__.__name__.replace('RegionType', '')
    def get_recursive_regions(regions, level):
        if level == 1:
            # stop recursion, filter classes
            if classes:
                return [r for r in regions if region_class(r) in classes]
            else
                return regions
        # find more regions recursively
        more_regions = []
        for region in regions:
            more_regions.append([])
            for class_ in ['Advert', 'Chart', 'Chem', 'Custom', 'Graphic', 'Image', 'LineDrawing', 'Map', 'Maths', 'Music', 'Noise', 'Separator', 'Table', 'Text', 'Unknown']:
                more_regions[-1] += getattr(region, 'get_{}Region'.format(class_))()
        if not any(more_regions):
            return get_recursive_regions(regions, 1)
        regions = [[r] + more for r, more in zip(regions, more_regions)]
        return get_recursive_regions(regions, level - 1 if level else 0)
    ret = get_recursive_regions([self], depth)
```

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
```diff
@@ -0,0 +1,37 @@
...
+                for elem in elements:
+                    regionrefs.append(elem.get_regionRef())
+                    if not isinstance(elem, (RegionRefType, RegionRefIndexedType)):
+                        regionrefs.extend(get_recursive_reading_order(elem))
+                return regionrefs
+            reading_order = get_recursive_reading_order(reading_order)
+        if reading_order:
+            id2region = dict([(region.id, region) for region in ret])
+            ret = [id2region[region_id] for region_id in reading_order if region_id in id2region]
+    ret = [r for r in ret if region_class(r) in classes]
```
bertsky, line 36:
```suggestion
```

ocrd_models/ocrd_models/ocrd_page_generateds.py
bertsky: @wrznr what would you say is most useful for writing clean and concise processors?

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
```diff
@@ -0,0 +1,37 @@
...
+        raise Exception("Argument 'order' must be either 'document' or 'reading-order', not '{}'".format(order))
+    if not classes:
+        classes = ['Advert', 'Chart', 'Chem', 'Custom', 'Graphic', 'Image', 'LineDrawing', 'Map', 'Maths', 'Music', 'Noise', 'Separator', 'Table', 'Text', 'Unknown']
+    def region_class(x):
+        return x.__class__.__name__.replace('RegionType', '')
+    ret = []
+    for region in classes:
+        ret += getattr(self, 'get_{}Region'.format(region))()
+    if order == 'reading-order':
+        reading_order = self.get_ReadingOrder().get_OrderedGroup() or reading_order.get_UnorderedGroup()
```
bertsky, line 19:
```suggestion
        reading_order = self.get_ReadingOrder()
        if reading_order:
            reading_order = reading_order.get_OrderedGroup() or reading_order.get_UnorderedGroup()
```

ocrd_models/ocrd_models/ocrd_page_generateds.py
kba: I think that both `@index` consistency as well as regions only partially listed in a RO are errors and should be raised in the PAGE validator. However, depending on how wide-spread these inconsistencies are, we should mitigate them in a way that no information is lost. That is not an issue here but it is for the region-not-in-RO problem above.

bertsky: > I think that both `@index` consistency as well as regions only partially listed in a RO are errors and should be raised in the PAGE validator.

My point was that the PAGE validator may be too late to be useful. For the case of regions missing in RO, I don't think this even _is_ an error (just a question of how the iterator should behave). And for inconsistent `@index` I believe this is so severe that we should not wait until serialization or even validation, but raise an exception right away (so offending processors get fixed soon), or even provide an API that does not allow making this mistake in the first place. 

For example, instead of `add_AllIndexed` (passing `@indexed` through after sorting) one could name the function `extend_AllIndexed` and rewrite all new elements such that their infimum is increased beyond the old elements' supremum (and then sorting).

ocrd_models/ocrd_models/ocrd_page_generateds.py
```diff
@@ -2853,22 +2853,40 @@ def __hash__(self):
     def get_AllRegions(self, classes=None, order='document'):
         """
         Get all the *Region element or only those provided by ``classes``.
-        Returned in random order unless ``reading_order`` is set (NOT CURRENTLY IMPLEMENTED)
+        Returned in document order unless ``order`` is ``reading-order`` is set (NOT CURRENTLY IMPLEMENTED)
```
bertsky, line 2898:
```suggestion
        Returned in document order unless ``order`` is ``reading-order``
```

kba:
@bertsky I've merged your proposals, refactored the inner functions to methods. Can you take a look and make sure I didn't botch the merge?

@wrznr @tboenig @cneud Can you weigh in on the `@index` consistency / "regions-not-listed-in-reading-order issue above, so we can finish and merge this soon?

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
```diff
@@ -1,22 +1,40 @@
 def get_AllRegions(self, classes=None, order='document'):
     """
     Get all the *Region element or only those provided by ``classes``.
-    Returned in random order unless ``reading_order`` is set (NOT CURRENTLY IMPLEMENTED)
+    Returned in document order unless ``order`` is ``reading-order`` is set (NOT CURRENTLY IMPLEMENTED)
```
bertsky, line 46:
```suggestion
    Returned in document order unless ``order`` is ``reading-order``
```

ocrd_models/ocrd_models/ocrd_page_generateds.py
kba: I have extended the `order` parameter to allow both solution 1 (now: `reading-order-only`) and solution 2 (`reading-order`, with non-reading-order regions added to the end of the list).

bertsky: That's even better – thanks!

ocrd_models/ocrd_models/ocrd_page_generateds.py
bertsky: @kba now you did rename from `add` to `extend` already, but the implementation is yet unchanged. Are you still going to? (If so, exceptions or index mitigation?)

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
```diff
@@ -0,0 +1,37 @@
+def get_AllRegions(self, classes=None, order='document'):
+    """
+    Get all the *Region element or only those provided by ``classes``.
+    Returned in random order unless ``reading_order`` is set (NOT CURRENTLY IMPLEMENTED)
+    Arguments:
+        classes (list) Classes of regions that shall be returned, e.g. ['Text', 'Image']
+        order ("document"|"reading-order") Whether to return regions sorted by document order (default) or by reading order
```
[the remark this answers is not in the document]
bertsky: Was there a particular reason for not including this suggestion? I'd really like to give that killer-use-case example you know...

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
```diff
@@ -0,0 +1,74 @@
...
+def _region_class(self, x): # pylint: disable=unused-argument
+    return x.__class__.__name__.replace('RegionType', '')
+
+def _get_recursive_regions(self, regions, level, classes=None):
+    if level == 1:
+        # stop recursion, filter classes
+        if classes:
+            return [r for r in regions if self._region_class(r) in classes]
+        # remove the first element (PageType)
+        return list(set(regions[1:]))
```
bertsky, line 11: What's the reason for this `list(set())` construct? It can't be to achieve the effect of `.copy()`, so I am guessing you want `set` to re-order by hash/ID?

tests/model/test_ocrd_page.py
```diff
@@ -184,5 +184,39 @@ def test_empty_groups_to_regionrefindexed(self):
...
+        https://github.com/OCR-D/core/pull/479
+        https://github.com/OCR-D/core/issues/240#issuecomment-493135797
+        """
+        with open('tests/model/TEMP1_Gutachten2-2.xml', 'r') as f:
+            pcgts = parseString(f.read().encode('utf8'), silence=True)
+            pg = pcgts.get_Page()
+            self.assertEqual(len(pg.get_AllRegions()), 45)
+            self.assertEqual(len(pg.get_AllRegions(depth=1)), 45)
+            self.assertEqual(len(pg.get_AllRegions(depth=2)), 65)
+            self.assertEqual(len(pg.get_AllRegions(depth=3)), 65)
```
bertsky, line 198: We should test infinite depth, too:
```suggestion
            self.assertEqual(len(pg.get_AllRegions(depth=3)), 65)
            self.assertEqual(len(pg.get_AllRegions(depth=0)), 65)
```

tests/model/test_ocrd_page.py
```diff
@@ -184,5 +184,39 @@ def test_empty_groups_to_regionrefindexed(self):
...
+        """
+        with open('tests/model/TEMP1_Gutachten2-2.xml', 'r') as f:
+            pcgts = parseString(f.read().encode('utf8'), silence=True)
+            pg = pcgts.get_Page()
+            with self.assertRaisesRegex(Exception, "Argument 'order' must be either 'document' or 'reading-order', not 'random'"):
+                pg.get_AllRegions(order='random')
+            self.assertEqual(len(pg.get_AllRegions(order='reading-order-only')), 20)
+            self.assertEqual(len(pg.get_AllRegions(order='reading-order-only', depth=2)), 40)
+            self.assertEqual(len(pg.get_AllRegions(order='reading-order', depth=1)), 45)
+            self.assertEqual(len(pg.get_AllRegions(order='reading-order', depth=2)), 65)
```
bertsky, line 217:
```suggestion
            self.assertEqual(len(pg.get_AllRegions(order='reading-order', depth=2)), 65)
            self.assertEqual(len(pg.get_AllRegions(order='reading-order', depth=0)), 65)
```

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
kba: To remove duplicates, without it, I get some ~1000 regions for depth 10, ~1500 for depth 11 etc.

tests/model/test_ocrd_page.py
kba: But with `depth==0`, no recursion happens, return list would be empty.

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
bertsky: Oh, right. But then the recursion itself is wrong. Let's change the loop such that the new regions are always derived from the _previous_ new regions only.

tests/model/test_ocrd_page.py
bertsky: That's because you modified my original suggestion for how to _enter_ the loop from `depth + 1 if depth else 0` to just `depth + 1`.

But apparently this would not have worked, because there was also an infinity loop – see above.

ocrd_models/ocrd_models/ocrd_page_generateds.py
kba: index mitigation, implementation will follow in this PR, it's a bit tricky.

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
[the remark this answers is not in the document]
kba: Was an oversight, I'll add the docs, thanks.

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
kba: Yes, I'm fixing it now.

tests/model/test_ocrd_page.py
kba: What was the purpose of `depth +1 if depth else 0`? The else clause would only be triggered if `depth == 0` or `depth == None`, right? So you want `depth == 0` to signify maximum depth, correct?

bertsky: > What was the purpose of `depth +1 if depth else 0`? 
> The else clause would only be triggered if `depth == 0` or `depth == None`, right? So you want `depth == 0` to signify maximum depth, correct?

Correct. The backend function discerns between always zero (termination only controlled by empty `more_regions`) and count-down (terminating at `level==1`). So the caller has to pass zero for infinite recursion but add an offset of 1 to ensure at least one level of recursion.

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
```diff
@@ -0,0 +1,86 @@
...
+            reading order with regions not in the reading order at the end of the
+            returned list (``reading-order``) or regions not in the reading order
+            omitted (``reading-order-only``)
+        depth (int) Recursive depth to look for regions at, set to `0` for all regions at any depth. Default: 0
+
+    For example, to get all text anywhere on the page in reading order, use:
+    ::
+        '\n'.join(line.get_TextEquiv()[0].Unicode
+                  for region in page.get_AllRegions(classes='Text', depth=0, order='reading-order')
+                  for line in region.get_TextLine())
```
bertsky, line 61:
```suggestion
        '\\n'.join(line.get_TextEquiv()[0].Unicode
                  for region in page.get_AllRegions(classes=['Text'], depth=0, order='reading-order')
                  for line in region.get_TextLine())
```

tests/model/test_ocrd_page.py
bertsky: That's erfect – thanks!

ocrd_models/ocrd_page_user_methods/get_AllRegions.py
bertsky: (I think quoting the backslash will help it survive in the docstring itself.)

ocrd_models/ocrd_models/ocrd_page_generateds.py
kba: `extend_AllIndexed(elements, validate_continuity=False` now increments the index of all elements passed. If `validate_continuity` is `True`, validates that lowest element `@index` is greater than the highest existing index (though no check for contiguousness), sorts elements by index and adds them (i.e. the previous behaviour plus validation). Otherwise (the default), elemetns are not sorted and added with incrementing `@index`.

ocrd_models/ocrd_page_user_methods/extend_AllIndexed.py
```diff
@@ -1,8 +1,22 @@
-# pylint: disable=line-too-long,invalid-name,missing-module-docstring,missing-function-docstring
-def extend_AllIndexed(self, elements):
+# pylint: disable=line-too-long,invalid-name,missing-module-docstring
+def extend_AllIndexed(self, elements, validate_continuity=False):
+    """
+    Add all elements in list ``elements``, respecting ``@index`` order.
+    """
```
bertsky, line 5:
```suggestion
    With ``validate_continuity``, check that all new elements come after all old elements
    (or raise an exception). 
    Otherwise, ensure this condition silently (by increasing ``@index`` accordingly).
    """
```

ocrd_models/ocrd_page_user_methods/get_UnorderedGroupChildren.py
```diff
@@ -0,0 +1,7 @@
+def get_UnorderedGroupChildren(self):
+    """
+    List all non-metadata children of an UnorderedGroup
+    """
+    # TODO: should not change order
```
bertsky, line 5: Note: IMO it's just not possible to keep the order after parsing. There's no descriptive datum that could preserve the XML ordering of the unordered group. – If that's what you meant by the Todo.

kba: I guess you're right, it is an **un**ordered group, after all. But still, for reproducability and non-invasiveness, I would prefer preserving the document order. It's not a blocking issue tho.

PR #483, opened 2020-05-14
export all generateDS types from ocrd_page, fix #480

### LINKED ISSUE #480: ocrd_models.ocrd_page: MetadataItemType gone
It seems that with the last generateDS update, `MetadataItemType` disappeared from `ocrd_page` (and is now under `ocrd_page_generateds` only) – maybe others, too.

kba:


ocrd_models/ocrd_models/ocrd_page.py
```diff
@@ -39,25 +70,58 @@
...
+    GridType,
     ImageRegionType,
     LabelType,
     LabelsType,
+    LayerType,
+    LayersType,
+    LineDrawingRegionType,
+    MapRegionType,
     MathsRegionType,
+    MetadataItemType,
```
bertsky, line 95: This was the culprit (not importing what was exported via `__all__`). But I am fine with exporting everything, too.

kba: That way, I hope I will realize to update the exports/__all__ whenever generateDS changes.

PR #484, opened 2020-05-14
Mets find multiple page selector

### LINKED ISSUE #463: process multiple page_ids conforming to spec
The OCR-D specifications currently [state](https://ocr-d.de/en/spec/cli#-g---page-id-id) that the `--page-id` CLI parameter is multi-valued, with comma as separator.

But core (and every processor based on it) does not support this: the Python decorators merely pass the value to `ocrd_models.ocrd_mets.OcrdMets.find_files`, [where](https://github.com/OCR-D/core/blob/e0ee7a12c0c52a14eb04b8add9b673314cbcc109/ocrd_models/ocrd_models/ocrd_mets.py#L150) it does not get comma-separated but directly passed into an XPath `@ID="%s"` expression. And bashlib has no multi-value provisions (like array conversion) whatsoever, and also merely delegates to `find_files` verbatim, so e.g. ocrd-olena-binarize and ocrd-im6convert don't support that either.

### LINKED ISSUE #482: performance of `page-id` on large workspaces
I am afraid my conjecture in [#448](https://github.com/OCR-D/core/pull/448#pullrequestreview-393997881) that we could become slower than before has now proven correct. I have a workspace here with more than 8000 pages (of a training corpus) times 18 fileGrps, and it became **unacceptably slow**. (I can provide the METS on request.)

Here is a little statistics:

| | **v2.6.0** | **#481 and #484** |
|---|---|---|
| `time ocrd workspace find -k url` | 7sec | 6sec |
| `time ocrd workspace find -k pageId` | 193min | 4min |
| `time ocrd workspace find -g page0001 -k url` | 261min | 2sec |

(The same holds for processors, so it's not only the `ocrd` CLI.)

Looking at [lxml's performance recommendations](https://lxml.de/performance.html), I find the following advice relevant:
> ## Tree traversal
> Another important area in XML processing is iteration for tree traversal. If your algorithms can benefit from step-by-step traversal of the XML tree and especially if few elements are of interest or the target element tag name is known, the .iter() method is a good choice
> ...
> In general, lxml is very fast for iteration, but loses ground against cET when many Elements are found and need to be instantiated. So, the more selective your search is, the faster lxml will run.
> ## XPath
> ...
> lxml provides more than one way of accessing it and you should take care which part of the lxml API you use. The most straight forward way is to call the xpath() method on an Element or ElementTree
> This is well suited for testing and when the XPath expressions are as diverse as the trees they are called on. However, if you have a single XPath expression that you want to apply to a larger number of different elements, the XPath class is the most efficient way to do it
> Note that this still allows you to use variables in the expression, so you can parse it once and then adapt it through variables at call time. In other cases, where you have a fixed Element or ElementTree and want to run different expressions on it, you should consider the XPathEvaluator
> ...
> Note that tree iteration can be substantially faster than XPath if your code short-circuits after the first couple of elements were found. The XPath engine will always return the complete result set, regardless of how much of it will actually be used.
> Here is an example where only the first matching element is being searched, a case for which XPath has syntax support as well

So I believe we should replace the large `.xpath('//mets:file')` with an `.iter(tag='mets:file')`, and at least factor the smaller `.xpath()` for `pageId` out of the loop body.

The latter I already did in #481, with stunning effect (see above).

For the `-k pageId` which fires `ocrd_mets.OcrdMets.get_physical_page_for_file()` I still don't know how to do better. Maybe when we know for `find_files` that we are also interested in pageIds, we could add a second main loop over `/mets:mets/mets:structMap[@TYPE="PHYSICAL"]/mets:div[@TYPE="physSequence"]/mets:div[@TYPE="page"]` mapping their `./mets:fptr/@FILEID` to their `./@ID`?

**EDIT** did the latter in #484 and updated the timings.

@kba I would consider this fixed, unless you want to follow through with `iter()`.

bertsky:
Fix #463 and fix #482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Changed:
+
+  * :fire: OcrdMets.physical_pages is now a method not a property, #484
+  * OcrdMets/ocrd workspace find: Search for multiple pages by fileid, ht @bertsky, #463, #484
+
 ## [2.6.1] - 2020-05-14
 
 Changed:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 Changed:
 
-  * :fire: OcrdMets.physical_pages is now a method not a property, #484
+  * `OcrdMets.get_physical_pages` method companion to `OcrdMets.physical_pages` a property, #484
   * OcrdMets/ocrd workspace find: Search for multiple pages by fileid, ht @bertsky, #463, #484
 
 ## [2.6.1] - 2020-05-14
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -184,6 +184,8 @@ def workspace_find(ctx, file_grp, mimetype, page_id, file_id, output_field, down
     """
     Find files.
     """
+    clobbered = False
+    ret = list()
     workspace = Workspace(ctx.resolver, directory=ctx.directory, mets_basename=ctx.mets_basename)
     for f in workspace.mets.find_files(
             ID=file_id,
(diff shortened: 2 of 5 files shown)


bertsky:
We also could use new tests securing #463 (e.g. querying `-g page7,page9`), and possibly also #482 (e.g. querying `-k ID -k pageId -k mimetype` where pageId gets computed in a separate main loop for efficiency and then sorted in again), or even both in combination.

Note 1: I had to change `OcrdMets.physical_pages` from a property to a function. I hope this is okay in general.

Note 2: That function is still not as efficient as it could be, because it cannot shortcut the inner loop when it finds a fptr for the page. The reason is that `for_fileIds` could contain multiple files for the same page. This may seem minor, but makes the difference between 4min and 7sec for my large workspace...

bertsky:
> Note 1: I had to change `OcrdMets.physical_pages` from a property to a function. I hope this is okay in general.

Just noticed myself that this is a **breaking change** of the API and breaks some processors. So I recommend against it myself. This is not worth it. We could just offer a new, alternative interface as a pure function for the file ID filter.

kba:
>  and breaks some processors. 

AFAICS the property is used in processor code directly only in ocrd_segment, dinglehopper and ocrd_cor_asv_ann, so we could probably mitigate any breaking because these projects are maintained by sensible quick-to-react people. However, we would need to release a new major version.

The easiest solution would be to keep `physical_pages` as a property, deprecate it and create a new method `get_physical_pages(self, for_fileId)`. Would that be okay with you?

bertsky:
> However, we would need to release a new major version.

Exactly. And so, like I said, not really worth it.



> The easiest solution would be to keep `physical_pages` as a property, deprecate it and create a new method `get_physical_pages(self, for_fileIds)`. Would that be okay with you?

Except for the deprecation (since there's nothing wrong with the unconditional implementation and it already has its own use-cases): perfectly okay.

Should I make the adjustments to this PR, or do you want to do it yourself?

kba:
> Should I make the adjustments to this PR, or do you want to do it yourself?

I'll update the PR (sans the deprecation).

PR #492, opened 2020-05-27
coordinates_for_segment: add optional parent_element…

### LINKED ISSUE #489: ocrd_utils.coordinates_for_segment: clip to parent?
We have the utility function `coordinates_for_segment` for conversion of polygons from some derived image and its coordinate transform to absolute coordinates of the original image.

This is all the function does, but of course the __expectation__ is more: Whereever that polygon was in the derived image, give me coordinates which I can annotate as final result of segmentation! (And unfortunately, that is also what it is mostly used for.)

However, that would entail also doing the inverse of `image_from_polygon`, i.e. clipping (or rather masking) the polygon to its _parent element_. This usually means ensuring the child element's polygon stays within the parent element's polygon. Except for the top-level parent (`Page` when there is no `Border`), where this means clipping coordinates to a range between zero and image size.

Thus, not having that functionality in core means segmenters will tend to produce inconsistent coordinates (failing the PAGE validator), and sometimes (e.g. when the page was rotated) even negative coordinates (failing XML validators and other PAGE libraries).

Unfortunately, we already made the first argument the segment's __polygon__ instead of its __element__ (as in `coordinates_of_segment` – an ugly asymmetry which itself originated from PAGE's `Border` idiosyncrasy). So we cannot silently enrich `coordinates_for_segment` by looking into the segment's `parent_object_` and clipping the polygon respectively. 

Or can we? How about making it polymorphic and checking the type of the first argument, as an opt-in for implicit conversion consistent with parent? To me that seems much better than providing a new function and trying to find processors which need it. We could even transform the function to be exclusively for the new usage by starting to deprecate the old usage at a later point and hunting it down.

@kba @wrznr if you give me your thumbs up I will kick this off with a PR.

bertsky:
…to ensure that the given polygon is fully contained
within that parent's polygon.

Fixes #489. 

But please don't merge yet, I have another, better idea

### DIFF
--- a/ocrd_utils/ocrd_utils/__init__.py
+++ b/ocrd_utils/ocrd_utils/__init__.py
@@ -4,7 +4,7 @@ Utility functions and constants usable in various circumstances.
 * ``coordinates_of_segment``, ``coordinates_for_segment``
 
     These functions convert polygon outlines for PAGE elements on all hierarchy
-    levels below page (i.e. region, line, word, glyph) between relative coordinates
+    levels (i.e. page, region, line, word, glyph) between relative coordinates
     w.r.t. a corresponding image and absolute coordinates w.r.t. the top-level image.
     This includes rotation and offset correction, based on affine transformations.
     (Used by ``Workspace`` methods ``image_from_page`` and ``image_from_segment``)
@@ -96,6 +96,8 @@ __all__ = [
     'points_from_x0y0x1y1',
     'points_from_xywh',
     'points_from_y0x0y1x1',
+    'points_of_segment',
+    'polygon_for_parent',
     'polygon_from_bbox',
     'polygon_from_points',
     'polygon_from_x0y0x1y1',
@@ -131,6 +133,7 @@ import contextlib
 
 import numpy as np
 from PIL import Image, ImageStat, ImageDraw, ImageChops
+from shapely.geometry import Polygon
 
 from .logging import * # pylint: disable=wildcard-import
 from .constants import *  # pylint: disable=wildcard-import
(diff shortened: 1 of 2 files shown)


ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -208,24 +213,32 @@ def coordinates_for_segment(polygon, parent_image, parent_coords):
        opposite direction, rotated purely, and translated back;
        the latter involves an additional offset from the increase
        in canvas size necessary to accommodate all points).
+    
+    If ``parent_element`` is given, then also see to it that the
+    resulting polygon is fully contained within the polygon of its parent
+    (via intersection). Note that this need not be the same element
+    that was used as "parent" for the image (the latter can be multiple
+    levels up).
```
kba, line 221: :+1:

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -547,6 +560,46 @@ def points_from_x0y0x1y1(xyxy):
...
+    or ``polygon`` unchanged otherwise.
+    """
+    childp = Polygon(polygon)
+    parentp = Polygon(polygon_from_points(points_of_segment(parent)))
+    if childp.within(parentp):
+        return polygon
+    interp = childp.intersection(parentp)
+    if interp.is_empty:
+        # FIXME: we need a better strategy against this
+        raise Exception("intersection of would-be segment with parent is empty")
```
kba, line 598: Do we want to be robust against such blatant errors?

ocrd_utils/requirements.txt
```diff
@@ -1,2 +1,3 @@
 Pillow >= 6.2.0
 numpy >= 1.17.0
+shapely
```
kba, line 3: Best to be specific with the versions

kba (approved):

ocrd_utils/ocrd_utils/__init__.py
bertsky: Yes, probably just show a warning. But we have no context for localizing the message at that point...

ocrd_utils/requirements.txt
bertsky: Mine is currently at `1.6.4.post2`. But I haven't had any problems with Shapely so far, ever.

PR #498, opened 2020-06-06
ocrd workspace add: check file exists with -C/--check-file-exists

### LINKED ISSUE #495: Alert when adding non-existing local file
Hello,
I'm wondering if it's really intended behavior that a non-existing local file is silently added to metsfile by `ocrd workspace add`:

```
ocrd workspace add -m "image/jpg" -G 'MAX' -i "IMG_MAX_1246734" "data/IMG_MAX_1246734.jpg"

=> mets.xml:

<mets:FLocat LOCTYPE="OTHER" OTHERLOCTYPE="FILE" xlink:href="data/IMG_MAX_1246734.jpg"/>
```
Is there a reason to do so? 
I'd expect an Exception in that case to halt ocr-ing, since any subsequent steps produce no valid output but do not complain either.

Version: OCRD V2.7.1
OS: Ubuntu 18.04 LTS

kba:
fix #495 @M3ssman Can you try please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Changed:
+
+  * `ocrd workspace add` supports `-C`/`--check-file-exists` to validate that `FNAME` is an existing local file, #495
+
 ## [2.8.1] - 2020-06-06
 
 Changed:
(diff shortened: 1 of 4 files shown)


M3ssman:
Works as expected. 
Without flag, nothing changed, but with `-C` or `--check-file-exists` it yields `Error: File does not exists <name>`. 

@kba 
Can you put the message to the logger? 
Please, if possible, replace 
`print("ERROR: File does not exists %s" % fname)` 
with 
`log.error("File '%s' does not exist, halt execution!" % fname)`

Since our ocrd-calls are part of other workflows and in case of trouble I only care for logs.

kba:
> Can you put the message to the logger?

Sure, I originally did but the log output isn't captured in the unit tests. Currently investigating why, then I'll update the PR.

kba:
> log output isn't captured in the unit tests

It's an issue with click's CliRunner where its capturing of `sys` interferes with pytest's capturing. I created a new `CapturingTestCase` class that makes it easier to make assertions on the behavior of our CLI interfaces. We should probably move all CLI tests to use that mechanism because of how important logging and consistent CLI are.

Non-existing files are now logged as errors with the format you proposed.

PR #507, opened 2020-06-08
document regex filtering on CLI

### LINKED ISSUE #458: allow regexp matching in workspace operations
IMO at least the following operations (both in the CLI and API) on METS deserve allowing regular expression matching:

- `remove_file`: instead of a fixed ID to look for
- `remove_file_group`: instead of a fixed ID to look for
- `find_files`: both MIME type (e.g. just `image/`) and ID
- `add_file_group`: for both file names and file IDs (possibly with back-references), cf. [discussion here](https://github.com/OCR-D/core/pull/428#discussion_r370615499)

To minimise possible ambiguity between a verbatim string and regex interpretation, while still keeping the existing argument/option names (and not introducing a new flavour each time), I recommend either using _POSIX Basic Regular Expression_ syntax (which is perhaps hard to get by in Python) or allowing some kind of extra notation in the input, e.g. a `re:` prefix.

bertsky:
Fixes #458 (even more)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Fixed:
+
+  * workspace.remove_file: fix for list-valued results, #507
+
+Changed:
+
+  * workspace prune-files CLI: support filtering (like workspace find), #507
+  * workspace CLI: update help strings (documenting regex support), #507
+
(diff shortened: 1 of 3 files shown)


kba:
The `prune-file` change led to an failing unit test. Can you investigate or shall I?

bertsky:
> The `prune-file` change led to an failing unit test. Can you investigate or shall I?

Sorry about that. Should be repaired now.

PR #508, opened 2020-06-08
bagit profile url should use ocr-d.de

kba:


kba:
Thinking about it, this will break a lot of things it should not.

PR #514, opened 2020-06-14
Smarter params

kba:
* `-p` can be repeated and the values will be merged right-to-left to allow for simple "inheritance" of JSON files/strings
* JSON can contain comments (same syntax as in Python/Perl/Bash but without support for trailing comments)

c.f. OCR-D/spec#158, OCR-D/spec#166

Couldn't find the branch where I worked on this, so here's a new attempt:

Processors accept a `-P/--parameter-override KEY VAL` argument that takes precedence over `-p/--parameter JSON_FILE_OR_STRING`. The aim is to simplify passing parameters to proceessors on the command line. 

E.g.
```sh
ocrd-whatever -p defaults.json -p '{"foo": "one"}' -P foo two
```

Will merge `{"foo": "one"}` into the JSON in `defaults.json`, then override the `foo` entry with value `two`.

<del>Missing yet is type handling. Currently this only works with strings. Not sure yet where to implement type casting. Or perhaps just parse `VAL` as JSON, which would require passing strings as `-P foo '"bar"'`?</del> <ins>Both, c.f. https://github.com/OCR-D/spec/pull/166</ins>

Now also implemented in `ocrd ocrd-tool tool parse-params` and `bashlib`, i.e. you can do

```sh
ocrd-im6convert -I FOO -O BAR -P output-format image/png
```


ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -784,28 +785,38 @@ def xywh_from_points(points):
...
+                value_parsed = parse_json_string_with_comments(value.strip())
+            except OSError:
+                value_parsed = parse_json_string_with_comments(value.strip())
+            if not isinstance(value_parsed, dict):
+                err = ValueError("Not a valid JSON object: '%s' (parsed as '%s')" % (value, value_parsed))
+        except json.decoder.JSONDecodeError as e:
+            err = ValueError("Error parsing '%s': %s" % (value, e))
+        if err:
+            raise err       # pylint: disable=raising-bad-type
+        ret = {**ret, **value_parsed}
```
bertsky, line 821: wow, haven't ever seen that before. Python _can_ be quite elegant!

kba: Destructuring is a ncie feature, I wish Python made even more use of it (probably does in newer versions). In JS you can write crazy concise code, define local variables based on object keys and much more.

Issue #517, opened 2020-06-15
workspace CLI: mets_basename is ignored


bertsky:
`ocrd workspace` announces itself with the following options to control where the workspace's local METS is located:
```
 -d, --directory WORKSPACE_DIR  Changes the workspace folder location.
                                 [default: .]
  -M, --mets-basename TEXT       The basename of the METS file.  [default:
                                 mets.xml]
```

So when it is not under CWD or not called `mets.xml`, I must use these options.

First, allow me to point out that this is in contrast to all the processor CLIs which have but a single (and more memorable) option for both at the same time:
```
  -m, --mets URL-PATH             URL or file path of METS to process
```

However, the workspace CLI does not always _implement_ the `--mets-basename` option correctly:
- some commands respect it: `init`, `clone`, `add-file`, `bulk-add`, `remove-file`, `remove-group`, `prune-files` and `set-id`
- other commands ignore it: `validate`, `list-group`, `list-page` and `get-id`

So, when my METS uses a different basename, the latter commands will just fail like this:
```
Exception: File does not exist: mets.xml
```

This can easily be fixed of course. 

But how about _also_ adding a `-m, --mets` option as an alternative? (The implementation would derive the directory and basename from it, but complain when both options are used inconsistently.)

kba:
`mets_basename` was a bad idea from the beginning and I'd be happy to be rid of it. So, I'd rather just replace `-M/--mets-basename` with `-m/--mets throughout. 

>  but complain when both options are used inconsistently

In other words:

  * If only `--directory` is set and `--mets` is not: Assume `--mets` to be `--directory/mets.xml`
  * If only `--mets` is set and a filename: If `--mets` has a dirname, set `--dirname` to `dirname(--mets`)
  * If both `--directory` and `--mets` are set: If `--mets` has a dirname: validate that `--dirname == dirname(--mets)`. Otherwise, prepend `--dirname` to `--mets`.

This way `--mets` can function like `--mets-basename` depending on `--directory`, right?

Should of course be tested but since this doesn't touch on processors, it wouldn't require a spec change and reduce complexity, so :+1:.

bertsky:
> `mets_basename` was a bad idea from the beginning and I'd be happy to be rid of it. So, I'd rather just replace `-M/--mets-basename` with `-m/--mets` throughout.

I see.

(So as with the processor CLIs, `--directory` would always be strictly local, whereas `--mets` could also be remote, in which case the METS gets cloned into `directory` first – right?)

[...]

PR #518, opened 2020-06-15
workspace CLI: respect --directory / --mets-basename

bertsky:
In partial fulfillment of the solutions discussed for #517: 
- still under the old interface (only `--directory` and `--mets-basename`, no `--mets`)
- make self-consistent (as a first step)
- don't break remote URL argument for `validate` (which should naturally be dropped in favour of `--mets`) yet, but make optional (using local METS instead of `'./mets.xml'` by default)
- don't break explicit directory argument for `clone` and for `init` (which should naturally be dropped in favour of `--directory`) yet, but make optional (using `--directory` setting instead of `'.'` by default)

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -88,18 +88,20 @@ def validate_workspace(ctx, mets_url, download, skip, page_textequiv_consistency
 @click.option('-f', '--clobber-mets', help="Overwrite existing METS file", default=False, is_flag=True)
 @click.option('-a', '--download', is_flag=True, help="Download all files and change location in METS file after cloning")
 @click.argument('mets_url')
+# should be deprecated:
 @click.argument('workspace_dir', default=None, required=False)
 @pass_workspace
 def workspace_clone(ctx, clobber_mets, download, mets_url, workspace_dir):
     """
-    Create a workspace from a METS_URL and return the directory
+    Create a workspace from METS_URL and return the directory
 
     METS_URL can be a URL, an absolute path or a path relative to $PWD.
 
-    If WORKSPACE_DIR is not provided, use the current working directory
+    If WORKSPACE_DIR is not provided, the new workspace will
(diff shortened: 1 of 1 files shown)


bertsky:
Regarding said breaking changes (in `validate`, `init` and `clone`), is there a way we can start the transition by CLI deprecation in `click`?

AFAICS, among ocrd_all: 
- `workspace clone` is used with a DIRECTORY argument in ocrd_keraslm and ocrd_olena,
- `workspace init` is used with a DIRECTORY argument (`.`) in ocrd_cis, ocrd_olena and workflow-configuration, and
- `workspace validate` is used with a METS_URL argument in assets

kba:
>  is there a way we can start the transition by CLI deprecation in click?

I'll investigate.

> * workspace clone is used with a DIRECTORY argument in ocrd_keraslm and ocrd_olena,
> * workspace init is used with a DIRECTORY argument (.) in ocrd_cis, ocrd_olena and workflow-configuration, and

These should be handled gracefully. However, since we know the authors of the mentioned projects to be quick-to-react, should be adapted to use `--directory` nonetheless. Would also be a good point to introduce `ocrd__minversion` to ocrd_olena.


> workspace validate is used with a METS_URL argument in assets

This should not block.

bertsky:
> > is there a way we can start the transition by CLI deprecation in click?
> 
> I'll investigate.

I saw [this](https://stackoverflow.com/a/50402799) yesterday, but I'd rather like to delegate to the existing `ocrd_utils.deprecated_alias` (including a warning on std logger)...



> should be adapted to use `--directory` nonetheless. Would also be a good point to introduce `ocrd__minversion` to ocrd_olena.

Absolutely. 

Or, if we want to change/extend the CLI from `--mets-basename` to `--mets` anyway, we could implement this first, and then drag along all the modules.

> This should not block.

I merely fgrepped in ocrd_all, but there might be more such usage...

Issue #519, opened 2020-06-16
Possible regression of "ocrd workspace add" in 2.10.2


kba:
@M3ssman in https://gitter.im/OCR-D/Lobby?at=5ee8b1e77c64f31f1163ea9b :

> I just met some ... different behavior. When I add a local file to a fresh workspace with ocrd V2.10.2, the Image file itself is not copied anymore to the directory with the according fGroup-Label, like it did just 2 weeks before ... even the brand new existance-check has nothing to complain about (but it complains quite fine if the file doesn't exist). So if I type `ocrd workspace add -m image/jpg -G MAX -i IMG_MAX_1818975 -C /home/hartwig/Projekte/work/mlu/ulb/ulb-it-vd-ocr/data/IMG_MAX_1818975.jpg` only the mets-file is updated and nothing else. Did I miss something?

bertsky:
@kba I believe the problem is here:
https://github.com/OCR-D/core/blob/86f788aed67f1ad0e7cf61141baa818f6e8dd504/ocrd/ocrd/cli/workspace.py#L158

The `ctx.directory` is converted into an absolute pathname in the context object's constructor. Thus, if you call with an absolute path name into the workspace, nothing happens

kba:
@M3ssman Is `/home/hartwig/Projekte/work/mlu/ulb/ulb-it-vd-ocr/data/` the workspace directory? And you want the file to be copied to `home/hartwig/Projekte/work/mlu/ulb/ulb-it-vd-ocr/data/MAX /IMG_MAX_1818975.jpg`? This does not work with the current implemntation because that file is already in the workspace, though not at the location you expect.

We could change the behavior to make the check not only if the file is in the workspace but also if it is at the location that the parameters imply. Hoever in that case, you would end up with copies of the files.

Would it not be easier to either put the files in the right subfolder in the first place or to copy the files from outside the workspace?

M3ssman:
No, the files are completely outside. 
`ulb-it-vd-ocr/data/` is just a folder with image data. The actual OCR-D-workspaces are located in `ulb-it-vd-ocr/TMP`, like `ulb-it-vd-ocr/TMP/1000`, `ulb-it-vd-ocr/TMP/1001`.. and so on.

For my workflows this is a rather breaking change since classically they were copied from any absolute local path into the workspace group folder. But if I know how to handle, it's fine as it is. 
There were some discussions about Release Management :smirk: 
Don't worry, everything will be alright :sunglasses: Feel free to close

[...]

PR #520, opened 2020-06-16
workspace add: os.path.join behavior led to faulty existence check

kba:
The problem was that `os.path.join` has a peculiar behavior when trying to concatenate absolute paths: It will just return the last one. Therefore the `exists` check only checked whether the source file exists - which it does, obviously - and never copied them to the workspace.

I added a test to reproduce #519 so this won't happen in the future.

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -154,15 +154,16 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, ignore, check_
 
     kwargs = {'fileGrp': file_grp, 'ID': file_id, 'mimetype': mimetype, 'pageId': page_id, 'force': force, 'ignore': ignore}
     log = getLogger('ocrd.cli.workspace.add')
+    log.debug("Adding '%s' (%s)", fname, kwargs)
     if not (fname.startswith('http://') or fname.startswith('https://')):
         if not fname.startswith(ctx.directory):
-            if exists(join(ctx.directory, fname)):
-                fname = join(ctx.directory, fname)
+            if exists(ctx.directory + '/' + fname):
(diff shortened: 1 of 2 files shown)


bertsky:
> The problem was that `os.path.join` has a peculiar behavior when trying to concatenate absolute paths: It will just return the last one.

Ouch!

ocrd/ocrd/cli/workspace.py
```diff
@@ -154,15 +154,16 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, ignore, check_
 
     kwargs = {'fileGrp': file_grp, 'ID': file_id, 'mimetype': mimetype, 'pageId': page_id, 'force': force, 'ignore': ignore}
     log = getLogger('ocrd.cli.workspace.add')
+    log.debug("Adding '%s' (%s)", fname, kwargs)
     if not (fname.startswith('http://') or fname.startswith('https://')):
         if not fname.startswith(ctx.directory):
-            if exists(join(ctx.directory, fname)):
-                fname = join(ctx.directory, fname)
+            if exists(ctx.directory + '/' + fname):
+                fname = ctx.directory + '/' + fname
```
bertsky, line 161: Is this really worth giving up the abstraction of `os.path.join` – what about support for other path name conventions (like Windows)?

One could specifically intercept the problematic case (absolute + absolute)…

kba: That was just a quick fix, I'll implement it with pathlib.

kba: I decided to just check whether `fname` is absolute. Once (if ever) we don't need to support 3.5 anymore, we can refactor a lot of os.path/shutil code with pathlib but until then we better avoid mixing the two.

PR #525, opened 2020-06-23
Image API: respect annotated order of features

### LINKED ISSUE #524: image API: respect order of derived image features
The functions `Workspace.image_from_page` and `image_from_segment` must crop according to `Page/Border/Coords` or `TextRegion/Coords`, and they must rotate according to `Page/@orientation` or `TextRegion/@orientation`.

But what if the segment has both?

1. crop-first: 
   - find the bbox of the border polygon, mask the parts outside of the polygon with background, and crop to that bbox
   - then rotate the cropped image, which increases the image size
2. rotate-first:
   - rotate, which increases the image size
   - find the bbox of the (rotated) border polygon in the rotated image, mask the parts outside of the polygon with background, and crop to that bbox

These 2 approaches create different images and different coordinates. So whatever gets used, must be used consistently. 

Now core is always responsible for the coordinate transformations, but not always responsible for producing derived images (as processors can create and annotate AlternativeImage themselves, which will have priority over core-generated ones if they meet all requested features). Thus, it has to _know_ which ordering of both operations has been applied when some `AlternativeImage` has both.

The only choice we have IMHO is making `AlternativeImage/@comments` order-sensitive. That is, `'binarized,cropped,deskewed'` is 1 and `'binarized,deskewed,cropped'` is 2.

When there is no AlternativeImage yet, the decision is up to core.

For the [implementation](https://github.com/OCR-D/core/blob/f708420d806ab6363917e3a53519d85505327568/ocrd/ocrd/workspace.py#L366-L535), this unfortunately means a large restructuring of both functions.

Or am I missing something?

bertsky:
Fix #524 

I am still testing, please don't merge yet.


kba:
😮 that was fast.

bertsky:
> that was fast.

too fast – the CI failure tells me I forgot rounding to integer…

bertsky:
There was a subtle mistake I did not notice earlier: the reference bbox dragged along with the transform as it is (re)construced must be initialized _before_ picking from AlternativeImages.

But the last version _seems_ to work fine. I haven't tested many different kinds of workflows yet, though. – This is another instance where not having dedicated unit tests for this complex and intricate functionality and also not having real-life workflow tests falls heavily on our toes.

ocrd/ocrd/workspace.py
```diff
@@ -461,64 +417,124 @@ def image_from_page(self, page, page_id,
...
+                                         not 'rotated-%d' % orientation in feature_filter.split(','))
+                                     else []) +
+                                    (['deskewed']
+                                     if (skew and
+                                         not 'deskewed' in page_coords['features'] and
+                                         not 'deskewed' in feature_filter.split(','))
+                                     else []) +
+                                    # not a feature to be added, but merely as a fallback position
+                                    # to always enter loop at i == len(alternative_image_features)
+                                    ['_check']):
```
kba, line 448: That's one massive for-in-clause but I get the idea, though not this last part, even with the comment. You're adding a *pseudo-feature* `"_check_"` - why? How does that pertain to the loop index being equal to the number of features given?

ocrd/ocrd/workspace.py
```diff
@@ -461,64 +417,124 @@ def image_from_page(self, page, page_id,
...
+            # image geometry vs feature consistency can only be checked
+            # after all features on the existing AlternativeImage have
+            # been adjusted for in the transform, and when there is a mismatch,
+            # additional steps applied here would only repeat the respective
+            # error message; so we only check once at the boundary between
+            # existing and new features
+            # FIXME we should check/enforce consistency when _adding_ AlternativeImage
+            if (i == len(alternative_image_features) and
+                not (page_xywh['w'] - 2 < page_image.width < page_xywh['w'] + 2 and
```
kba, line 457: A constant `PAGE_CROP_TOLERANCE` might make the intention clearer than `2`.

kba:
>  This is another instance where not having dedicated unit tests for this complex and intricate functionality and also not having real-life workflow tests falls heavily on our toes.

I agree that more unit tests for these complex functions and the many if-else-paths would inspire more confidence in the correctness of the tools. I can write such tests, make sure coverage is as complete as possible etc. We do however need [more test data with AlternativeImages](https://github.com/OCR-D/assets) for that. It doesn't have to be a complete workflow for the unit tests, just PAGE-XML with AlternativeImages with different combination of features in `@comment`.

ocrd/ocrd/workspace.py
bertsky: > How does that pertain to the loop index being equal to the number of features given?

…to the number of _already_ existing features.

Image size must be checked _after_ all existing features (from the AlternativeImage) have been reconstructed in the transform, but _before_ applying missing features (because operations in core are always consistent from there on, and I don't want to repeat the warning about the AlternativeImage multiple times).

But in case there are no missing features, I have to make sure to at least enter the loop body once more.

At this point, I know what you are going to say: why not just insert the pseudo-feature _before_ the added features and check for it directly instead of the position? – Consider it done!

ocrd/ocrd/workspace.py
bertsky: Indeed.

ocrd/ocrd/workspace.py
kba: thx, I get it now. Yes, making the image size `_check` explicit is a good idea though.

bertsky:
Wait, I still get occasional errors beyond 2px tolerance, investigating...

bertsky:
> Wait, I still get occasional errors beyond 2px tolerance, investigating...

That was actually a [subtle bug](https://github.com/OCR-D/ocrd_tesserocr/commit/726fede7a261b4f4de522d3328d54ec23f05d5a3) in the cropper.

This PR is fine AFAICT!

> I agree that more unit tests for these complex functions and the many if-else-paths would inspire more confidence in the correctness of the tools. I can write such tests, make sure coverage is as complete as possible etc.

That would be fantastic. I will help you of course.

> We do however need [more test data with AlternativeImages](https://github.com/OCR-D/assets) for that. It doesn't have to be a complete workflow for the unit tests, just PAGE-XML with AlternativeImages with different combination of features in `@comment`.

You started `kant_aufklaerung_1784-complex` – this is the right approach IMO. Maybe add processing results for another workflow (including cropping and segmentation), and add another workspace with more skew (like `weigel_gnothi02_1618`)...

bertsky:
Is there any further reason for this to wait? We could add the unit and workflow tests later, and the spec is hopelessly out of sync in this area anyway...

kba:
No reason to wait, I am eager to merge this, the parameter stuff and the utils addtitions we discussed in ocrd_cis as soon as possible, later today or over the weekend.

PR #530, opened 2020-07-10
Processor helpers

kba:
* assert_file_grp_cardinality as discussed in https://github.com/cisocrgroup/ocrd_cis/pull/57
* make_file_id as discussed in https://github.com/cisocrgroup/ocrd_cis/pull/57 with the ability to detect sequences of 3+ numbers in the ID and use that instead of `n` if available


ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -809,3 +811,26 @@ def parse_json_string_or_file(value='{}'):    # pylint: disable=unused-argument
...
+    """
+    Derive the file ID of an output file from an input file, the ``output_file_grp`` and a fallback counter ``n``.
+    """
+    ret = ocrd_file.ID.replace(ocrd_file.fileGrp, output_file_grp)
+    if ret == ocrd_file.ID:
+        m = re.match(r'.*?(\d{3,}).*', ret)
+        if m:
+            n = m.group(1)
+        ret = concat_padded(output_file_grp, n)
+    return ret
```
bertsky, line 836: I still don't think we should require the processor to determine `n`, esp. not from the current iterator position in `Processor.input_files`.

See reasoning [here](https://github.com/cisocrgroup/ocrd_cis/pull/57#discussion_r451450578).

```suggestion
def make_file_id(ocrd_file, output_file_grp):
    """
    Derive a new file ID for an output file from an existing input file ``ocrd_file``
    and the name of the output file's ``fileGrp/@USE``, ``output_file_grp``.
    If ``ocrd_file``'s ID contains the input file's fileGrp name, then replace it by ``output_file_grp``.
    Otherwise use ``output_file_grp`` together with the position of ``ocrd_file`` within the input fileGrp
    (as a fallback counter).
    """
    ret = ocrd_file.ID.replace(ocrd_file.fileGrp, output_file_grp)
    if ret == ocrd_file.ID:
        m = re.match(r'.*?(\d{3,}).*', ocrd_file.pageId or '')
        if m:
            n = m.group(1)
        else:
            files = ocrd_file.mets.find_files(fileGrp=ocrd_file.fileGrp, mimetype=ocrd_file.mimetype)
            n = files.index(ocrd_file)
        ret = concat_padded(output_file_grp, n)
    return ret
```

kba: I find `OcrdFile.mets` to be code smell (hard to test) but I admit your solution is elegant. I'll try to mock `OcrdFile.mets` in the unit test.

bertsky: > I find `OcrdFile.mets` to be code smell (hard to test)

I don't understand that – do you mean we should not assume `OcrdFile` and `OcrdMets` to be tightly coupled? Are there any contexts in which we (want to) create `OcrdFile` independent of its METS reference?

bertsky: > I'll try to mock `OcrdFile.mets` in the unit test.

Thx!

kba: The tight coupling always runs the danger that either the OcrdFile or the OcrdMets reference goes out of scope.  For example in `OcrdMets.find_files`, the OcrdFIles returned will be **new instances**, so checks like `list_of_files.index(ocrd_file)` are not reliable. As long as we are aware of that and program defensively, not relying on object references and rather on the actual data, it is not a problem.

I've adapted your solution with that in mind and added a loop to prevent duplicate IDs and created a METS-based unit test.

bertsky: > The tight coupling always runs the danger that either the OcrdFile or the OcrdMets reference goes out of scope. For example in `OcrdMets.find_files`, the OcrdFIles returned will be **new instances**, so checks like `list_of_files.index(ocrd_file)` are not reliable. As long as we are aware of that and program defensively, not relying on object references and rather on the actual data, it is not a problem.

Oh, right, I had just assumed that our `OcrdFile` would have an `__eq__` which is not more than the sum of its (non-etree) parts (ID, MIME type, pageId, filename, METS ref). But we don't define such a predicate (yet). And even then, the same thing (going out of scope and re-instantiating) could happen to the METS ref...

bertsky: > I've adapted your solution with that in mind and added a loop to prevent duplicate IDs and created a METS-based unit test.

Oh, yes, that's much better. There's no need to look at `OcrdFile` refs directly.

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -829,15 +829,21 @@ def make_file_id(ocrd_file, output_file_grp):
...
         if m:
             n = m.group(1)
         else:
-            files = ocrd_file.mets.find_files(fileGrp=ocrd_file.fileGrp, mimetype=ocrd_file.mimetype)
-            n = files.index(ocrd_file)
+            ids = [f.ID for f in ocrd_file.mets.find_files(fileGrp=ocrd_file.fileGrp, mimetype=ocrd_file.mimetype)]
+            try:
+                n = ids.index(ocrd_file.ID)
+            except ValueError:
+                n = len(ids)
```
bertsky, line 844: Much better!

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -829,15 +829,21 @@ def make_file_id(ocrd_file, output_file_grp):
...
         ret = concat_padded(output_file_grp, n)
+        while ocrd_file.mets.find_files(ID=ret):
+            n += 1
+            ret = concat_padded(output_file_grp, n)
```
bertsky, line 848: Ah, yes of course, very useful!

ocrd_utils/ocrd_utils/__init__.py
```diff
@@ -809,3 +811,26 @@ def parse_json_string_or_file(value='{}'):    # pylint: disable=unused-argument
...
+
+def assert_file_grp_cardinality(grps, n):
+    """
+    Assert that a string of comma-separated fileGrps contains exactly ``n`` entries.
+    """
+    if isinstance(grps, str):
+        grps = grps.split(',')
+    assert len(grps) == n, \
+            "Expected exactly %d output file group%s, but '%s' has %d" % (
+                n, '' if n == 1 else 's', grps, len(grps))
```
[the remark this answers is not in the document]
kba: Since I'm not a language person: Is "cardinality" correct, should it not be "arity"?

ocrd_utils/ocrd_utils/__init__.py
kba: >  I had just assumed that our OcrdFile would have an `__eq__`

No idea why it does not. I'll add it in a separate PR.

kba: https://github.com/OCR-D/core/pull/532

ocrd_utils/ocrd_utils/__init__.py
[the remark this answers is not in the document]
bertsky: > Since I'm not a language person: Is "cardinality" correct, should it not be "arity"?

I'd say _cardinality_ (of the set of file groups) is the general concept, _arity_ is the cardinality of some operator, so both terms apply in this case.

PR #532, opened 2020-07-12
Ocrd file eq

kba:
Implement `__eq__`, so different instances of equivalent OcrdFile can be compared with `==` and `!=`.

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_file.py
+++ b/ocrd_models/ocrd_models/ocrd_file.py
@@ -61,6 +61,13 @@ class OcrdFile():
         ])
         return '<OcrdFile ' + props + ']/> '
 
+    def __eq__(self, other):
+        print(self, other)
+        return self.ID == other.ID and \
+               self.url == other.url and \
+               self.mimetype == other.mimetype and \
+               self.fileGrp == other.fileGrp
+
     @property
     def basename(self):
         """
--- a/ocrd_models/ocrd_models/ocrd_file.py
+++ b/ocrd_models/ocrd_models/ocrd_file.py
@@ -3,7 +3,7 @@ API to ``mets:file``
 """
 from os.path import splitext, basename
 
-from ocrd_utils import is_local_filename, get_local_filename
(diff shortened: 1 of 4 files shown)


ocrd_utils/ocrd_utils/constants.py
```diff
@@ -38,6 +38,7 @@
 
 MIME_TO_EXT = {
     'image/tiff': '.tif',
+    'image/tif': '.tif',
```
bertsky, line 41: That's not a valid MIME type, right? You only want to be a bit tolerant, right?

ocrd_models/ocrd_models/ocrd_file.py
```diff
@@ -61,6 +61,13 @@ def __str__(self):
         ])
         return '<OcrdFile ' + props + ']/> '
 
+    def __eq__(self, other):
+        print(self, other)
```
bertsky, line 65:
```suggestion
```

ocrd_models/ocrd_models/ocrd_file.py
```diff
@@ -61,6 +61,13 @@ def __str__(self):
...
+        return self.ID == other.ID and \
+               self.url == other.url and \
+               EXT_TO_MIME[MIME_TO_EXT[self.mimetype]] == EXT_TO_MIME[MIME_TO_EXT[other.mimetype]] and \
+               self.fileGrp == other.fileGrp
```
bertsky, line 69: I wonder why we should go beyond `ID` here at all. METS file IDs are unique, and we cannot test for equality of the METS reference anyway.

ocrd_utils/ocrd_utils/constants.py
kba: Yes, b/c it is used (wrongly) like that sometimes.

ocrd_models/ocrd_models/ocrd_file.py
kba: You're probably right, I'll keep the other attribute checks commented out though, in case we need more complex `==` chcecks in the future.

PR #559, opened 2020-08-10
Processor resource discovery

kba:
Implementing https://github.com/OCR-D/spec/pull/163

* processors have a new method `resolve_resource` that gets passed a parameter value and tries to find it in one of the locations defined in OCR-D/spec#163
* core now has a builtin list (extensible via user list) of known resources for processors, e.g. the gt4histocr models for tesseract and calamari
* new CLI `ocrd resmgr` that allows browsing the builtin list, listing installed resources and downloading new resources
* resources that are listed in the builtin list can be referenced by URL or name


ocrd/ocrd/resource_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        List installed resources, matching with registry by ``name``
+        """
+        ret = []
+        for executable in [executable] if executable else self.database.keys():
+            reslist = []
+            for res_filename in list_all_resources(executable):
+                res_name = Path(res_filename).name
+                resdict = [x for x in self.database[executable] if x['name'] == res_name]
+                if not resdict:
+                    # TODO handle gracefully
```
bertsky, line 64: I'm confused: What exact circumstances would lead us into this branch? 

My test already fails earlier: After…

    ocrd resmgr download -n ocrd-cis-ocropy-recognize https://github.com/zuphilip/ocropy-models/raw/master/en-default.pyrnn.gz

…I still get _no_ results for…

    ocrd resmgr list-installed

…(because the unregistered download does not touch the user resources – see other comment – and the outer loop only looks for registered ones), but…

    ocrd resmgr list-installed -e ocrd-cis-ocropy-recognize

…already fails with…

```
  File "venv/lib/python3.6/site-packages/ocrd/cli/resmgr.py", line 50, in list_installed
    for executable, reslist in resmgr.list_installed(executable):
  File "venv/lib/python3.6/site-packages/ocrd/resource_manager.py", line 62, in list_installed
    resdict = [x for x in self.database[executable] if x['name'] == res_name]
KeyError: 'ocrd-cis-ocropy-recognize'
```

So it would seem that you have to enter either via a loaded and validated resource database (in which case IIUC there will always be an entry for any of its resources), or just by name (in which case IIUC no database lookup is warranted anyway)...

ocrd/ocrd/processor/base.py
```diff
@@ -147,33 +147,34 @@ def add_metadata(self, pcgts):
...
-                             parameter_name)
-        if val.startswith('http:') or val.startswith('https:'):
-            return OcrdResourceManager().download(executable, val)
-        ret = next([cand for cand in list_resource_candidates(executable, val) if exists(cand)])
+        if exists(val):
+            return val
+        ret = [cand for cand in list_resource_candidates(executable, val) if exists(cand)]
         if ret:
-            return ret
-        bundled_fpath = resource_filename(__name__, val)
```
bertsky: Why did you remove the `pkg_resources.resource_filename` (Python fall-back) mechanism entirely? Isn't this even required by the spec now?

(I know it's harder to get these under the centralized `ocrd resmgr list-installed`. But if `Processor.list_all_resources()` would append the files from pkg_resources, then the former could delegate to these via CLI calls.)

ocrd/ocrd/processor/base.py
```diff
@@ -107,6 +147,41 @@ def add_metadata(self, pcgts):
...
+            name=resdict['name'],
+            path_in_archive=resdict['path_in_archive'],
+            resource_type=resdict['type']
+        ))
+
+    def list_all_resources(self):
+        """
+        List all resources found in the filesystem
+        """
+        return list_all_resources(self.ocrd_tool['executable'])
```
bertsky, line 183: Plus append by anything in `pkg_resources.resource_filename` (see comment on `resolve_resource`...)?

ocrd_models/ocrd_page_user_methods.py
```diff
@@ -113,6 +113,7 @@ def _add_method(class_re, method_name, file_name=None):
     _add_method(r'^(BorderType|RegionType|TextLineType|WordType|GlyphType)$', 'set_Coords'),
     _add_method(r'^(PageType)$', 'set_Border'),
     _add_method(r'^(CoordsType)$', 'set_points'),
+    _add_method(r'^(PageType)$', 'get_AllTextLine'),
```
bertsky, line 116: By analogy to our `get_AllRegions`, IMO this should also be plural here:

```suggestion
    _add_method(r'^(PageType)$', 'get_AllTextLines'),
```

ocrd_models/ocrd_page_user_methods/get_AllTextLine.py
```diff
@@ -0,0 +1,9 @@
+def get_AllTextLine(self):
+    """
+    Return all the TextLine in the document
+    """
+    ret = []
+    for reg in self.get_AllRegions(['Text']):
+        ret += reg.get_TextLine()
+    return ret
+
```
bertsky, line 9: Besides the plural, I recommend also offering an `order` kwarg here, delegating to `get_AllRegions` at the region level, and additionally respecting `textLineOrder` at the line level.

bertsky (commented):
Very impressive!

Not sure I understand all ramifications yet. What would need to be adapted
1. processors code
2. model deployment in ocrd_all and modules
3. documentation (workflows, usage, installation)

For 1 and 2, it would help to see an example, say for ocrd_tesserocr.

For 2, there's also the question of how to incorporate Python package resources and git submodules.

Your beta pre-release is an excellent vehicle for realistic tests in that direction.

But let's first sort out some details (in the comments)...

kba:
> Not sure I understand all ramifications yet. What would need to be adapted
> 
> 1. processors code

Basically, replace `self.parameter['my-file-parameter']` with `self.resolve_resource(self.parameter['my-file-parameter'])`. 

> 2. model deployment in ocrd_all and modules

`make install-models*` should be adapted to store models in the search path (e.g. use `venv/share/ocrd-calamari-recognize` instead of `venv/share/ocrd_calamari/`. 

> 3. documentation (workflows, usage, installation)

The models documentation would need a rewrite with examples of `ocrd resmgr` and an explanation of the search path logic. Also, we should decide on a convention for docker, e.g. mounting a host directory of models as `$HOME/.cache`.

> 
> For 1 and 2, it would help to see an example, say for ocrd_tesserocr.
> 

Here's the change for ocrd_calamari: https://github.com/OCR-D/ocrd_calamari/pull/53/files, here for ocrd_anybaseocr https://github.com/OCR-D/ocrd_anybaseocr/pull/78/files


> For 2, there's also the question of how to incorporate Python package resources and git submodules.

I'm not sure I understand - can you elaborate pls?

ocrd/ocrd/resource_manager.py
```diff
@@ -0,0 +1,142 @@
...
+
+from .constants import RESOURCE_LIST_FILENAME
+
+from ocrd_validators import OcrdResourceListValidator
+from ocrd_utils import getLogger
+from ocrd_utils.constants import HOME, XDG_CACHE_HOME
+from ocrd_utils.os import list_resource_candidates, list_all_resources, pushd_popd
+
+builtin_list_filename = Path(RESOURCE_LIST_FILENAME)
+user_list_filename = Path(HOME, 'ocrd', 'resources.yml')
```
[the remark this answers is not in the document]
kba: Agreed, `$HOME/ocrd/resources.yml` was a typo anyway.

ocrd/ocrd/cli/resmgr.py
```diff
@@ -0,0 +1,114 @@
...
+    if not reslist:
+        log.info("No resources found in registry")
+        if is_url and any_url:
+            log.info("Downloading unregistered resource %s" % url_or_name)
+            with requests.head(url_or_name) as r:
+                content_length = int(r.headers.get('content-length'))
+            with click.progressbar(length=content_length) as bar:
+                fpath = resmgr.download(
+                    executable,
+                    url_or_name,
```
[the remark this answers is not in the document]
kba: > IIUC this means the new / unregistered resources will not be visible under list-installed afterwards.

They will, but currently there's just question marks where the description and URL would be for a registered resource:

```
$ touch $HOME/.cache/ocrd-tesserocr-recognize
$ ocrd resmgr list-installed
ocrd-tesserocr-recognize
- bar (???)
  ???
- Fraktur_GT4HistOCR.traineddata (https://ub-backup.bib.uni-mannheim.de/~stweil/ocrd-train/data/Fraktur_5000000/tessdata_fast/Fraktur_50000000.334_450937.traineddata)                       
  Tesseract LSTM model trained on GT4HistOCR
```

You mean, create an entry with URL and name in `$HOME/.config/resources.yml` and create that if it doesn't exist yet? Yes that would work.

> ```ocrd resmgr download -n ocrd-cis-ocropy-recognize https://github.com/zuphilip/ocropy-models/raw/master/en-default.pyrnn.gz```

My reasoning was: If you use a URL once, you will use this URL in the future (and not the short name), hence the mapping from URL to name should be simple to make this reproducible. But I guess "last URL segment sans fragment identifier and query" would work just as well.

ocrd/ocrd/cli/resmgr.py
```diff
@@ -0,0 +1,114 @@
...
+            XDG_CONFIG_HOME if location == 'config' else \
+            getcwd()
+    is_url = url_or_name.startswith('https://') or url_or_name.startswith('http://')
+    find_kwargs = {'executable': executable}
+    if url_or_name != '*':
+        find_kwargs['url' if is_url else 'name'] = url_or_name
+    reslist = resmgr.find_resources(**find_kwargs)
+    if not reslist:
+        log.info("No resources found in registry")
+        if is_url and any_url:
```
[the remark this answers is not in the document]
kba: > For example, I might have mounted my models via NFS or downloaded them via some other means already

You mean, you want to allow absolute file paths in addition to HTTP(S) URL? Sure, that would be doable.

bertsky:
> > For 2, there's also the question of how to incorporate Python package resources and git submodules.
> 
> I'm not sure I understand - can you elaborate pls?

Modules might want to distribute config files or models via Python package resources. This should be standardized for OCR-D and supported by the resmgr / Processor reference implementation.

The git submodules part refers to how can we generalize ocrd_all's `install-models` to conveniently cover other modules, with models either under some makefile target or in a data submodule.

ocrd/ocrd/resource_manager.py
kba: > …I still get no results for…

The problem is: what do we iterate? Do we iterate over all the items in all the XDG_FOLDERS that follow our convention (`ocrd-.*`) or over the keys in the `resource_list.yml`? I chose the latter for simplicity's sake but I understand how that can be confusing.

The branch is triggered if there is a resource for one of the processors for which we know there are resources (has a key in `resource_list.yml`) but not the resource with this particular name.

I'll change the outer loop to actually inspect the file system.

ocrd/ocrd/cli/resmgr.py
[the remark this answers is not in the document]
bertsky: > You mean, create an entry with URL and name in `$HOME/.config/resources.yml` and create that if it doesn't exist yet? Yes that would work.

Yes, exactly that.

> > `ocrd resmgr download -n ocrd-cis-ocropy-recognize https://github.com/zuphilip/ocropy-models/raw/master/en-default.pyrnn.gz`
> 
> My reasoning was: If you use a URL once, you will use this URL in the future (and not the short name), hence the mapping from URL to name should be simple to make this reproducible. But I guess "last URL segment sans fragment identifier and query" would work just as well.

You mean the user will pass the URL in the parameter JSON? But then why does it say 
> Use in parameters as 'httpsgithubcomzuphilipocropymodelsrawmasterendefaultpyrnngz'

? I would expect either 
> Use in parameters as 'https://github.com/zuphilip/ocropy-models/raw/master/en-default.pyrnn.gz'

or (better)
> Use in parameters as 'en-default.pyrnn.gz'

ocrd/ocrd/processor/base.py
kba: > Why did you remove the pkg_resources.resource_filename (Python fall-back) mechanism again entirely? Isn't this even required by the spec now?

To be honest, I was trying to get this right for ocrd_typegroups_classifier but I found it more intuitive to offer the bundled resource as a regular resource in `resources_list.yml` (since it is already in GitHub and hence HTTP-downloadable) and let ocrd_typegroups_classifier handle the case where no nerwork parameter was given itself. It's a bit of a waste of space but makes the code for `list_all_resources` and `resmgr` simpler.

It would be good to have an example project that makes use of bundled resources to test this with and if we don't, it would be better to change the spec TBH. 

BTW, I wanted to also list https://github.com/ASVLeipzig/cor-asv-ann-models but gave up on getting a "lightweight" solution working to download from LFS.

ocrd_models/ocrd_page_user_methods.py
kba: yep, it's also `get_AllAlternativeImages` now, I'll fix this.

ocrd/ocrd/cli/resmgr.py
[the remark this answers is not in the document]
bertsky: Yes, absolute or CWD-relative paths – it does not matter, because you would have to copy (or link) them to a standard location (where `list_all_resources` would expect them) anyway.

ocrd/ocrd/cli/resmgr.py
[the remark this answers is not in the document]
kba: > `Use in parameters as 'httpsgithubcomzuphilipocropymodelsrawmasterendefaultpyrnngz'`

Excellent point, my reasoning was flawed :)

ocrd/ocrd/processor/base.py
bertsky: > > Why did you remove the pkg_resources.resource_filename (Python fall-back) mechanism again entirely? Isn't this even required by the spec now?
> 
> To be honest, I was trying to get this right for ocrd_typegroups_classifier but I found it more intuitive to offer the bundled resource as a regular resource in `resources_list.yml` (since it is already in GitHub and hence HTTP-downloadable) and let ocrd_typegroups_classifier handle the case where no nerwork parameter was given itself. It's a bit of a waste of space but makes the code for `list_all_resources` and `resmgr` simpler.

I see. Well, `ocrd_typegroups_classifier` or [document-image-binarization](https://github.com/ajgallego/document-image-binarization) are examples of packages which distribute their model files (not sure about binary vs sdist installation though), so it _should_ already be available locally. But I agree that the waste of bandwidth/space is probably not a big deal in general because if the models are too large, devs will not include them in the package (or even the same repo) anyway.

For modules with very large files like `ocrd_anybaseocr` – not all of which might be needed – I guess it makes no sense to expect them to be distributed along with the base pkg. (Although the PyPI server could still help distribute dependent packages like `ocrd-anybaseocr-models-dewarping`, `ocrd-anybaseocr-models-segmentation` etc.)

But then there's modules like ocrd_wrap: I want to put many small preset files there in such a way that they can be referenced on the CLI  (`-p param_im6convert-denoise-raw.json -P level-of-operation region`). But if that's (going to be) possible by just doing `ocrd resmgr download ocrd-preprocess-image '*'` (which I could even delegate to in my `make install` rule), then fine. (I might even register resources for external tools there.)

> It would be good to have an example project that makes use of bundled resources to test this with and if we don't, it would be better to change the spec TBH.

Yes, agreed: If we don't actually need package resources as a mechanism, we shouldn't have it in the spec. Wrapping an existing pkg resource to an OCR-D file resource spec should be easy.

But that's actually something I still don't see here: we have a _centralized_ `resource_list.yml` now, but not a way for modules to feed their contributions...

> BTW, I wanted to also list https://github.com/ASVLeipzig/cor-asv-ann-models but gave up on getting a "lightweight" solution working to download from LFS.

I'll have a look myself later-on. Github LFS always gave us problems in the past already (we usually exceed the monthly quota without even knowing why). There's still the Gitlab backup, but...

ocrd_models/ocrd_page_user_methods/get_AllTextLine.py
kba: 4def1d97, implemented a `region_order` kwarg delegating to `order` in `get_AllRegions`. As for the `textLineOrder`, I've asked @tboenig to clarify what this means and how we can test it.

ocrd_models/ocrd_page_user_methods/get_AllTextLines.py
```diff
@@ -0,0 +1,19 @@
...
+    Arguments:
+        region_order ("document"|"reading-order"|"reading-order-only") Whether to
+            return regions sorted by document order (``document``, default) or by
+            reading order with regions not in the reading order at the end of the
+            returned list (``reading-order``) or regions not in the reading order
+            omitted (``reading-order-only``)
+        textline_order ("top-to-bottom"|"bottom-to-top"|left-to-right"|"right-to-left")
+            The order of text lines within a block (not currently used)
+    """
+    # TODO handle textLineOrder
```
bertsky, line 14: I think the mnemonics should be different, `textline_order` should be a boolean.

If true, then do something like:
```python
    for reg in self.get_AllRegions(['Text'], order=region_order):
        lo = reg.get_textLineOrder() or self.get_textLineOrder() or 'top-to-bottom'
        lines = reg.get_TextLine()
        ret += lines if lo in ['top-to-bottom', 'left-to-right'] else list(reversed(lines))
    return ret
```

Not entirely sure about the PAGE-XML semantics here TBH. Must XML ordering always be top-to-bottom (or left-to-right in case of vertical text)? In that case, you would need to _apply_ `@textLineOrder` to get the correct text concatenation (`right-to-left` for classical Chinese). But the attribute could also be interpreted to _describe_ just how the current XML ordering looks like (which you could verify by its coordinates). In that case, you would never re-order anything.

ocrd_models/ocrd_page_user_methods/get_AllTextLine.py
bertsky: > As for the `textLineOrder`, I've asked @tboenig to clarify what this means and how we can test it.

Ah, great – I was just going to ask you for that, cf. [this comment](https://github.com/OCR-D/core/pull/559/files/4bf12fb7eaa39fc2a73874df3b69c4b6b24b5965..4def1d974e9645753db924b5ca0c5ed1e7682ed5#r551472862)

ocrd/ocrd/resource_manager.py
kba: 79116038

ocrd/ocrd/cli/resmgr.py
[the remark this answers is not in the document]
kba: BTW, https://github.com/zuphilip/ocropy-models/raw/master/en-default.pyrnn.gz is already in our resource_list.yml, I just had a typo in it, fixed in 54a214a0

ocrd/ocrd/resource_manager.py
```diff
@@ -55,16 +56,27 @@ def list_installed(self, executable=None):
...
+            all_executables = [executable]
+        else:
+            # resources we know about
+            all_executables = list(self.database.keys())
+            # resources in the file system
+            parent_dirs = [XDG_CACHE_HOME, XDG_CONFIG_HOME, XDG_DATA_HOME]
+            if 'VIRTUAL_ENV' in environ:
+                parent_dirs += [join(environ['VIRTUAL_ENV'], 'share')]
+            for parent_dir in parent_dirs:
+                all_executables += [x for x in listdir(parent_dir) if x.startswith('ocrd-')]
```
bertsky, line 69: This does cover all circumstances for unregistered resources I could think of, but it also includes false positives now:
```
ocrd-im6convert
- ocrd-tool.json (???)
  ???
```

It seems that `venv/share/ocrd-im6convert/ocrd-tool.json` matched the `startswith` pattern here. So require `os.path.isdir(x)`, too?

kba: Yeah, I saw that too. But `isdir(x)` wouldn't help because `ocrd-im6convert` is a directory. I think I'll just make a new im6convert release with `SHAREDIR=$VIRTUAL_ENV/share/ocrd_im6convert` (note the `_` instead of `-`).

ocrd/ocrd/resource_manager.py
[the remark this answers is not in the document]
kba: e33346b4 New behavior: Look up resources with the builtin list and user list at `XDG_CONFIG_HOME/ocrd/resources.yml`. If an unknown resource is encountered for the first time, creates a stub entry in the user list.

ocrd/ocrd/resource_manager.py
bertsky: Yes, that makes sense. (ocrd_olena, ocrd_pagetopdf and ocrd_fileformat all use the underscore already.)

ocrd/ocrd/cli/resmgr.py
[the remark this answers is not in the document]
bertsky: Maybe (besides `nth_url_segment`) even normalize URL encoding (eg. replace `%C5%BF` by `ſ`)?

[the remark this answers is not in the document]
kba: 3ee66ce5 Now:

```
ocrd resmgr download -n ocrd-foo https://github.com/zuphilip/ocropy-french-models/raw/master/models/french-balzac-model-00105000.pyrnn.gz                                                 
15:13:11.315 INFO ocrd.cli.resmgr - No resources found in registry
15:13:11.315 INFO ocrd.cli.resmgr - Downloading unregistered resource https://github.com/zuphilip/ocropy-french-models/raw/master/models/french-balzac-model-00105000.pyrnn.gz               
  [####################################]  100%
15:13:16.579 INFO ocrd.cli.resmgr - Downloaded https://github.com/zuphilip/ocropy-french-models/raw/master/models/french-balzac-model-00105000.pyrnn.gz to /home/kba/.cache/ocrd-foo/french-ba
lzac-model-00105000.pyrnn.gz
15:13:16.579 INFO ocrd.cli.resmgr - Use in parameters as 'french-balzac-model-00105000.pyrnn.gz'

```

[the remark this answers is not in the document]
kba: > Maybe (besides nth_url_segment) even normalize URL encoding (eg. replace %C5%BF by ſ), perhaps even dropping all query/fragment components from the path?

Good point b21d4628 Do you have an example URL to test with?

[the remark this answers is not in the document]
bertsky: I do – and it _fails_ because the HEAD request does not get answered with a `content-length` header in Gitlab:
- my [s-ſ reconstruction model](https://git.informatik.uni-leipzig.de/ocr-d/cor-asv-ann-models/-/raw/master/s2s.gt4histocr.s-%C5%BF.d2.w0512.adam.attention.stateless.variational-dropout.char.transfer-lm.h5?inline=false)

```
  File "venv/lib/python3.6/site-packages/ocrd/cli/resmgr.py", line 85, in download
    content_length = int(r.headers.get('content-length'))
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

See [here](https://stackoverflow.com/a/44299915/14474237) for a solution (i.e. use GET but with `stream=True`)...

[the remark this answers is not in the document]
kba: Thanks, I will test with that and replace HEAD with bodyless GET. And good to know these resources are mirrorred on your GitLab, I was trying to get support for Git LFS but it is surprisingly tricky to implement without additional dependencies or lots of copy/paste code.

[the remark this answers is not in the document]
kba: d8d97af1 

```
13:38:08.997 INFO ocrd.cli.resmgr - Downloaded https://git.informatik.uni-leipzig.de/ocr-d/cor-asv-ann-models/-/raw/master/s2s.gt4histocr.s-%C5%BF.d2.w0512.adam.attention.stateless.variation
al-dropout.char.transfer-lm.h5?inline=false to /home/kba/.cache/ocrd-foo/s2s.gt4histocr.s-ſ.d2.w0512.adam.attention.stateless.variational-dropout.char.transfer-lm.h5                         
13:38:08.998 INFO ocrd.cli.resmgr - Use in parameters as 's2s.gt4histocr.s-ſ.d2.w0512.adam.attention.stateless.variational-dropout.char.transfer-lm.h5'
```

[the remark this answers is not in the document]
bertsky: Fantastic!

ocrd_models/ocrd_page_user_methods/get_AllTextLine.py
bertsky: cf. https://github.com/PRImA-Research-Lab/PAGE-XML/issues/26

ocrd/ocrd/cli/resmgr.py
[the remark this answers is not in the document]
kba: 509200c8

ocrd/ocrd/processor/base.py
kba: > But that's actually something I still don't see here: we have a centralized resource_list.yml now, but not a way for modules to feed their contributions...

There is the option of a user-defined list in addition to the centralized list. Another option would be to extend the ocrd-tool.json schema so that project authors can themselve define the resources and we introspect that at runtime?

bertsky: > Another option would be to extend the ocrd-tool.json schema so that project authors can themselve define the resources and we introspect that at runtime?

Excellent idea! So ocrd_all installation rules could aggregate ocrd's `resource_list.yml` or `~/.config/ocrd/resource.yml` from that (enabling `make install-models` after `make all`). But also, each individual module makefile could get a local `install-models` rule...

ocrd/ocrd/processor/base.py
bertsky: > Plus append by anything in `pkg_resources.resource_filename` (see comment on `resolve_resource`...)?

We don't want that anymore (https://github.com/OCR-D/spec/commit/fa6e00a269713d8593e77f9b2a3d8f19560e7649)

ocrd/ocrd/processor/base.py
kba: Since integrating this into ocrd-tool.json requires changes to the spec and we should probably adapt the metadata scheme to be interoperable with [@mittagessen's schema](https://github.com/mittagessen/kraken/blob/master/kraken/metadata.schema.json) I will do that in a separate set of PRs.

ocrd_models/ocrd_page_user_methods/get_AllTextLines.py
kba: Thanks, I'll implement it like that for now until we agreed on the right interpretation in https://github.com/PRImA-Research-Lab/PAGE-XML/issues/26

ocrd/ocrd/processor/base.py
bertsky: So IIUC you want to merge the centralized resource list first, and adapt ocrd_all, and then work towards decentralized, aggregated resource definitions in tool.json?

But the Kraken model schema does not even have a URL attribute! Or are you planning for some schema extension? And isn't it specific to OCR models (`script`, `graphemes`)? Also, `accuracy` by itself does not make sense to me. At least, some of these features should not be `required`...

bertsky: accuracy: Yes, the method / metric often is a parameter itself, but more importantly, different models will usually also have different test data...

Kraken, Okralact, Calamari config, ... so many options. I don't think we should even try to use some interoperable / unified scheme.

ocrd_models/ocrd_page_user_methods/get_AllTextLines.py
kba: 012e49ea The kwarg is now `respect_textline_order` and the algorithm like you proposed.

bertsky (commented):
If I could make another wish: It would be useful if `list-installed` would show the location type registered ressources were found in (`data`, `config`, `cache`, `virtualenv`), and likewise for unregistered resources (instead of the full path after `Found at`).

kba:
> If I could make another wish: It would be useful if list-installed would show the location type registered ressources were found in (data, config, cache, virtualenv), and likewise for unregistered resources (instead of the full path after Found at).

Sure, that makes sense.

kba:
> If I could make another wish: It would be useful if `list-installed` would show the location type registered ressources were found in (`data`, `config`, `cache`, `virtualenv`), and likewise for unregistered resources (instead of the full path after `Found at`).

a3cff9e8

```
$ ocrd resmgr download -n ocrd-tesserocr-recognize https://github.com/tesseract-ocr/tessdata/raw/master/fin.traineddata

$ cat $HOME/.config/ocrd/resources.yml
cat ~/.config/ocrd/resources.yml 
# OCR-D private resource list (consider sending a PR with your own resources to OCR-D/core)
ocrd-tesserocr-recognize:
- description: Found at virtualenv on 2021-01-21 19:18:30.631387
  name: fin.traineddata
  size: 21140513
  url: https://github.com/tesseract-ocr/tessdata/raw/master/fin.traineddata
  version_range: ???

$ ocrd resmgr list-installed -e ocrd-tesserocr-recognize                                                               
ocrd-tesserocr-recognize
- fin.traineddata @ virtualenv (https://github.com/tesseract-ocr/tessdata/raw/master/fin.traineddata)
  Found at virtualenv on 2021-01-21 19:18:30.63138
```

kba:
OK, so adapted to what @bersky and I discussed and up-to-date with https://github.com/OCR-D/spec/pull/178 🤞

ocrd/ocrd/resource_manager.py
```diff
@@ -71,9 +70,7 @@ def list_installed(self, executable=None):
             # resources we know about
             all_executables = list(self.database.keys())
             # resources in the file system
-            parent_dirs = [join(x, 'ocrd-resources') for x in [XDG_CACHE_HOME, XDG_CONFIG_HOME, XDG_DATA_HOME]]
-            if VIRTUAL_ENV:
-                parent_dirs += [join(VIRTUAL_ENV, 'share', 'ocrd-resources')]
+            parent_dirs = [join(x, 'ocrd-resources') for x in [XDG_DATA_HOME, '/usr/local/share']]
```
bertsky, line 73: What about the CWD? I know it's probably not realistic, because at processor runtime we are in the workspace, whereas during `ocrd resmgr` we might not, but we could still try :-)

```suggestion
            parent_dirs = [join(x, 'ocrd-resources') for x in [XDG_DATA_HOME, '/usr/local/share', getcwd()]]
```

bertsky (commented):
Maybe we could also import and expose `OcrdResourceManager` directly from `ocrd`, so use-cases like ocrd_tesserocr are more accessible.

ocrd/ocrd/resource_manager.py
```diff
@@ -0,0 +1,221 @@
...
+        for executable in [executable] if executable else database.keys():
+            for resdict in database[executable]:
+                if not name and not url:
+                    ret.append((executable, resdict))
+                elif url and url == resdict['url']:
+                    ret.append((executable, resdict))
+                elif name and name == resdict['name']:
+                    ret.append((executable, resdict))
+        return ret
+
```
bertsky, line 133: Also, it might still be helpful to encapsulate the default path here (think ocrd_tesserocr again):

```suggestion

    @property
    def default_resource_dir(self):
        return self.location_to_resource_dir('data')
```

bertsky (changes requested):

ocrd/ocrd/resource_manager.py
kba: Originally omitted because we didn't have the `ocrd-resources` directory and listing everything in PWD seemed overkill. But there's no reason now not include it.

kba:
> Maybe we could also import and expose OcrdResourceManager directly from ocrd, so use-cases like ocrd_tesserocr are more accessible.

a5858ec4

I also added a flag `--allow-uninstalled` to control whether to allow downloading resources for processors that are not installed. Without it `ocrd resmgr ocrd-foo /vmlinuz` will now cause an error (unless you have `ocrd-foo` in your `$PATH`).

Also added `ocrd resmgr download '*`` which will download all resources for all installed processors (or all resources in the registry if `--allow-uninstalled`).

I really hope we've got it right this time but I will test some more before merging just to be sure.

ocrd/ocrd/cli/resmgr.py
```diff
@@ -2,8 +2,9 @@
 from os import getcwd
 from os.path import join
 from pathlib import Path
-import requests
+from distutils.spawn import find_executable as which
```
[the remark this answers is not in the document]
kba: > but it also means to rule out processors/resources as installed if they are behind a sub-venv again. 

How do you mean? If the wrapper scripts are in `$PATH`, then why can we not assume that they are installed?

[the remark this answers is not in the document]
bertsky: I'm sorry – you can scratch that entire comment. (I forgot this already _is_ search in `PATH`, not in Python packages. Also, I already used that exact pattern in the workflow server...)

bertsky (approved):

PR #560, opened 2020-08-11
workspace.add_file: ensure processors consciously pass pageId (or not)

kba:
This change forces processors to provide an explicit `pageId` kwarg when using `workspace.add_file`. Non-page-specific data can still be added with `pageId=None`.

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -205,6 +205,8 @@ class Workspace():
             file_grp,
             kwargs.get('local_filename'),
             content is not None)
+        if 'pageId' not in kwargs:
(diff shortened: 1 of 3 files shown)


ocrd/ocrd/workspace.py
```diff
@@ -205,6 +205,8 @@ def add_file(self, file_grp, content=None, **kwargs):
             file_grp,
             kwargs.get('local_filename'),
             content is not None)
+        if 'pageId' not in kwargs:
+            raise ValueError("workspace.add_file must be passed a 'pageId' kwarg, even if it is None.")
```
bertsky, line 209: Yes, that's an elegant solution. There is no in between page-associated or global annotations, and processors must not be allowed to forget deciding which is right.

bertsky (approved):

PR #390, opened 2020-01-07
:bug: PIL Image.open was never closed in workspace, #377

kba:

PR #564, opened 2020-08-13
workspace: ensure PIL.Image.open FD gets closed

bertsky:
This can be considered a follow-up to #390. We missed that instance of `PIL.Image.open` usage without closing the FD. Since in this function we must return the `Image` object, we cannot use a context manager, but have to just give up the FD itself. See [here](https://pillow.readthedocs.io/en/stable/reference/open_files.html#file-handling)

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -278,6 +278,7 @@ class Workspace():
 
         with pushd_popd(self.directory):
             pil_image = Image.open(image_filename)
+            pil_image.load() # alloc and give up the FD
 
         if coords is None:
             return pil_image


bertsky:
CI failure: after https://github.com/OCR-D/ocrd_tesserocr/issues/137 is fixed (which [also affects core](https://github.com/OCR-D/ocrd_tesserocr/pull/138#issuecomment-673479328)), we have to merge master to move on here

ocrd/ocrd/workspace.py
```diff
@@ -278,6 +278,7 @@ def _resolve_image_as_pil(self, image_url, coords=None):
 
         with pushd_popd(self.directory):
             pil_image = Image.open(image_filename)
+            pil_image.load() # alloc and give up the FD
```
kba, line 281: Good that you added a comment, it's not obvious [that `Image.load` closes the FD](https://pillow.readthedocs.io/en/stable/reference/open_files.html#file-handling)

PR #566, opened 2020-08-15
Refactor "ocrd process" to be interoperable with OCR-D/spec#171

kba:
* ValidationReport -> ocrd_models
* Decouple parsing and validation of tasks into OcrdWf (workflow), OcrdWfStep and OcrdWfValidator


kba:
This is now very obsolete, we decided to use nextflow for productive workflows. `ocrd process` as it-is is enough for simple workflows, so it is beyond time to close this.

bertsky:
Agreed. Can we get something similar in core based on OtoN, though?

PR #575, opened 2020-08-24
circle ci: add python 3.8 test, upgrade minor releases

kba:


kba:
That went without a hitch. So let's support Python 3.8 moving forward.

bertsky:
> That went without a hitch. So let's support Python 3.8 moving forward.

Despite the fact Tensorflow will not support older releases (esp. 1.15) on that Python version?

kba:
> > That went without a hitch. So let's support Python 3.8 moving forward.
> 
> Despite the fact Tensorflow will not support older releases (esp. 1.15) on that Python version?

Yes, while we cannot support 3.8 in ocrd_all at the moment, it is still important to be future-proof.

PR #576, opened 2020-08-24
Generateds gdscollector

kba:
Here's the initial work on extending generateDS' GdsCollector to offer more helpful help messages.

This replaces the buildin `parse/parseString` methods with variants that know about the filename (and passes it on to OcrdGdsCollector.

OcrdGdsCollector should probably have the same interface, i.e. inherit from ValidationReport (which needs to support an alias `add_message` -> `add_error`).

Now that the foundations are here, we can discuss what kind of information should be captured and how messages should be formatted.


bertsky:
> Now that the foundations are here, we can discuss what kind of information should be captured and how messages should be formatted.

We discussed this briefly in yesterday's tech call: now that we have a `ValidationReport` along with the PAGE object, which makes validation errors _actionable_ (without the need to repeat validation in dedicated tools/steps), we could add _hooks_ to `to_xml` for post-processing (e.g. coordinate repair). As a first step, this could be setting negative `points` to zero automatically.

Issue #322, opened 2019-09-23
make writing to METS exclusive for parallelization


bertsky:
In #278 we already had the case were the METS got broken due to ctrl+C in a single process. We discussed about parallelization but agreed to defer that topic. Now this is about providing a locking mechanism that allows parallel processing.

Currently it is impossible to run in parallel over a workspace, simply because it is too dangerous: the result of one actor could be lost due to another without any indication of it. This hampers [makefilization](https://gist.github.com/bertsky/18f836c941698832a60f6a3dd4cbf0e4) and similar efforts.

IIUC, getting FS locking in a [portable way](https://stackoverflow.com/questions/489861/locking-a-file-in-python) is tricky, if you want to get it right.

Does anyone have experience or solutions?

wrznr:
No experience let alone solutions. Sry.

kba:
Portability isn't of the essence though because I don't foresee the need to support non-POSIX filesystems.

No experience doing this with Python but the principle is simple and https://github.com/dmfrey/FileLock/blob/master/filelock/filelock.py (the accepted answer you linked to) looks like a reasonable solution.

bertsky:
I'd like to bring in another idea: Maybe file locking is the wrong way to think of page-level parallelization altogether. We have dependencies between successive processors in a workflow, so it does not make much sense to think of running these in parallel (without going into elaborate synchronization mechanisms like semaphores for all pages), except for a very narrow set of circumstances where workflows contain parallel branches. On the other hand, what really can be run independently (at least for most processors and workflows) are the pages _within_ one processing step. Even now, a processor implementation can do that, e.g. by using a `multiprocessing.Pool.map()` on the `workspace.input_files`.

And there's the rub really: In the current design, `Processor.process()` implementations must iterate over input files _themselves_. The responsibility for parallelization therefore is in the module provider (instead of the base class in core). 

So for 3.0, what we should aim at is the following:

1. deprecate `Processor.process()`
2. create `Processor.process_page()` as an abstract method, taking a single `OcrdFile` as input – and change all implementations in the modules accordingly (which is a lot of work, but trivial).
[...]

PR #577, opened 2020-08-25
utils/image: add polygon_for_parent w/ a test

kba:


bertsky (commented):
I don't see the point. I have already stalled #492 (which does the same thing), because [like I said](https://github.com/OCR-D/core/issues/489#issuecomment-634731069) this should be addressed at the PAGE model implicitly, rather than by requiring extra calls in every processor. (I thought we agreed on that?)

kba:
We have been discussing GT validation of coordinates again today, as well as possible applications for ocrd-sanitize here, that's why I wanted to understand how to use that function. I see this as a step towards the eventual goal of PAGE-level validation/cl;ipping, the adaptions to enable better error reporting in GdsCollector is in the same vein. I just momentarily wasn't aware of #492. Can't we not try to bring the coordinate logic from tesserocr into OCR-D/core first, while working towards PAGE-level methods?

bertsky:
> Can't we ~~not~~ try to bring the coordinate logic from tesserocr into OCR-D/core first, while working towards PAGE-level methods?

But the exception was merely an experiment. We don't want that in production code. So not only would we have to add the call to nearly every processor, it would also have to catch the exception each time. That's just too much effort IMO.

https://github.com/OCR-D/ocrd_tesserocr/pull/145 already demonstrates that ocrd_tesserocr correctly would have to catch the exception, ignoring the candidate. (I just rewrote with return type `None`.) 

Thinking about this from the PAGE model is a different picture, though. Functions like `add_TextRegion` would have to throw some `CoordsNotInParentException` which needs to be handled differently in the caller/processor.

I think you might draw 2 different conclusions from that:
1. Moving the check into the PAGE model implicitly is not worth it, because the processors still need to be updated (catching the exception), so one might as well keep it a separate function and opt-in (tolerating processors which don't use it and thus may produce inconsistent coords).
2. Moving the check into the PAGE model implicitly allows us to enter a (non-production) phase where users will see exceptions as long as the respective processors have not been fixed.

bertsky:
> Thinking about this from the PAGE model is a different picture, though. Functions like `add_TextRegion` would have to throw some `CoordsNotInParentException` which needs to be handled differently in the caller/processor.
> 
> I think you might draw 2 different conclusions from that:
> 
> 1. Moving the check into the PAGE model implicitly is not worth it, because the processors still need to be updated (catching the exception), so one might as well keep it a separate function and opt-in (tolerating processors which don't use it and thus may produce inconsistent coords).
> 2. Moving the check into the PAGE model implicitly allows us to enter a (non-production) phase where users will see exceptions as long as the respective processors have not been fixed.

After today's tech call, the picture has become even larger. #576 offers a new alley for making implementors aware of these problems – by feeding PAGE validation failures (along with call stack context if necessary) into dedicated loggers. _But_ most of these failures very likely would result in subsequent crashes of the follow-up processors anyway (because we cannot expect to make all of them robust against all sorts of input invalidities, even in the long term). Plus we already face the problem that processors can fail (due to exception or exit) on single pages (often only after computing a long series of pages correctly). This is not tolerable for a system aiming at productive use. So is option 2 above. Therefore, we already need a mechanism of catching a processor's exceptions and falling back to (say) the input page as output page. That very mechanism could then be used to let exceptions fall through when requested. This could be controlled on the command line with a new option (say `--debug`). The single place to handle this would be `ocrd.Processor.process`. (But for the page-wise fallback we would also need #322.) Processor implementors could then run (and `make test`) their code with `--debug` and start catching the exceptions from the PAGE model themselves accordingly.

PR #578, opened 2020-08-25
Deprecate workspace argument for ocrd workspace clone/init

kba:


bertsky (approved):
Looks good!

That leaves only `ocrd.cli.workspace`'s `validate_workspace.mets_url` argument to be deprecated yet.

BTW can we use the same mechanism to deprecate `--mets-basename` and `--directory` in favour of a new `--mets`? I am asking because it would be good to adapt usage/documentation only once...

kba:
> Looks good!
> 
> That leaves only `ocrd.cli.workspace`'s `validate_workspace.mets_url` argument to be deprecated yet.

https://github.com/OCR-D/core/pull/578/commits/10bf7a5bdc263abbc6c3d5d0d5c812594a99cf93

> BTW can we use the same mechanism to deprecate `--mets-basename` and `--directory` in favour of a new `--mets`? I am asking because it would be good to adapt usage/documentation only once...

https://github.com/OCR-D/core/pull/578/commits/fb2579ddda39491fe8d4ba130174c33994217113

I changed cli/workspace with the `--mets`/`--directory` logic we discussed in #517. Note that the deprecation logic is still missing so, at this change, this backwards-incompatible. I'll correct that tomorrow. But can you have a look whether I missed something?

ocrd/ocrd/cli/workspace.py
```diff
@@ -21,10 +22,20 @@ def clean_id(dirty):
...
 class WorkspaceCtx():
 
-    def __init__(self, directory, mets_basename, automatic_backup):
-        self.directory = directory
+    def __init__(self, directory, mets_url, automatic_backup):
+        if directory and not mets_url:
+            mets_url = join(directory, 'mets.xml')
+        elif not directory and mets_url:
+            if mets_url.startswith('http') or mets_url.startswith('https:'):
+                raise ValueError("--mets is an http(s) URL but no --directory was given")
```
bertsky, line 30: Yes, I think that is the only useful response under these circumstances.

ocrd/ocrd/cli/workspace.py
```diff
@@ -21,10 +22,20 @@ def clean_id(dirty):
...
+            directory = dirname(mets_url) or getcwd()
+        else:
+            directory = getcwd()
+            mets_url = join(directory, 'mets.xml')
```
bertsky, line 34: I believe there is a case missing here: **both** `directory` **and** `mets_url` might be given. And as a sub-case, if `mets_url` is local and its dirname does not equal the `directory` (perhaps after canonicalization of both), then also raise an exception `"--mets has a directory part inconsistent with --directory"`.

ocrd/ocrd/cli/workspace.py
```diff
@@ -34,15 +45,15 @@ def __init__(self, directory, mets_basename, automatic_backup):
 # ----------------------------------------------------------------------
 
 @click.group("workspace")
-@click.option('-d', '--directory', envvar='WORKSPACE_DIR', default='.', type=click.Path(file_okay=False), metavar='WORKSPACE_DIR', help='Changes the workspace folder location.', show_default=True)
-@click.option('-M', '--mets-basename', default="mets.xml", help='The basename of the METS file.', show_default=True)
+@click.option('-d', '--directory', envvar='WORKSPACE_DIR', type=click.Path(file_okay=False), metavar='WORKSPACE_DIR', help='Changes the workspace folder location.', show_default=True)
+@click.option('-m', '--mets', default=None, help='The path/URL of the METS file.')
```
bertsky, line 49: Now we don't have `show_default` anymore (because the defaults are non-trivially calulcated in WorkspaceCtx), we at least need to write them descriptively into `help`, e.g. `"[default: METS_URL directory or .]"` and `"[default: WORKSPACE_DIR/mets.xml]"`

bertsky (changes requested):
I think currently two things are still missing:
- a backwards-compatible way for `-M/--mets-basename` including deprecation warning
- not showing the already deprecated `mets_url` arguments in `init/clone/validate` in the help text anymore

ocrd/ocrd/cli/workspace.py
kba: Indeed, thanks. a199d78

ocrd/ocrd/cli/workspace.py
kba: b78e8b8

kba:
> a backwards-compatible way for -M/--mets-basename including deprecation warning

b78e8b8

ocrd/ocrd/cli/workspace.py
kba: 84dc4e1

ocrd/ocrd/cli/workspace.py
```diff
@@ -38,7 +42,7 @@ def __init__(self, directory, mets_url, automatic_backup):
         self.directory = directory
         self.resolver = Resolver()
         self.mets_url = mets_url
-        self.mets_basename = basename(mets_url)
+        self.mets_basename = mets_basename if mets_basename else basename(mets_url)
```
bertsky: IMO this is incorrect: If the user called with `-d somewhere -M METS.XML`, then this will (issue the deprecation warning and) set `self.mets_url = somewhere/mets.xml` (instead of `somewhere/METS.XML`) and then set `self.mets_basename` correctly (but without effect).

I suggest to let `mets_basename` default to `mets.xml` and use that in the two conditions joining `mets_url`. You should also drop `self.mets_basename` from the context class to avoid mistakenly using it henceforth.

bertsky (changes requested):

ocrd/ocrd/cli/workspace.py
kba: e9e76bc I hope I didn't mess up the mets_basename logic for the Workspace constructor. Tested various combinations of `-m`/`-M`/`-d` on the command line and added tests to cover all conditionals for WorkspaceCtx.

tests/test_utils.py
```diff
@@ -158,12 +158,24 @@ def test_pil_version(self):
...
         cwd = getcwd()
         with pushd_popd('/tmp'):
             self.assertEqual(getcwd(), '/tmp')
         self.assertEqual(getcwd(), cwd)
 
+    def test_pushd_popd_tempdir(self):
+        cwd = getcwd()
+        with pushd_popd(tempdir=True) as newcwd:
+            self.assertEqual(getcwd(), newcwd)
+            self.assertTrue(newcwd.startswith('/tmp'))
```
bertsky, line 171: Is this reliable across platforms? (I thought even on Unix Python must first try `TMPDIR`, `TEMP` and `TMP` first.)

bertsky (approved):
Excellent. Very thorough testing. Backwards compatibility is worth all the effort I think – users already have all kinds of scripts etc.

The one wish I still had (not showing the METS_URL argument in `clone/init/validate`'s help text anymore) is not easy to fulfill with current click I guess, so let's ignore this for now (and only make sure to avoid them in the other documentation).

tests/test_utils.py
kba: b74c5adf

kba:
> The one wish I still had (not showing the METS_URL argument in `clone/init/validate`'s help text anymore) is not easy to fulfill with current click I guess, so let's ignore this for now (and only make sure to avoid them in the other documentation).

bfe04ac8 Turns out, this is straightforward to implement: subclass `click.Command` with regex replacements against the genereated help.

> METS_URL argument in `clone`

`METS_URL` is still a required parameter to clone. You mean `WORKSPACE_DIR` for clone and `DIRECTORY` for init, right?

bertsky:
> [bfe04ac](https://github.com/OCR-D/core/commit/bfe04ac87df46c5a6c7c75a8bcc21267051ef5ae) Turns out, this is straightforward to implement: subclass `click.Command` with regex replacements against the genereated help.

Awesome!

> > METS_URL argument in `clone`
> 
> `METS_URL` is still a required parameter to clone. You mean `WORKSPACE_DIR` for clone and `DIRECTORY` for init, right?

Yes, sorry – that's what I actually meant.

Issue #539, opened 2020-07-16
Create workspaces from OAI-PMH links


wrznr:
Many libraries deliver METS files via a machine-readable OAI-PMH interface. E.g.,
![image](https://user-images.githubusercontent.com/26707219/87673582-12478700-c775-11ea-9ee5-fbd816b5ab36.png)

with https://digital.slub-dresden.de/oai/?verb=GetRecord&metadataPrefix=mets&identifier=oai:de:slub-dresden:db:id-497166623

Such links are not compatible with the `ocrd workspace clone` command. Since the direct link to the METS file is not provided, it is hard to create a workspace in this context. It would be fantastic if a harvesting option for OAI-PMH could be directly provided by `ocrd`.

mikegerber:
Yes, definitely an important task that I needed myself in the past!

stweil:
I don't understand this issue. https://digital.slub-dresden.de/oai/?verb=GetRecord&metadataPrefix=mets&identifier=oai:de:slub-dresden:db:id-497166623 is the direct link to the METS.

kba:
> I don't understand this issue. https://digital.slub-dresden.de/oai/?verb=GetRecord&metadataPrefix=mets&identifier=oai:de:slub-dresden:db:id-497166623 is the direct link to the METS.

It's not the METS itself, it's a [OAI-PMH GetRecord response](https://www.openarchives.org/OAI/openarchivesprotocol.html#GetRecord).

mikegerber:
> > I don't understand this issue. [https://digital.slub-dresden.de/oai/?verb=GetRecord&metadataPrefix=mets&identifier=oaideslub-dresden:db:id-497166623](https://digital.slub-dresden.de/oai/?verb=GetRecord&metadataPrefix=mets&identifier=oai:de:slub-dresden:db:id-497166623) is the direct link to the METS.
> 
> It's not the METS itself, it's a [OAI-PMH GetRecord response](https://www.openarchives.org/OAI/openarchivesprotocol.html#GetRecord).

Which contains the METS ;) 

@wrznr and me want an easy way to create a workspace from such a OAI-PMH response, i.e. make unwrapping the METS work with `clone`.

bertsky:
I assume this was fixed exhaustively by #581?

mikegerber:
I think this can be closed, as it's working fine:

~~~sh
% cd `mktemp -d`    
[...]

Issue #544, opened 2020-07-23
RFC: ocrd-sanitize script to preprocess/postprocess OCR-D workspaces`


kba:
METS/PAGE/ALTO provided by digitization workflow software or repositories will not always adhere [to the conventions we have in OCR-D](https://ocr-d.de/en/spec). OTOH the workspaces that are the result of OCR-D workflows contains a lot of redundant information that is not relevant for ingestion into production systems or contradict the local conventions of the production system.

Also, our conventions have been shifting and will continue to do so to meet the needs of users and developers.

Many users therefore have developed scripts to preprocess input and postprocess output of OCR-D.

OCR-D/core should provide a processor `ocrd-sanitize` which is only concerned with "housekeeping" of workspaces. Possible actions include:

* *Pruning* of `mets:fileGrp`, either by *allowlist* or *denylist*. I.e. remove `mets:fileGrp` and containing `mets:file` (and files on disk) that are not required anymore
* regex-based replacement of all `xlink:href` to match local conventions
* Removing all but the lowest level of `page:TextEquiv` information in PAGE-XML
* Approximating polygons with bounding boxes in PAGE-XML to support full-text-indexing
* Upgrading older PAGE-XML namespaces to the latest version (#503)
* Assigning persistent identifiers to work, pages, files ...

These are just some ideas, we'd love to hear yours. Please share your post-processing/post-processing scripts or feature requests for such a tool so we can develop a solution together for common tasks.

kba:
[@mikegerber](https://gitter.im/OCR-D/Lobby?at=5f197bbc15e5e72fb36edf7a)

> here is my collection of METS/PAGE file fixer scripts, as mentioned in the call: https://github.com/mikegerber/sbb-useful-hacks/tree/master/mets-fixers - not to be used lightly, no warranty, you have been warned 🚧 🚨 🚧

mikegerber:
I don't know if I missed the point a bit, but I do see two different groups of use cases here:

1. Repairing/maintaining invalid or outdated METS/workspaces:
  * Tools like https://github.com/mikegerber/sbb-useful-hacks/tree/master/mets-fixers
  * Upgrading older PAGE-XML namespaces to the latest version (#503)
  * regex-based replacement of all xlink:href to match local conventions (possibly)
  * Assigning persistent identifiers to work, pages, files ... (possibly)

2. Other post-processing

[...]

PR #581, opened 2020-08-28
handle oai response with workspace clone

M3ssman:
New Feature to enable workspace creation from OAI-Responses.
See https://github.com/OCR-D/core/issues/539 for details.


kba (approved):
LGTM, thank you very much. I like how concisely you implemented this and I've learned about pytest/unittest features I haven't used yet but will in the future.

I have refactored a few things and from my POV this can be merged.

However, I would really like to see some tests with real-world OAI-PMH endpoints to make sure the solution is robust enough. @wrznr @mikegerber can you try this and give short feedback whether this PR fits your use case?

wrznr (approved):

kba:
I've tested this with a few samples provided by @wrznr @maria-federbusch and @cneud from OAI-PMH endpoints at @SLUB and @StaatsBibliothekBerlin and they all work flawlessly

mikegerber:
> However, I would really like to see some tests with real-world OAI-PMH endpoints to make sure the solution is robust enough. @wrznr @mikegerber can you try this and give short feedback whether this PR fits your use case?

Sorry for not replying earlier: It's working fine 👍 (The METS needs more massaging after that, i.e. #544, but that's a different issue.)

PR #586, opened 2020-09-02
show help for no-arg processor calls, fix #562, fix #274

### LINKED ISSUE #274: output_file_grp not set to ocrd-tool.json default for processors
e.g.

```yaml
{
  output_file_grp: "OCR-D-IMG-BIN",
}
```

but in processor's `process` method:

```python
print(self.output_file_grp)
```

prints `OUTPUT` not `OCR-D-IMG-BIN`.

### LINKED ISSUE #562: What to do if --mets is unspecified and no mets.xml is present
Currently, a processor call, if `--mets` was not given and no `mets.xml` is present in $PWD, the `--help` message is given. The reasoning is that if a user calls a processor without any arguments, it makes more sense to provide general help instead of the very specific error message `mets.xml could not be found`.

However from feedback from @dla-kramski, @ElexaDigitalia_twitter and others, this behavior (in core since 2.12.0 https://github.com/OCR-D/core/issues/438 https://github.com/OCR-D/spec/pull/156) is confusing because it is also triggered if `-I` and `-O` are provided.

So I propose to change the behavior to this:

* If no options/arguments *at all* are provided to a processor, then print the `--help`
* If METS cannot be found, raise an error saying so

kba:
<del>This is a surprisingly tricky-to-implement problem: We want processor calls without any arguments to print the help screen and exit with code `1`, i.e. if `sys.argv[2:] == []`.</del>

It really isn't, it was just hard to test because sys.argv contained the arguments to pytest, not to the CLI. Turns out, sys.argv can be overridden at runtime for the tests and checking `not sys.argv[1:]` was enough

<del>
However, I could not find a way to get the un-procesed command line arguments to click. `get_current_context()` does have an `args` property but that only contains the remaining positional arguments after parsing.

If anyone knows a better way to get the click equivalent of `sys.argv[2:]`, I'd be interested!

For now, I've disabled the defaults for `input_file_grp`/`output_file_grp` (as proposed in https://github.com/OCR-D/core/issues/274) and check for all relevant CLI options one-by-one:

```python
    if not (dump_json or help or version or overwrite) and \
            mets == 'mets.xml' and \
            not (kwargs['input_file_grp'] or kwargs['output_file_grp']):
        processorClass(workspace=None, show_help=True)
        sys.exit(1)
```

As I said before, I would prefer a more robust solution. The clause above will also trigger for processors being called with `--mets mets.xml` but neither input nor output file group (such as the ocrd-sanitize processor...).
</del>

### DIFF
--- a/ocrd/ocrd/decorators/__init__.py
+++ b/ocrd/ocrd/decorators/__init__.py
@@ -30,6 +30,11 @@ def ocrd_cli_wrap_processor(
     overwrite=False,
     **kwargs
 ):
+    if not (dump_json or help or version or overwrite) and \
+            mets == 'mets.xml' and \
+            not (kwargs['input_file_grp'] or kwargs['output_file_grp']):
+        processorClass(workspace=None, show_help=True)
+        sys.exit(1)
     if dump_json or help or version:
         processorClass(workspace=None, dump_json=dump_json, show_help=help, show_version=version)
         sys.exit()
--- a/ocrd/ocrd/decorators/__init__.py
+++ b/ocrd/ocrd/decorators/__init__.py
@@ -30,9 +30,7 @@ def ocrd_cli_wrap_processor(
     overwrite=False,
     **kwargs
 ):
-    if not (dump_json or help or version or overwrite) and \
-            mets == 'mets.xml' and \
-            not (kwargs['input_file_grp'] or kwargs['output_file_grp']):
+    if not sys.argv[1:]:
         processorClass(workspace=None, show_help=True)
         sys.exit(1)
     if dump_json or help or version:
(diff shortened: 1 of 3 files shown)


bertsky (approved):
Energize!

mikegerber:
This *sometimes* breaks running a test using `click.testing.CliRunner`, e.g.:

~~~
   # Run through the OCR-D interface                                                                                   
    with working_directory(str(test_workspace_dir)):                                                                    
        runner = CliRunner()                                                                                            
        result = runner.invoke(ocrd_dinglehopper, [                                                                                                          
            '-m', 'mets.xml',                                                                                           
            '-I', 'OCR-D-GT-PAGE,OCR-D-OCR-CALAMARI',                                                                   
            '-O', 'OCR-D-OCR-CALAMARI-EVAL'                                                                             
        ])                                                                                                              
    print(result.output, file=sys.stderr)                                                                               
    assert result.exit_code == 0
~~~

The call/`invoke` prints help and exits with 1. (Funny, it's also a Heisenbug, it doesn't break when I run the test alone, but does when I run all the tests.)

Issue #589, opened 2020-09-04
ocrd process does not work with ocropy in docker


beckstefan:
When calling `cis-ocropy-binarize` tool with `ocrd process`

```
docker run --rm -u $(id -u) -v data/:/data -w /data -- ocrd/all:maximum-git ocrd process 'cis-ocropy-binarize -I OCR-D-IMG -O OCR-D-SEG-PAGE'
```

I get following error:

```
Traceback (most recent call last):
  File "/usr/bin/ocrd", line 33, in <module>
    sys.exit(load_entry_point('ocrd', 'console_scripts', 'ocrd')())
  File "/usr/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/build/core/ocrd/ocrd/cli/process.py", line 27, in process_cli
    run_tasks(mets, log_level, page_id, tasks, overwrite)
  File "/build/core/ocrd/ocrd/task_sequence.py", line 130, in run_tasks
    validate_tasks(tasks, workspace, page_id, overwrite)
  File "/build/core/ocrd/ocrd/task_sequence.py", line 99, in validate_tasks
    first_task.validate()
  File "/build/core/ocrd/ocrd/task_sequence.py", line 73, in validate
    param_validator = ParameterValidator(self.ocrd_tool_json)
  File "/build/core/ocrd/ocrd/task_sequence.py", line 54, in ocrd_tool_json
    self._ocrd_tool_json = json.loads(result.stdout)
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 342, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 1 column 5 (char 4)
```
[...]

Issue #540, opened 2020-07-16
ocrd process fails due to processor's logging mixed with JSON dump


GrazingScientist:
**Problem Description**
We use the docker image `ocrd/all:minimum` for testing purposes. We are running the [ocrd process example from the documentation](https://ocr-d.de/en/user_guide#calling-several-processors) in a slightly derived form:

```
ocrd process -l DEBUG --overwrite  
'cis-ocropy-binarize -I OCR-D-IMG -O OCR-D-SEG-PAGE'   
'tesserocr-segment-region -I OCR-D-SEG-PAGE -O OCR-D-SEG-BLOCK'  
 'tesserocr-segment-line -I OCR-D-SEG-BLOCK -O OCR-D-SEG-LINE'  
 'tesserocr-recognize -I OCR-D-SEG-LINE -O OCR-D-OCR-TESSEROCR'
```

(Please not that we removed `-p param-tess-fraktur.json` from the original example)
When running these processes each seperately, everything is fine and runs without error.

However, the `ocrd process` pipeline causes the following log output:

```
2020-07-15 14:13:32,208.208 INFO root - Overriding log level globally to DEBUG
2020-07-15 14:13:32,209.209 DEBUG ocrd.resolver - Deriving dst_dir /data/images from /data/images/mets.xml
2020-07-15 14:13:32,209.209 DEBUG ocrd.resolver - workspace_from_url
mets_basename='mets.xml'
mets_url='/data/images/mets.xml'
src_baseurl='/data/images'
dst_dir='/data/images'
2020-07-15 14:13:32,210.210 DEBUG ocrd.resolver.download_to_directory - directory=|/data/images| url=|/data/images/mets.xml| basename=|mets.xml| if_exists=|skip| subdir=|None|
2020-07-15 14:13:32,210.210 DEBUG ocrd.resolver.download_to_directory - Stop early, src_path and dst_path are the same: '/data/images/mets.xml' (url: '/data/images/mets.xml')
Traceback (most recent call last):
  File "/usr/bin/ocrd", line 8, in <module>
    sys.exit(cli())
  File "/usr/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
[...]

PR #591, opened 2020-09-04
remove preceding/trailing after { } when parsing --dump-json

kba:
Workaround for #540 / #589 

Removes any preceding/trailing content around the expected --dump-json output.

### DIFF
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -2,7 +2,8 @@
 Helper methods for running and documenting processors
 """
 from time import time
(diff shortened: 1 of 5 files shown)


bertsky (commented):
As I said in #589, `--dump-json` is not the only CLI action that needs to suppress import-time log messages. So do `--version`, `--show-resource` and `--list-resources`, and probably also the `ocrd bashlib` commands. This PR is not a solution, but a workaround that introduces more complexity.

kba:
Hopefully, not necessary anymore with #594

PR #600, opened 2020-09-14
ocrd_validators.page_validator: cope with empty RO

bertsky:
This fixes a crash in case `ReadingOrder` is empty (which is not allowed in PAGE, but does happen on the input side).


kba (approved):

bertsky:
Now that it's too late, I found that you have already integrated the (fixed) version from ocrd_tesserocr in core's ocrd_page_generateds – it's called `_get_recursive_reading_order`. How about exposing this to the API so `ocrd_validators` or processors can delegate to it? (There are still a couple of copies flying around ...)

kba:
>  you have already integrated the (fixed) version from ocrd_tesserocr in core's ocrd_page_generateds

Have I? Was not aware of that. Yes, exposing that at runtime is no problem.

PR #603, opened 2020-09-15
Do not serialize empty ReadingOrder in PAGE

kba:


bertsky (approved):
Looks good, thanks!

stweil:
Congratulation, that was commit 1800. Starting a new century ...

PR #606, opened 2020-09-16
ocrd_utils.image: fix bg tuple vs scalar (depending on mode)

bertsky:
as a workaround for python-pillow/Pillow#4925

### DIFF
--- a/ocrd_utils/ocrd_utils/image.py
+++ b/ocrd_utils/ocrd_utils/image.py
@@ -259,10 +259,14 @@ def rotate_image(image, angle, fill='background', transparency=False):
         image = image.copy()
         image.putalpha(255)
     if fill == 'background':
-        background = ImageStat.Stat(image).median
-        if image.mode in ['RGBA', 'LA']:
-            background[-1] = 0 # fully transparent
-        background = tuple(background)
+        background = ImageStat.Stat(image)
+        if len(background.bands) > 1:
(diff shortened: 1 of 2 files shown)


kba (approved):
LGTM AFAICT. A test that triggers the situation that led to the error would be useful.

bertsky:
> A test that triggers the situation that led to the error would be useful.

That could only be provided by suitable images. Perhaps we should create them artificially via `PIL.Image.new`. Or be extended existing tests on real images after converting them to other modes on the fly.

The error was triggered by 32-bit (mode `I` or `F`) or 16-bit (mode `I;16`) images.

kba:
> > A test that triggers the situation that led to the error would be useful.
> 
> That could only be provided by suitable images. Perhaps we should create them artificially via `PIL.Image.new`. Or be extended existing tests on real images after converting them to other modes on the fly.
> 
> The error was triggered by 32-bit (mode `I` or `F`) or 16-bit (mode `I;16`) images.

I've added a tiny test that fails with https://github.com/python-pillow/Pillow/issues/4925 in master but should succeed here 🤞

bertsky:
> I've added a tiny test that fails with [python-pillow/Pillow#4925](https://github.com/python-pillow/Pillow/issues/4925) in master but should succeed here :crossed_fingers:

Excellent!

Issue #403, opened 2020-01-14
Too restrictive file mode for mets.xml


stweil:
When `mets.xml` is created or modified by `ocrd`, the new file mode only allows access by the owner which is typically not sufficient for practical use:

    -rw------- 1 stweil stweil 822382 Jan 13 22:43 mets.xml

So it is always necessary to add a `chmod` command after processing with OCR-D.

The code should use the default file mode which respects umask.

stweil:
The problem is caused by `atomic_write`.

stweil:
https://www.oreilly.com/library/view/python-cookbook/0596001673/ch04s25.html shows a portable alternative solution.

Do we need atomic writes here at all? Just using `open` would fix the problem, too, and we could remove a dependency.

kba:
> I doubt that atomic_write does what is really needed

You can check the code here, it's really short https://github.com/untitaker/python-atomicwrites/blob/master/atomicwrites/__init__.py

I don't want to move away from `atomicwrites` unless there is a real reason not to use it, it seems to do what it says on the package.

> The code should use the default file mode which respects umask.

There's some entry points in atomicwrites to influence this, I'll have a look. See also https://github.com/untitaker/python-atomicwrites/issues/42

stweil:
> I don't want to move away from atomicwrites unless there is a real reason not to use it, it seems to do what it says on the package.

What is required? Example: Should it be possible to run several `ocrd workspace add` commands which add new entries to the same METS file in parallel?

Without trying it I would say it cannot work: Task 1 reads the data, task 2 reads / modifies / writes the data, task 1 writes the data and overwrites the data which was written by task 2. I did not see locks to prevent that scenario, and atomic_write would not help here.

stweil:
> You can check the code here.

I checked it before writing that I habe doubts.

bertsky:
> What is required? Example: Should it be possible to run several `ocrd workspace add` commands which add new entries to the same METS file in parallel?

One of the reasons for this was safe SIGINT (ctrl+c) IIRC.

stweil:
SIGINT handling could be done with `signal.signal`. I think it would be sufficient to ignore SIGINT while writing the METS file. Or do we expect that this file can be so large that it should be possible to abort the writing?

Maybe it would even make sense to block SIGINT while processing a page and handle it only after a page was processed.

kba:
> I think it would be sufficient to ignore SIGINT while writing the METS file

[...]

PR #608, opened 2020-09-19
Fix mode of METS file after atomic write (fix issue #403)

stweil:

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -1,5 +1,5 @@
 import io
-from os import makedirs, unlink, listdir, path
+from os import chmod, makedirs, umask, unlink, listdir, path
 from pathlib import Path
 
 import cv2
@@ -248,8 +248,13 @@ class Workspace():
         log.info("Saving mets '%s'", self.mets_target)
         if self.automatic_backup:
             WorkspaceBackupManager(self).add()
+
+        mask = umask(0)
+        umask(mask)
+        mode = 0o666 & ~mask
         with atomic_write(self.mets_target, overwrite=True) as f:
             f.write(self.mets.to_xml(xmllint=True).decode('utf-8'))
+        chmod(self.mets_target, mode)
 
     def resolve_image_exif(self, image_url):
         """


kba (approved):

ocrd/ocrd/workspace.py
```diff
@@ -248,8 +248,13 @@ def save_mets(self):
         log.info("Saving mets '%s'", self.mets_target)
         if self.automatic_backup:
             WorkspaceBackupManager(self).add()
+
+        mask = umask(0)
+        umask(mask)
+        mode = 0o666 & ~mask
```
bertsky, line 254: Shouldn't this use the permissions `self.mets_target` had _before_ (if it existed at all, otherwise `0o666`)?

stweil: That would require additional code. In most (= all typical) cases the result would be the same. Therefore I don't think that is necessary.

bertsky: Sure, but then why do you care about the umask? You could just set it to 666 regardless.

There are use cases for this already BTW: when serving workspaces on the web (e.g. as part of a repository) the service daemon should usually not be allowed write access. Or on a cluster users might share their workspaces but disallow write access beyond the group.

stweil: I use umask to get the default file mode for new files (typically `0622`). That's the same mode which we would have if we did not use `atomic_write`. And it fits the use cases which you describe. `0666` would allow write access for anyone which is not desired in most cases.

bertsky: > That's the same mode which we would have if we did not use `atomic_write`.

Not true. Python's `open` does not change the permissions if the file already existed, but your code (still) does.

> And it fits the use cases which you describe.

It happens to (if the umask is set that way). My point was that it's up to the _user_ to decide file permissions, and that should always be respected by programs like ours.

> `0666` would allow write access for anyone which is not desired in most cases.

Neither is reverting a file's permissions to the umask default.

stweil:
How should we proceed with the issue and this fix?

I think that the current situation with a mets.xml which cannot be read by anyone beside the owner is bad. Several times I now have used such files on our web server and only noted later that they could not work.

This pull request makes sure that the mets.xml has the same permissions like other files which are created by OCR-D processors. I think that is good and sufficient, and it solves my problem.

@bertsky thinks that any file permissions of an existing mets.xml should be preserved when the file is rewritten by OCR-D. I disagree, don't think that such additional code is necessary and would prefer to keep the code simple. If others think that preserving the permissions is better than using the default permissions and that it should be implemented now (it could be added any time later if someone misses it) I would not object. The final result would be the same for my use cases, so it would solve the initial problem, too.

bertsky:
> I think that the current situation with a mets.xml which cannot be read by anyone beside the owner is bad.

Of course.

> I think that is good and sufficient, and it solves my problem.

It just shifts the problem (still changed permissions).

> @bertsky thinks that any file permissions of an existing mets.xml should be preserved when the file is rewritten by OCR-D. I disagree, don't think that such additional code is necessary and would prefer to keep the code simple.

You still did not say _why_ the permissions should not be preserved.

Saving the actual permissions before the `atomic_write` and imposing them afterwards would actually be _simpler_ than your approach based on umask.

stweil:
> Saving the actual permissions before the atomic_write and imposing them afterwards would actually be simpler than your approach based on umask.

Feel free to prove that. I have no solution which is simpler than mine.

kba:
I also favor not messing with umask and just resetting the mets.xml file mode after atomic_write. I'll send a PR.

kba:
https://github.com/OCR-D/core/pull/625

stweil:
This PR is superseded by PR #625, so it can be closed.

ocrd/ocrd/workspace.py
kba: > Not true. Python's open does not change the permissions if the file already existed, but your code (still) does.

`os.open` does take umask into account: https://docs.python.org/3.5/library/os.html

> os.open(path, flags, mode=0o777, *, dir_fd=None)
Open the file path and set various flags according to flags and possibly its mode according to mode. When computing mode, the current umask value is first masked out.

Issue #343, opened 2019-11-11
save_image_file should set DPI for derived images


bertsky:
Currently, any information on image resolution provided in the original image (and made available via `OcrdExif` in `Workspace.image_from_page`) is ignored when saving derived images in the workspace (via `Workspace.save_image_file`). Due to PIL.Image format internals, the PNG then contains a setting of 72 DPI however. This might create problems for processors that look at the derived image files alone.

But this is hard to fix in core: the image passed to `save_image_file` could come from anywhere (and usually does not have an `info['dpi']`; even simple PIL.Image operations omit that in the result).

Realistically though, it will have been created some way from the source image file under the same `pageId`, and since rescaling is currently not permitted in the spec, one could assume the same DPI for all derived images.

wrznr:
> and usually does not have an info['dpi']

Why can't we check whether such information is available? Copy it, if that's the case and delete the default value if not? We should definitely not write default values which we know to be wrong (I added the label `bug`).

bertsky:
> > and usually does not have an info['dpi']
> 
> Why can't we check whether such information is available?

We could, but as I said: usually it is not. Or should I say: almost always. Any non-mutating PIL.Image operation (even those that wouldn't affect the meta-data) omits `info` from the result. We cannot change that. And we cannot processors to give us much more than what PIL.Image does.

> We should definitely not write default values which we know to be wrong (I added the label `bug`).

Okay, but that's probably a bug in Pillow's `save` for the PNG format.

bertsky:
> > We should definitely not write default values which we know to be wrong (I added the label `bug`).
> 
> Okay, but that's probably a bug in Pillow's `save` for the PNG format.

Oh, just found that this appears to have been fixed in Pillow 6.2.1. Since we cannot likely get any better, I will close for now.

kba:
> https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst#621-2019-10-21

Should we update to 6.2.1 then? Do you have a sample for me to test? Thanks!

bertsky:
Sorry, I was imprecise: It could have been fixed in an earlier version already.

I saw the following with 5.4.1 (leading up to this issue):
```sh
python -c "import PIL.Image; PIL.Image.open('repo/data/assets/scribo-test/data/OCR-D-IMG/OCR-D-IMG-orig_tiff.tif').save('test.png')"
identify -verbose test.png | grep Resolution:
 Resolution: 72x72
```

[...]

PR #611, opened 2020-09-21
workspace.save_image_file: set dpi on image.save, #343

kba:


ocrd/ocrd/workspace.py
```diff
@@ -836,6 +836,7 @@ def image_from_segment(self, segment, parent_image, parent_coords,
     def save_image_file(self, image,
                         file_id,
                         file_grp,
+                        dpi=None,
```
bertsky, line 839: The position of the new kwarg makes `test_save_image_file` (which does not use keys) fail.

ocrd/ocrd/workspace.py
```diff
@@ -851,7 +852,15 @@ def save_image_file(self, image,
         if not force and self.overwrite_mode:
             force = True
         image_bytes = io.BytesIO()
-        image.save(image_bytes, format=MIME_TO_PIL[mimetype])
+        if dpi and not isinstance(dpi, tuple):
+            dpi = (dpi, dpi)
+        elif not dpi:
+            # TODO - brittle, will this always find the original image?
+            orig_img_file = self.mets.find_files(pageId=page_id, mimetype='//^image')[0]
+            orig_img_pil = self.resolve_image_exif(orig_img_file.url)
```
bertsky, line 860: You're right. I did not think right. There might be many images (even derived images) for that pageId already. It's probably better than nothing though. But we should allow empty results here, or at least raise an appropriate error.

Also, I wonder whether this does not make preprocessing much more inefficient. I don't know about lxml's `xpath` and `iter` functions, but we surely should make our `find_files` into a generator so we don't have to step through the entire METS for just one match.

ocrd/ocrd/workspace.py
```diff
@@ -851,7 +852,15 @@ def save_image_file(self, image,
...
+            exif = OcrdExif(orig_img_pil)
+            dpi = (exif.xResolution, exif.yResolution)
```
bertsky, line 862: Should probably also check `exif.resolutionUnit == 'inches'` or divide by 2.54.

bertsky (changes requested):

ocrd/ocrd/workspace.py
kba: > There might be many images (even derived images) for that pageId already. It's probably better than nothing though. 

We could retrieve the PAGE XML by pageId and use the `imageFilename` attribute to find the original image by URL - but that will be even more inefficient I guess.

> but we surely should make our find_files into a generator so we don't have to step through the entire METS for just one match.

Good idea, on it.

ocrd/ocrd/workspace.py
kba: 552a7c9 test still fails because file isn't found but that's another issue, see below.

ocrd/ocrd/workspace.py
kba: 552a7c9

ocrd/ocrd/workspace.py
kba: https://github.com/OCR-D/core/issues/613

ocrd/ocrd/workspace.py
```diff
@@ -1024,7 +1025,17 @@ def save_image_file(self, image,
         if not force and self.overwrite_mode:
             force = True
         image_bytes = io.BytesIO()
-        image.save(image_bytes, format=MIME_TO_PIL[mimetype])
+        if dpi and not isinstance(dpi, tuple):
+            dpi = (dpi, dpi)
+        elif not dpi:
+            # TODO - brittle, will this always find the original image?
+            orig_img_file = self.mets.find_files(pageId=page_id, mimetype='//^image')[0]
```
bertsky, line 1032: We must adapt for the generator now, I believe.

```suggestion
            orig_img_file = next(self.mets.find_files(pageId=page_id, mimetype='//^image', local_only=True))
```

bertsky (approved):
We should move this forward. 

Looking up any existing image for DPI info (without making sure it's the original image) will work, as long as images only ever enter via this function (or some other DPI-preserving method).

Regarding (in)efficiency, we should immediately take action on the processor side, so the `dpi` kwarg is used everywhere. 

(BTW, if we split this into the opt-in / dpi-kwarg and fallback / find_files part, we will get a window where we can adapt processors without paying the costs already.)

bertsky:
Another idea I had: couldn't we at least try to force PIL.Image objects we ourselves produce (`Workspace.image_from_page` and `image_from_segment`) to retain or piggy-back the DPI info? Or even store that info in the workspace instance (e.g. by attaching the OcrdExif to the OcrdFile of the PAGE-XML, which itself could be cached in the OcrdMets; this would require rewriting Processor.input_files, but allow reusing these iterations)?

Issue #613, opened 2020-09-22
Refactor OcrdMets.find_files as a generator


kba:
If we're only interested in the first result, no need to iterate through all of them.

bertsky:
Plus it should help reduce the RSS when processing very large METS.

I believe it would be crucial to be able to use lxml in a generator-like fashion.

PR #614, opened 2020-09-22
make OcrdMets.find_files a generator, #613

kba:


ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -181,8 +189,7 @@ def find_files(self, ID=None, fileGrp=None, pageId=None, mimetype=None, url=None
             # If only local resources should be returned and f is not a file path: skip the file
             if local_only and not is_local_filename(f.url):
                 continue
-            ret.append(f)
-        return ret
+            yield f
```
bertsky, line 192: Nice, just a tiny change!

So IIUC, as long as `etree.xpath()` is an iterator (and not a list), we don't need to search through all files first (and  heap them up in memory). But as soon as `pageId` is passed in, because a pageId can be covered in multiple fileIds, we are forced to enumerate the files again. Correct?

bertsky (approved):
LGTM.

ocrd_models/ocrd_models/ocrd_mets.py
kba: Correct, AFAICT

PR #616, opened 2020-09-23
ocrd_utils make_file_id: respect mets:file @ID syntax, fix #605

### LINKED ISSUE #605: make_file_id does not ensure METS REGEX_FILE_ID
The current implementation of `make_file_id` does not yet guarantee that the generated ID succeeds later-on during serialization. 

Thus, if the generated ID begins with a number (e.g. because the `fileGrp/@USE` does, which is allowed) then `to_xml` will cause `ocrd_mets.add_file` to raise an exception `Invalid syntax for mets:file/@ID`.

I propose checking that `ocrd_models.ocrd_mets.constants.REGEX_FILE_ID` matches before returning the result, and attempting repairs (like prepending `"id"` or substituting forbidden characters) in case it does not.

kba:


ocrd_utils/ocrd_utils/constants.py
```diff
@@ -81,6 +83,9 @@
 # Prefix to denote query is regular expression not fixed string
 REGEX_PREFIX = '//'
 
+# Regex for valid mets:file/@ID
+REGEX_FILE_ID = regex_compile('^[A-Za-z][^:]*$')
```
bertsky, line 87: That would include spaces and arbitrary Unicode codepoints after the first letter.

[Here](https://stackoverflow.com/a/12332162) is a proposal for a general xs:ID regex. (Probably better to not be over-restrictive either, who knows what we will use that constant for in the future.)

```suggestion
REGEX_FILE_ID = regex_compile(r'^[a-zA-Z_][\w.-]*$')
```

ocrd_utils/ocrd_utils/str.py
```diff
@@ -79,6 +80,9 @@ def make_file_id(ocrd_file, output_file_grp):
         while ocrd_file.mets.find_files(ID=ret):
             n += 1
             ret = concat_padded(output_file_grp, n)
+    if not REGEX_FILE_ID.fullmatch(ret):
+        ret = ret.replace(':', '_')
+        ret = re.sub(r'^(\d+)', r'id_\1', ret)
```
bertsky, line 85: Why not attempt full repair:
```suggestion
        ret = re.sub(r'^([^a-zA-Z_])', r'id_\1', ret)
        ret = re.sub(r'[^\w.-]', r'', ret)
```

This still does not guarantee we are unique, but this is as good as it gets.

bertsky (changes requested):
Excellent!

Issue #617, opened 2020-09-24
`ocrd workspace remove-group` fails to remove existing group


stweil:
[This METS file](https://digi.bib.uni-mannheim.de/fileadmin/digi/477429599/477429599.xml) contains a `FULLTEXT` file group made for ABBYY. It must be removed before adding a new file group with OCR-D results. This does not work with `ocrd remove-group`:

    ocrd workspace -m 477429599.xml remove-group --recursive FULLTEXT
    [...]
    Exception: File not locally available <OcrdFile fileGrp=TEMP, ID=FILE_0001_ALTO, mimetype=text/xml, url=https://digi.bib.uni-mannheim.de/fileadmin/digi/477429599/alto/477429599_0001.xml, local_filename=---]/>

kba:
Try `remove-group --recursive --force FULLTEXT`

>   `-f, --force`      Continue removing even if group or containing files not found in METS

stweil:
I did that of course. It works neither, but throws a different error.

stweil:
Pull request #618 fixes the `AttributeError` with `--force`.

stweil:
PR #618 was merged, so this is fixed now, thank you.

PR #618, opened 2020-09-24
Allow removal of file groups without local file (fix issue #617)

stweil:


stweil:
@kba, it would help to include this in the next release of ocrd_all.

kba:
> @kba, it would help to include this in the next release of ocrd_all.

It could have even been in the current release ;)

PR #619, opened 2020-09-25
Travis CI: Test on modern versions of Python

cclauss:


bertsky:
Wow, I am surprised to see we still use TravisCI.

@kba is there any reason to keep it? We could easily do the coverage tests in CircleCI as well...

kba:
> @kba is there any reason to keep it? We could easily do the coverage tests in CircleCI as well...

No, I just have to switch the codecov.io integration from Travis to Circle CI.

Also, we will have to support 3.5 until at least Q1 2021.

Issue #351, opened 2019-11-20
Makefile: sequential install does not work


bertsky:
When I do `make install`, this will currently run one `pip install` for each package in core sequentially. But since each of those depend on each other, and the independently called pip does not know about the others, it will try to satisfy those from the package index instead of the local installation. This can create inconsistencies when the features / API change, but the version does not. 

For example this happened recently:

```
Traceback (most recent call last):
  File "/usr/local/bin/ocrd", line 5, in <module>
    from ocrd.cli import cli
  File "/usr/local/lib/python3.6/site-packages/ocrd/cli/__init__.py", line 3, in <module>
    from ocrd.cli.ocrd_tool import ocrd_tool_cli
  File "/usr/local/lib/python3.6/site-packages/ocrd/cli/ocrd_tool.py", line 7, in <module>
    from ocrd.decorators import parameter_option
  File "/usr/local/lib/python3.6/site-packages/ocrd/decorators.py", line 7, in <module>
    from ocrd_utils import (
ImportError: cannot import name 'parse_json_string_or_file'
```

This is `ocrd.decorators` from master (2.0.0) trying to use `ocrd_utils` from PyPI (2.0.0).

IMO we need to bring all packages together into one pip call. I know this is [problematic](https://github.com/pypa/pip/issues/1386), since pip semantics is _all-or-nothing_ instead of _one-by-one_. But maybe we can at least come up with something that usually does work correctly, and breaks visibly when it does not?

Ideas/options:
- calculate the hashes of all subpackages and then use `--require-hashes`
- install one by one, but then reinstall all at once with `--no-index`
- build wheel files locally, then install these all at once

bertsky:
>     * build wheel files locally, then install these all at once

This does not work. Installation from wheels behaves exactly the same as installation from directories.

Simpler (and successful in the current situation):
* install one by one, but then install again one by one with `--no-deps`

bertsky:
@kba do you want me to make a PR based on the latter solution? (We need this urgently for ocrd_all.)

bertsky:
> We need this ~urgently~ for ocrd_all – perhaps I should first add a workaround there?)

See [here](https://github.com/bertsky/ocrd_all/commit/a6843733d66f20bce0f5d821cfac1eedb05e26d6)

kba:
[...]

PR #620, opened 2020-10-01
Require numpy<1.19.0 to satisfy requirements of tensorflow

stweil:

### DIFF
--- a/ocrd_utils/requirements.txt
+++ b/ocrd_utils/requirements.txt
@@ -1,2 +1,4 @@
 Pillow >= 7.2.0
-numpy >= 1.17.0
+# All current versions of TensorFlow require numpy < 1.19.0,
+# so make sure that now newer version is installed here.
+numpy >= 1.17.0, < 1.19.0


stweil:
It works, but only for a moment. As we force a reinstall for all pip installations, numy is soon replaced with a newer version again. So this is not a working fix, sorry.

stweil:
@bertsky, what was the reason why Makefile currently uses `--force-reinstall`? The comment for that code is rather short: "install again forcefully without depds (to ensure the binary itself updates)".

bertsky:
> what was the reason why Makefile currently uses `--force-reinstall`? The comment for that code is rather short: "install again forcefully without depds (to ensure the binary itself updates)".

It is a way to allow `make` to see what `pip` is doing. When we `pip install` successfully (after it was deemed necessary by make), it is not guaranteed that the actual entry point (script file) which we model as target in `make` gets updated. The timestamp is necessary to avoid re-making (e.g. multiple scripts for one module) and force propagation (e.g. other module depends on script). So by doing a normal installation with all dependencies followed by a forced re-installation without dependencies we ensure all scripts are new. 

I know the `pip install && pip install --no-deps --force-reinstall` recipe is clunky, but have not conceived of a better solution yet. There is no unique file that we can count on gets always installed by pip (as there are too many different packaging metadata conventions/cases). We could `touch -c $@` instead, but that just has different problems. What if the module changed the name of a script? What if pip failed but did update the script?

bertsky:
I think the elephant in the room here is still pip being unable to solve conflicting requirements. If core now wants `>=1.17, <1.19` but TF wants `<2.0,>=1.16.0`, why don't we end up with the former? Instead, the core request is processed first, giving us some 1.18, but then the TF request overrides it, giving us 1.19. (Correct?)

Maybe we can (temporarily) solve this by making `numpy >=1.17, <1.19` required by all our modules themselves, which should be the last request encountered by pip?

stweil:
> TF wants <2.0

TF wants <1.19.0. That's why I added that for core here.

bertsky:
> TF wants <1.19.0. That's why I added that for core here.

Right, my hypothesis was wrong. It does not happen like that, but because of the **other dependencies** that pip happens to process after it sees our `>=1.17, <1.19` from `ocrd_utils`. So the elephant is still the same (pip's non-reproducible order-dependent conflict resolution).

For example, I saw after ocrd_utils:
- `numpy>=1.13.3` from `PyYAML`
- `numpy>=1.17.0` from `pyrsistent`

But more pressingly, our core install recipe seems to be wrong, because it **also** brings in `numpy>=1.17.0` for the older `ocrd_models` and `ocrd_modelfactory` via PyPI. Frankly, I have no idea why in #351 we used a different recipe (i.e. `pip install  --force-reinstall` followed by `pip install --no-deps`) than in the other rules (i.e. `pip install` followed by `pip install --no-deps --force-reinstall`). If I exchange that, the problem goes away – even without the need for `PIP_OPTIONS=--use-feature=2020-resolver`).

See https://github.com/OCR-D/ocrd_all/pull/204

@kba since this is urgent AFAICT.

PR #623, opened 2020-10-09
processor help: add module/processor/process docstring

bertsky:
Produces:

```
ocrd-cis-ocropy-deskew -h

Usage: ocrd-cis-ocropy-deskew [OPTIONS]

  Deskew regions with ocropy (by annotating orientation angle and adding AlternativeImage)

  > Deskew the regions of the workspace.

  > Open and deserialise PAGE input files and their respective images,
  > then iterate over the element hierarchy down to the TextRegion
  > level.

  > Next, for each file, crop each region image according to the layout
  > annotation (via coordinates into the higher-level image, or from the
  > alternative image), and determine the threshold for binarization and
  > the deskewing angle of the region (up to ``maxskew``). Annotate the
  > angle in the region.

  > Add the new image file to the workspace along with the output
  > fileGrp, and using a file ID with suffix ``.IMG-DESKEW`` along with
  > further identification of the input element.

  > Produce a new output file by serialising the resulting hierarchy.

Options:
  -I, --input-file-grp USE        File group(s) used as input
  -O, --output-file-grp USE       File group(s) used as output
  -g, --page-id ID                Physical page ID(s) to process
  --overwrite                     Remove existing output pages/images
                                  (with --page-id, remove only those)
  -p, --parameter JSON-PATH       Parameters, either verbatim JSON string
                                  or JSON file path
  -P, --param-override KEY VAL    Override a single JSON object key-value pair,
                                  taking precedence over --parameter
  -m, --mets URL-PATH             URL or file path of METS to process
  -w, --working-dir PATH          Working directory of local workspace
  -l, --log-level [OFF|ERROR|WARN|INFO|DEBUG|TRACE]
                                  Log level
  -J, --dump-json                 Dump tool description as JSON and exit
  -h, --help                      This help message
  -V, --version                   Show version

Parameters:
   "maxskew" [number - 5.0]
    modulus of maximum skewing angle to detect (larger will be slower, 0
    will deactivate deskewing)
   "level-of-operation" [string - "region"]
    PAGE XML hierarchy level granularity to annotate images for
    Possible values: ["page", "table", "region"]

Default Wiring:
  ['OCR-D-SEG-BLOCK', 'OCR-D-SEG-LINE'] -> ['OCR-D-SEG-BLOCK', 'OCR-D-SEG-LINE']


```

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -6,6 +6,7 @@ __all__ = ['Processor', 'generate_processor_help', 'run_cli', 'run_processor']
 
 import os
 import json
+import inspect
 from ocrd_utils import VERSION as OCRD_VERSION, MIMETYPE_PAGE
 from ocrd_validators import ParameterValidator
 from ocrd_models.ocrd_page import MetadataItemType, LabelType, LabelsType
@@ -65,7 +66,15 @@ class Processor():
         self.parameter = parameter
 
     def show_help(self):
-        print(generate_processor_help(self.ocrd_tool))
+        doc = ''
+        module = inspect.getmodule(self)
+        if module and module.__doc__:
+            doc += '\n' + inspect.cleandoc(module.__doc__)
+        if self.__doc__:
+            doc += '\n' + inspect.cleandoc(self.__doc__)
+        if self.process.__doc__:
+            doc += '\n' + inspect.cleandoc(self.process.__doc__)
+        print(generate_processor_help(self.ocrd_tool, doc=doc))
 
     def show_version(self):
         print("Version %s, ocrd/core %s" % (self.version, OCRD_VERSION))
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -6,7 +6,6 @@ __all__ = ['Processor', 'generate_processor_help', 'run_cli', 'run_processor']
 
 import os
 import json
-import inspect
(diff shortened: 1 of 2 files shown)


bertsky:
(force pushed after fixing the indentation and trailing newlines)

ocrd/ocrd/processor/base.py
```diff
@@ -65,7 +66,15 @@ def __init__(
         self.parameter = parameter
 
     def show_help(self):
-        print(generate_processor_help(self.ocrd_tool))
+        doc = ''
```
kba, line 69: Can you move the docstring logic to generate_processor_help, passing a `processor_module` argument to it? That way all help generating code is in one place.

kba (changes requested):
Good idea!

ocrd/ocrd/processor/base.py
bertsky: What do you mean by `processor_module`, the **instance** (`self`) of the Processor implementation? (Because that's what it would take – `generate_processor_help` only has the tool json, but we need to access all the implementation's docstring attributes.)

kba: Yes, `self`:

```python
generate_processor_help(ocrd_tool, processor_instance=self)
```

resp.

```python
def generate_processor_help(ocrd_tool, processor_instance=None):
  # ...
  if processor_instance:
    module = inspect.getmodule(processor_instance)
    # ...
```

bertsky: Done!

kba (approved):

PR #624, opened 2020-10-09
Processor.input_files: handle list of pageId

kba:


bertsky (changes requested):
Sorry, I am just seeing this now: We actually have to rewrite the whole function. Up above, we filter by MIME type alone. So whenever (for some list of pageIds) there is even a **single** page with PAGE annotation, only that page survives.

What we need instead:
- `find_all_files` filtering MIME for PAGE **or** `image/*`
- aggregate results by page ID, and iterate:
- if the page has a PAGE hit, keep only that
- else if it has a single hit, keep it
- else raise

bertsky:
Superseded by #635

PR #625, opened 2020-10-15
wrap atomic_write preserving permissions of mets.xml

kba:


ocrd_utils/ocrd_utils/os.py
```diff
@@ -53,4 +55,12 @@ def unzip_file_to_dir(path_to_zip, output_directory):
...
     z.close()
 
-
+@contextlib.contextmanager
+def atomic_write(fpath, overwrite=False):
+    if exists(fpath):
+        mode = stat(fpath).st_mode
+    else:
+        mode =  0o664
+    with atomic_write_(fpath, overwrite=True) as f:
```
bertsky, line 64:
```suggestion
    with atomic_write_(fpath, overwrite=overwrite) as f:
```

But I am not sure about `overwrite` generally. The documentation says:
> :param overwrite: If set to false, an error is raised if ``path`` exists.
> Errors are only raised after the file has been written to.  Either way,
> the operation is atomic.

I cannot see any exceptions being raised in the implementation (but perhaps `os.link` does if the target exists?). But I would consider this semantic a bad interface – after all, if I don't want to overwrite, then that case should be catched _before_ writing. Since you introduced that distinction here anyway, we could `raise` ourself...

tests/test_workspace.py
```diff
@@ -278,6 +279,19 @@ def test_image_from_page_basic(self):
             img, info, exif = ws.image_from_page(pcgts.get_Page(), page_id='PHYS_0017')
             self.assertEquals(info['features'], 'binarized,clipped')
 
+    def test_mets_permissions(self):
+        initLogging()
+        with TemporaryDirectory() as tempdir:
+            ws = self.resolver.workspace_from_nothing(tempdir)
+            ws.save_mets()
+            mets_path = join(ws.directory, 'mets.xml')
+            assert filemode(stat(mets_path).st_mode) == '-rw-rw-r--'
```
stweil, line 288: This test fails for me:

     FAILED tests/test_workspace.py::TestWorkspace::test_mets_permissions - AssertionError: assert '-rw-r--r--' == '-rw-rw-r--'

I am afraid that it won't work without using umask.

ocrd_utils/ocrd_utils/os.py
kba: The difference is that with `overwrite=True`, it uses `os.rename`, otherwise `os.link` followed by `os.unlink`.

> (but perhaps os.link does if the target exists?)

Yes, it does, I just tested it.

Now I'm not sure how `os.rename` does things internally but `os.rename` is what I would use, hence the `overwrite=True`. It shouldn't be parameterizable, I see no reason to use link/unlink instead of rename.

bertsky: I see, I agree, I just don't understand why you still expose an `overwrite` kwarg then, though. (You could also simplify the usage in workspace.py.)

kba: > I just don't understand why you still expose an overwrite kwarg then

That's what I meant, we should not expose this. Will change that.

kba: f92c5a3c

tests/test_workspace.py
kba: @stweil Can you test again with f92c5a3c? This incorporates your `umask` approach from #608 with a workaround by @pabs3 from https://github.com/untitaker/python-atomicwrites/issues/42

pabs3:
Please note the security and other issues mentioned in the python- tomicwrites issue.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

kba:
> Please note the security and other issues mentioned in the python- tomicwrites issue.

Noted. We're using atomicwrites to avoid `SIGINT` in the middle of writing, permission escalation is not an issue here. Thanks for the workaround.

tests/test_workspace.py
```diff
@@ -286,7 +286,7 @@ def test_mets_permissions(self):
             mets_path = join(ws.directory, 'mets.xml')
             mask = umask(0)
             umask(mask)
-            assert '%o' % (stat(mets_path).st_mode & ~mask) == '100644'
+            assert (stat(mets_path).st_mode & ~mask) == 0o100664 & ~mask
```
stweil, line 289: I think that masking on the left side should not be needed.

kba: Thanks! That was driving me crazy :)

Issue #590, opened 2020-09-04
logging: datefmt not picked up under certain conditions


kba:
(from https://github.com/OCR-D/core/issues/589)

When setting up logging, we define the `%(asctime)s` part of the log message as `%H:%M:%S`.

However, when running in docker, this is not set correctly, apparently:

```
$ docker run -v /tmp:/data --rm -it ocrd/all:minimum ocrd log info foo
2020-09-04 11:11:43,513.513 INFO root - foo
$ ocrd log info foo
11:11:51.714 INFO root - foo
```

When I run the `test_logging.py` test (which does verify the format with a regex) in docker, all works fine, log messages have the right date format. But outside the unit tests, it's the wrong date format.

I'm temporarily baffled. If anyone has an idea as to why this happens or how to fix it, I'd appreciate to hear.

kba:
ht @M3ssman we have typos in the ocrd_logging.conf. `datefm` should be `datefmt`.

Issue #626, opened 2020-10-18
CLI log-level ignored if config file is present


bertsky:
It seems we were not thorough enough in #599: Currently `--log-level` has no effect when some `~/ocrd_logging.conf` is present. 

IIUC this is because `ocrd_utils.logging.initLogging` comes after `setOverrideLoglevel` (so all its logger `setLevel`s get wiped out and only `global _overrideLogLevel` survives), but `logging.config.fileConfig` of course does not honour `ocrd_utils.logging.getLogger`'s `_overrideLogLevel`.

M3ssman:
I'm totally aware of stuff like this is hard to catch, but I really like to see a test-fixture setup for this situation that can be integrated into `test_logging.py` even as a sort of regression test and, as first step, to reproduce the mentioned behavior.

kba:
> It seems we were not thorough enough in #599: Currently `--log-level` has no effect when some `~/ocrd_logging.conf` is present.

How can I reproduce this?

```
cp $OCRD/ocrd_utils/ocrd_logging.conf .
ocrd -l DEBUG log debug yay
```

does log the debug message.

```
ocrd -l ERROR log error foo
```

does NOT log, as is expected.

I'm trying to make sure that https://github.com/OCR-D/core/pull/628 fixes the issue.
```

**Comment by bertsky** [comment]
> How can I reproduce this?
> 
> ```
> cp $OCRD/ocrd_utils/ocrd_logging.conf .
> ocrd -l DEBUG log debug yay
> ```
> 
> does log the debug message.
> 
> ```
> ocrd -l ERROR log error foo
> ```

Not sure about `ocrd log`, but a real processor should show the pattern.

**Comment by kba** [comment]
I've tried with `ocrd-dummy`:

```
cp ocrd_utils/ocrd_logging.conf .
ocrd-dummy -l ERROR -I FOO -O BAR
```

logs nothing as expected

```
ocrd-dummy -l TRACE -I FOO -O BAR
17:44:08.253 DEBUG ocrd.resolver.workspace_from_url - Deriving dst_dir /home/kba/build/github.com/OCR-D/monorepo/ocrd_all/core from /home/kba/build/github.com/OCR-D/monorepo/ocrd_all/core/mets.xml
17:44:08.255 DEBUG ocrd.resolver.workspace_from_url - workspace_from_url
...
```

I tried it with `olahd-client` as well, same result. What am I doing wrong here?

bertsky:
[...]

PR #628, opened 2020-10-19
[app][fix] fix default logging conf #590

M3ssman:
Attempts to fix https://github.com/OCR-D/core/issues/590 and add testcase


kba (approved):
Thanks! I'll test it and merge if that works.

kba:
Tests pass and #590 is solved by this, https://github.com/OCR-D/core/issues/626 remains. Thanks @M3ssman

Issue #595, opened 2020-09-07
CUDA version woes


kba:
@mikegerber has been systematically testing CUDA and tensorflow versions. Unfortunately, the compatibility table apparently goes like this:

| tensorflow | CUDA |
| ---- | ---- |
| >= 1.15, < 2.1 | 10.0 |
| 2.1, 2.2, 2.3 | 10.1 |
| 2.3 | 11 (not sure about that either) |

I cannot test this - @bertsky but it basically worked for you with nvidia-docker / CUDA 11, i.e. core itself but you haven't tested with ocrd_all I suppose?

bertsky:
> I cannot test this - @bertsky but it basically worked for you with nvidia-docker / CUDA 11, i.e. core itself but you haven't tested with ocrd_all I suppose?

Not yet, sorry. And it's possible the TF I ran was self-built, I am not sure. I'll report back when my CUDA device and I found some spare time.

mikegerber:
It's not really systematic (yet), I just had problems and had a look on combinations that work reproducibly. However I can produce a complete matrix of working and non-working combination. I'll check it out this week.

I did not/will not test self-built TensorFlow versions, because that will be an even greater pain in the butt :-)

mikegerber:
I've added CUDA 11.0 in a new script https://cvs.moegen-wir.net/mikegerber/test-nvidia/src/branch/master/run-docker-compatibility-matrix, no success with that one.

Tested: Combinations of
* pip-installable `tensorflow-gpu == 1.15.*` and `tensorflow == 2.*` (= TF 2.3 currently)
* CUDA Docker images 10.0, 10.1, 10.2 with CUDNN 7, 11.0 with CUDNN 8

Only combinations that work with GPU support:
* `tensorflow-gpu == 1.15.*` with CUDA 10.0
* `tensorflow == 2.*` with CUDA 10.1

Full log:
~~~
== tf1 nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
GPU 0: GeForce RTX 2080 (UUID: GPU-612ce75c-1340-772b-039c-2a83a3ea5c95)
TensorFlow 1.15.3
GPU available: True
== tf1 nvidia/cuda:10.1-cudnn7-runtime-ubuntu18.04
GPU 0: GeForce RTX 2080 (UUID: GPU-612ce75c-1340-772b-039c-2a83a3ea5c95)
TensorFlow 1.15.3
GPU available: False
[...]

PR #629, opened 2020-10-19
Use nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04 as the base image

Witiko:
As discussed in https://github.com/OCR-D/ocrd_calamari/issues/46, `calamari-predict` takes [more than 1/3 of the duration](https://github.com/OCR-D/ocrd_calamari/issues/46#issuecomment-710028497) of [the recommended workflow](https://ocr-d.de/en/workflows#best-results-for-selected-pages) on the CPU. Therefore, the `ocrd/all:*-cuda` images should above all support `calamari-predict`, so that it can use GPUs.

Currently, the `ocrd/all:*-cuda` images are based on the `nvidia/cuda:11.0-runtime-ubuntu18.04` image, whereas `calamari-predict` [uses Tensorflow 1.15](https://github.com/OCR-D/ocrd_calamari/blob/master/requirements.txt#L1), which [requires CUDA Toolkit 10.0](https://www.tensorflow.org/install/source#gpu) and the CUDNN7 dynamic library.

This pull request changes the base image to `nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04`, so `calamari-predict` can use GPUs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ docker docker-cuda:
 	docker build -t $(DOCKER_TAG) --build-arg BASE_IMAGE=$(DOCKER_BASE_IMAGE) $(DOCKER_ARGS) .
 
 # Build docker GPU / CUDA image
-docker-cuda: DOCKER_BASE_IMAGE = nvidia/cuda:11.0-runtime-ubuntu18.04
+docker-cuda: DOCKER_BASE_IMAGE = nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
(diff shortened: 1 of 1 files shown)


kba:
Thanks, @mikegerber what do you think?

mikegerber:
It's beneficial as this would at least support TF1.15 (but not TF2.3, which ocrd_calamari will depend on soon.) The current situation is that none of the TF versions are supported by the CUDA 11 base image.

See #595. What's the state of the discussion around thin container images (which I understand as "put every processor in its own image/container")?

Issue #621, opened 2020-10-02
document initLogging requirement


bertsky:
Using any of core's API (even just `ocrd_utils`) now requires calling `initLogging` at some point prior to usage, or the respective subpackage's `getLogger` will complain with `CRITICAL`.

So not only is it not correct anymore to do `ocrd_utils.getLogger` at module level, any program or library using core must also do `ocrd_utils.initLogging` during its initialization (but not necessarily on the module level) now.

This must be reflected somewhere in the documentation _prominently_.

M3ssman:
Three thumbs up!
Without this call any logging records in any other application that uses stuff from `ocrd-core` gets dropped. Why become all loggers silently `PropagationShyLoggers`? All our logs vanished after updating from 2.16 (behavior changed from v2.16 to v2.17, as I realized just by now).

bertsky:
> Without this call any logging records in any other application that uses stuff from `ocrd-core` gets dropped. Why become all loggers silently `PropagationShyLoggers`?

Sorry to hear this hit you in the back. 

I don't understand what you say there exactly, though. So you have a program that imports (some of) core's packages, but does not use the Processor CLI decorator (which brings `ocrd_utils.initLogging`), and that program does not give you _any_ log messages at all anymore, or just not the OCR-D log messages? 

Also, what's a `PropagationShyLogger`?

kba:
> The reason to introduce this was to have a short-cut to silence logging from all 3rd-party-libs ocrd uses ... ?

The reason for doing this was to prevent libraries like tensorflow non-processing-related logging in processors from interfering with `sys.stdout`/`sys.stderr`, so as not to break output like `--dump-json`.

>  This also turns any other logger from an app that uses ocrd_core-library into this type. 

That was intentional and this behavior, that `PropagationShyLogger` is set as the default logger, has been around a while. What is new is that in the absence of initialization, logging is completely disabled. I can see how this can be a problem, sorry for the inconvenience.

Where should I document this?

M3ssman:
If the main reason to turn off logging totally is the behavior of tensorflow, question: It is possible to integrate tensowflow into the default logging conf file? If so, I'd prefer this, since this too documents that there's a problem with tensorflow logging if left alone.

kba:
> It is possible to integrate tensowflow into the default logging conf file?

We should probably mirror the builtin defaults to the logging.conf, i.e. 

```
        logging.getLogger('').setLevel(logging.INFO)
[...]

PR #630, opened 2020-10-20
Fix/enhance configuration

M3ssman:
Some draft regarding https://github.com/OCR-D/core/issues/621

### DIFF
--- a/ocrd_utils/ocrd_logging.conf
+++ b/ocrd_utils/ocrd_logging.conf
@@ -47,15 +47,15 @@ handlers=consoleHandler
 # "qualname" must match the logger label used in the corresponding 
 # ocrd modul
 # see in the modul-of-interrest (moi)
-
-
+#
+# example configuration entry
 #
 # logger ocrd.workspace
 #
-[logger_ocrd_workspace]
-level=DEBUG
-handlers=fileHandler
-qualname=ocrd.workspace
+#[logger_ocrd_workspace]
+#level=DEBUG
+#handlers=fileHandler
+#qualname=ocrd.workspace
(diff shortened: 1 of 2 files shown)


ocrd_utils/ocrd_logging.conf
```diff
@@ -52,6 +53,19 @@ handlers=consoleHandler
 #handlers=fileHandler
 #qualname=ocrd.workspace
 
+
```
kba, line 56: The PIL restriction from the builtin config omitted because it is `INFO` anyway, correct?

kba (approved):
LGTM, thanks.

ocrd_utils/ocrd_logging.conf
M3ssman: Yes, since it is the default level. But If you mind, I can add this as well

M3ssman: Shall `ocrd` be integrated, too? 
Be aware, this might have side-effects, because it is the far most top ocrd logging channel.

kba: > Shall `ocrd` be integrated, too?
> Be aware, this might have side-effects, because it is the far most top ocrd logging channel.

Since we don't specify logging behavior for `ocrd*` in the builtin config, we should not in the config file.

I also would leave the `[logger_ocrd_workspace]` commented out as an example - setting this to `DEBUG` by default will be very verbose.

M3ssman: This makes sense, I adopt your suggestions asap

PR #631, opened 2020-10-20
logging: set non-root logger to NOTSET also when loading from config file

kba:
@bertsky This was why I originally had `setOverrideLoglevel` as part of `intiLogging` in #599: When loading from config file and setting the root logger to the override loglevel, all existing loggers must be set to `NOTSET`.

### DIFF
--- a/ocrd_utils/ocrd_utils/logging.py
+++ b/ocrd_utils/ocrd_utils/logging.py
@@ -141,6 +141,10 @@ def initLogging():
         logging.getLogger('tensorflow').setLevel(logging.ERROR)
 
     if _overrideLogLevel:
+        # for existing loggers that won't have getLogger do this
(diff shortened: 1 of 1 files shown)


ocrd_utils/ocrd_utils/logging.py
```diff
@@ -141,6 +141,11 @@ def initLogging():
         logging.getLogger('tensorflow').setLevel(logging.ERROR)
 
     if _overrideLogLevel:
```
bertsky, line 143:
```suggestion
    if _overrideLogLevel:
        # for existing loggers that won't have getLogger do this
        # (because they were instantiated through logging.getLogger
        #  instead of ours, or come from logging.config.fileConfig),
        # unset log levels so the global override can apply:
```

bertsky (approved):
Understood! Better document this though.

PR #635, opened 2020-10-23
ocrd.processor.base: add property zip_input_files

bertsky:
AFAICS this would already be useful for:
- [ocrd-cor-asv-ann-evaluate](https://github.com/ASVLeipzig/cor-asv-ann/blob/master/ocrd_cor_asv_ann/wrapper/evaluate.py) (where I drafted it)
- [ocrd-cis-align](https://github.com/cisocrgroup/ocrd_cis/blob/master/ocrd_cis/align/cli.py) (original implementation)
- [ocrd-dinglehopper](https://github.com/qurator-spk/dinglehopper/blob/master/qurator/dinglehopper/ocrd_cli.py)
- [ocrd-segment-evaluate](https://github.com/OCR-D/ocrd_segment/blob/master/ocrd_segment/evaluate.py)
- [ocrd-segment-from-masks](https://github.com/OCR-D/ocrd_segment/blob/master/ocrd_segment/import_image_segmentation.py)
- [ocrd-segment-replace-page](https://github.com/OCR-D/ocrd_segment/blob/maskrcnn-cli/ocrd_segment/replace_page.py)

Most of these recently broke with the `ocrd_mets.find_files` generator change.

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -6,7 +6,7 @@ __all__ = ['Processor', 'generate_processor_help', 'run_cli', 'run_processor']
 
 import os
 import json
-from ocrd_utils import VERSION as OCRD_VERSION, MIMETYPE_PAGE
+from ocrd_utils import VERSION as OCRD_VERSION, MIMETYPE_PAGE, getLogger
 from ocrd_validators import ParameterValidator
 from ocrd_models.ocrd_page import MetadataItemType, LabelType, LabelsType
 
@@ -15,9 +15,11 @@ from .helpers import run_cli, run_processor, generate_processor_help # pylint: d
 
 class Processor():
     """
-    A processor runs an algorithm based on the workspace, the mets.xml in the
-    workspace (and the input files defined therein) as well as optional
-    parameter.
+    A processor is an OCR-D compliant command-line-interface for executing
+    a single workflow step on the workspace (represented by local METS). It
+    reads input files for all or requested physical pages of the input fileGrp(s),
+    and writes output files for them into the output fileGrp(s). It may take 
+    a number of optional or mandatory parameters.
     """
 
     def __init__(
@@ -123,3 +125,48 @@ class Processor():
                 self.input_file_grp
                 ))
         return ret
+
+    @property
+    def zip_input_files(self, require_first=True):
+        """
+        List tuples of input files (for multiple input file groups).
+
+        Processors that expect/need multiple input file groups,
+        cannot use ``input_files``. They must align (zip) input files
+        across pages. This includes the case where not all pages
+        are equally present in all file groups.
+
+        This function does not make much sense for non-PAGE fileGrps,
+        so it uses a fixed MIME type filter for PAGE-XML.
+
+        Args:
+             require_first (bool): If true, then skip a page entirely
+             whenever it is not available in the first input fileGrp.
+        """
+
+        LOG = getLogger('ocrd.processor.base')
+        ifgs = self.input_file_grp.split(",")
+        # Iterating over all files repeatedly may seem inefficient at first sight,
+        # but the unnecessary OcrdFile instantiations for posterior fileGrp filtering
+        # can actually be much more costly than traversing the ltree.
+        # This might depend on the number of pages vs number of fileGrps.
+
+        pages = dict()
+        for i, ifg in enumerate(ifgs):
+            for file_ in self.workspace.mets.find_all_files(
+                    pageId=self.page_id, fileGrp=ifg, mimetype=MIMETYPE_PAGE):
+                if not file_.pageId:
+                    continue
+                LOG.debug("adding page %s to input file group %s", file_.pageId, ifg)
+                ift = pages.setdefault(file_.pageId, [None]*len(ifgs))
+                ift[i] = file_
+        ifts = list()
+        for page, ifiles in pages.items():
+            for i, ifg in enumerate(ifgs):
+                if not ifiles[i]:
+                    # other fallback options?
+                    LOG.error('found no page %s in file group %s',
+                              page, ifg)
+            if ifiles[0] or not require_first:
+                ifts.append(tuple(ifiles))
+        return ifts
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -126,7 +126,7 @@ class Processor():
                 ))
         return ret
 
-    def zip_input_files(self, require_first=True):
+    def zip_input_files(self, require_first=True, mimetype=MIMETYPE_PAGE):
         """
         List tuples of input files (for multiple input file groups).
 
@@ -153,7 +153,7 @@ class Processor():
         pages = dict()
         for i, ifg in enumerate(ifgs):
             for file_ in self.workspace.mets.find_all_files(
-                    pageId=self.page_id, fileGrp=ifg, mimetype=MIMETYPE_PAGE):
+                    pageId=self.page_id, fileGrp=ifg, mimetype=mimetype):
                 if not file_.pageId:
                     continue
                 LOG.debug("adding page %s to input file group %s", file_.pageId, ifg)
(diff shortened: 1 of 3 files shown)


ocrd/ocrd/processor/base.py
```diff
@@ -15,9 +15,11 @@
 
 class Processor():
     """
-    A processor runs an algorithm based on the workspace, the mets.xml in the
-    workspace (and the input files defined therein) as well as optional
-    parameter.
+    A processor is an OCR-D compliant command-line-interface for executing
```
kba, line 18: @EEngl52 @cneud perhaps update our glossary entry on *processor* https://ocr-d.de/en/spec/glossary#processor?

kba (approved):
Nice, thanks. I'll test it some more and merge.

ocrd/ocrd/processor/base.py
```diff
@@ -123,3 +125,48 @@ def input_files(self):
                 self.input_file_grp
                 ))
         return ret
+
+    @property
```
kba, line 129:
```suggestion

```

If it is a property, it shouldn't have arguments.

ocrd/ocrd/processor/base.py
```diff
@@ -123,3 +125,47 @@ def input_files(self):
...
+    def zip_input_files(self, require_first=True):
+        """
+        List tuples of input files (for multiple input file groups).
+
+        Processors that expect/need multiple input file groups,
+        cannot use ``input_files``. They must align (zip) input files
+        across pages. This includes the case where not all pages
+        are equally present in all file groups.
+
+        This function does not make much sense for non-PAGE fileGrps,
```
kba, line 138: It does make sense for e.g. `dinglehopper` to have ALTO/PAGE, though - @mikegerber?

ocrd/ocrd/processor/base.py
bertsky: Oh, sorry, forgot to remove that when I added `require_first`.

ocrd/ocrd/processor/base.py
bertsky: Yes, sorry, using ALTO input as if it were PAGE input does make sense. That's already true for processors like `ocrd-dinglehopper` or `ocrd-fileformat-transform`. Also, even image files may be needed as input, e.g. in `ocrd-segment-from-masks`. 

So IMO we should not filter by MIME type by default, and perhaps just offer this as a kwarg (for efficiency where filtering is in fact necessary).

I wonder whether we should invest extra effort to detect when multiple files match for a single pageId, though. (Currently, the dict item just gets silently overwritten by the last match.)

kba: > multiple files match for a single pageId

You mean multiple PAGE/ALTO files in the same file group? I think that should not happen (TM) but maybe I'm overlooking something.

> just offer this as a kwarg

I'll add that right away.

ocrd/ocrd/processor/base.py
EEngl52: For me this definition is way more understandable than the one currently on our website

ocrd/ocrd/processor/base.py
mikegerber: @kba Yes, for example I might get ALTO from a pre-existing ABBYY fulltext I want to compare to some GT I made in PAGE.

kba: You can pass a `mimetype` kwarg to `self.zip_input_files()`, e.g. to get PAGE-XML or ALTO files:

```python
self.zip_input_files(mimetype=r'//application/(vnd.prima.page|alto)\+xml')
```

ocrd/ocrd/processor/base.py
kba: https://github.com/OCR-D/ocrd-website/pull/183

ocrd/ocrd/processor/base.py
bertsky: > > multiple files match for a single pageId
> 
> You mean multiple PAGE/ALTO files in the same file group? I think that should not happen (TM) but maybe I'm overlooking something.

No, I mean _image files_ for the same pageId (as in `Processor.input_files`). That was the reason for having a MIME filter in the first place here: we can now have PAGE and derived images mixed in the fileGrp. If image files are not expected at all, this can be easily avoided. But if (original) image files are in fact a valid case (as in `ocrd-segment-from-masks`), then we again need to have precautions against multiple (derived image) results.

kba: > then we again need to have precautions against multiple (derived image) results

So, this then boils down to the same problem as in https://github.com/OCR-D/core/pull/611#discussion_r492288287 - how to detect the "main" image representing the page in this file group, right? How about this:

1. Search for images in the filegroup. If only 1 image: use this one.
2. Take the candidates, look up the PAGE-XML for this file group, look for candidates matching `AlternativeImage/filename`, use first one

bertsky: Let's step back a little: `zip_input_files`, like `input_files` is a very general function that could be used for many different scenarios. This may include cases where the caller knows a-priori which MIME types are correct (ALTO, PAGE, image), or the special case where the caller wants to ensure PAGE (creating an empty annotation if only an image is present). But this is (strictly, at least from the point of the implementation) independent of whether pageIds repeat and how to deal with that (ignore/select or error). There we have the special case where there can be multiple (derived) images _and_ 1 PAGE, or _only_ 1 (original) image. Everything else should probably always be an error (which at some cost we should try to detect).

> So, this then boils down to the same problem as in [#611 (comment)](https://github.com/OCR-D/core/pull/611#discussion_r492288287) - how to detect the "main" image representing the page in this file group, right?

No, I would not say so. We don't necessarily need to find out the original image (possibly in another fileGrp) here. I suggest the following:

1. apply MIME filter (default: none)
1. iterate resulting _pages_ (sets of files)
1. if multiple files per page:
   1. if no MIME filter:
      1. if PAGE is present: take this, ignore others
      1. else: show error, skip (for this fileGrp)
   1. else: show error, skip (for this fileGrp)
1. else: take this

So 3.i.a. is the derived images case (but extends to non-images here, too). Where I said `skip` I could also have said `select-first` or `abort` – we'll have to decide. The 2 error messages are distinct, though.

bertsky: I have updated the PR according to the above proposal. I decided to support different kinds of error handling (`skip`, `first`, `last` and `abort`). I also added the special case multiple PAGE matches (always abort). Plus I have extended the tests.

bertsky: @kba if you don't mind, I'll make an analogous change to `input_files` here, following [my similar proposal](https://github.com/OCR-D/core/pull/624#pullrequestreview-505644915) in #624 which would become obsolete then.

bertsky: > if you don't mind, I'll make an analogous change to `input_files` here, following [my similar proposal](https://github.com/OCR-D/core/pull/624#pullrequestreview-505644915) in #624 which would become obsolete then.

Maybe it would be prudent to simply let `Processor.input_files` delegate to the new, more general `Processor.zip_input_files(mimetype=None, on_error='abort')`, simply reducing the resulting single-tuples to atoms?

kba: > @kba if you don't mind, I'll make an analogous change to input_files here, following my similar proposal in #624 which would become obsolete then.

I don't mind!

> Maybe it would be prudent to simply let Processor.input_files delegate to the new, more general Processor.zip_input_files(mimetype=None, on_error='abort'), simply reducing the resulting single-tuples to atoms?

Single-element-tuples are handled like single values? Then yes, sure, if we can simplify, DRY :+1:

bertsky: > Single-element-tuples are handled like single values?

If there's only one input fileGrp, and we don't use `skip`, then this should behave like `input_files` (after we "untuple").

bertsky: Done.

PR #636, opened 2020-10-27
Switch from fchmod to chmod to allow dev on win

b2m:
The method `os.fchmod` is only available on Unix system.

Therefore developing for `ocrd_utils` or any packages depending on it is not possible on Windows.

As of Python 3.3 `os.chmod` also accepts a file descriptor.

Switching to `os.chmod` has the risk of masking raise conditions that should be prohibited by `AtomicWriterPerms`, but only **on Windows**.
AFAIK running natively on Windows is not a goal of OCR-D (right now).


kba:
> AFAIK running natively on Windows is not a goal of OCR-D (right now).

It is not but if we can accomodate Windows users with reasonable effort (like this PR, thanks!) then we should.

kba (approved):

PR #640, opened 2020-11-10
Image API: refactor, re-crop and re-interprete

bertsky:
Here's the latest in our series: "this time I have gotten it right, I'm sure"! 

(Let's just hope we keep failing better.)

The first commit refactors both workspace functions, not only reducing code duplication between page and lower levels, but also helping with the second commit: 

We must __re-crop__ to the new bounding box after rotation! Until now we forgot that. Though we did so consistently between coordinates and images, under some circumstances inconsistency would nevertheless be unavoidable. Plus it helps getting tighter crops from rotated non-rectangular polygons.

The third commit is where things get complicated again: Until now we had a _wrong interpretation_ of what a missing `@orientation` means on the region level. When we extended PAGE with `@orientation` on the page level, no one clarified explicitly what semantics the two should have w.r.t. each other. Should the region **inherit** from the page level now, or still be interpreted as default to **zero**?

1. inherit: assume that when `TextRegion/@orientation` is missing, it defaults to the next higher level (could be another region or the page level).
   - con: requries support for distinction `None` vs `0` (which we do have in generateDS)
2. default to zero:
   - con: derived images need to be unrotated (i.e. if you want deskewed regions, either don't deskew on the page level, or also deskew all regions)
   - con: lower levels (without `@orientation`) must be different again (because otherwise no deskewing would ever apply to them)
3. "special zero" (current implementation): no distinction between missing `@orientation` and zero, _and_ zero interpreted as inherited
   - con: small angles around zero would behave totally different
   - con: if deskewing actually resulted in zero (as opposed to some other value on the page level), this could not be represented at all

Clearly, 1 is preferable, and is what this PR brings.

Since there are so many intricacies involved in _applying_ rotation and reflection alone (division into minimal rotation and reflection, interpolation and filling, increase of canvas, re-cropping), I recommend always using `workspace.image_from_*` to generate derived images when deskewing (as done in https://github.com/OCR-D/ocrd_tesserocr/pull/158 and https://github.com/cisocrgroup/ocrd_cis/pull/77).

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -187,7 +187,7 @@ class Processor():
                     continue
                 ift = pages.setdefault(file_.pageId, [None]*len(ifgs))
                 if ift[i]:
-                    LOG.debug("another page %s in input file group %s", file_.pageId, ifg)
+                    LOG.debug("another file %s for page %s in input file group %s", file_.ID, file_.pageId, ifg)
                     # fileGrp has multiple files for this page ID
                     if mimetype:
                         # filter was active, this must not happen
@@ -226,7 +226,7 @@ class Processor():
                         else:
                             raise Exception("Unknown 'on_error' strategy '%s'" % on_error)
                 else:
-                    LOG.debug("adding page %s to input file group %s", file_.pageId, ifg)
+                    LOG.debug("adding file %s for page %s to input file group %s", file_.ID, file_.pageId, ifg)
                     ift[i] = file_
         ifts = list()
         for page, ifiles in pages.items():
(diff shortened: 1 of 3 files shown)


kba (approved):

bertsky:
@kba I tail-gated b80f730 here. This fixes the realistic case where a METS contains mets:file entries without mets:FLocat (but only mets:FContent). We should probably have such an example in our workflow regression test (like kant_aufklaerung...complex). Mine was from Grenzboten...

kba:
> @kba I tail-gated [b80f730](https://github.com/OCR-D/core/commit/b80f730c2a09edd2553151bd21d42e8bfd6c0706) here. This fixes the realistic case where a METS contains mets:file entries without mets:FLocat (but only mets:FContent). We should probably have such an example in our workflow regression test (like kant_aufklaerung...complex). Mine was from Grenzboten...

Understood, thanks. I'll merge and release this later today.

PR #645, opened 2020-11-20
ocrd workspace: workspace.directory should be abspath, fix #644

### LINKED ISSUE #644: ocrd workspace find --download does not use correct workspace prefix
There is a flaw in the workspace directory resolution of the CLI:

    ocrd workspace -d Brsfded_39946221X-18920413 find -G ORIGINAL --download

This will:

- change remote links in `Brsfded_39946221X-18920413/mets.xml` to local ones, now pointing to `ORIGINAL/*`
- download the files under `Brsfded_39946221X-18920413/Brsfded_39946221X-18920413/ORIGINAL` (one level too deep)

kba:


bertsky (approved):
Absolutely!

Issue #607, opened 2020-09-17
ocrd.process.profile: log peak RSS besides CPU seconds


bertsky:
We should track the runtime memory requirements besides runtime duration itself.

And, if at all possible, the profile logger could also be fed with the image sizes (pixel resolution) of the pages. (Perhaps after `Workspace.image_from_page`'s `OcrdExif` instantiation?)

bertsky:
Moreover, the current time counts measure the wall time, not the CPU time. We should at least complement this figure. (Somewhat confusingly, the CPU time is called `time.clock` in python, whereas wall time is `time.time`.)

bertsky:
And time spent waiting for I/O!

PR #647, opened 2020-11-26
run_cli: measure both wall and CPU time, #607

kba:


bertsky (approved):
Splendid. As to peak RSS, I found [this Python lib](https://pypi.org/project/timemory/1.1.8rc0/), but it's probably not worth the dependency. We could also just parse output of GNU time here...

PR #648, opened 2020-11-26
prune_ReadingOrder: also check empty groups

bertsky:
This prevents invalid PAGE-XML from common sloppiness: processors which leave an empty group (e.g. segmentation that did not find regions).

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_page_generateds.py
+++ b/ocrd_models/ocrd_models/ocrd_page_generateds.py
@@ -2,8 +2,8 @@
 # -*- coding: utf-8 -*-
 
 #
-# Generated Sat Oct 31 00:59:08 2020 by generateDS.py version 2.35.20.
-# Python 3.6.7 (default, Oct 22 2018, 11:32:17)  [GCC 8.2.0]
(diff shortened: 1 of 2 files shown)


bertsky:
Do you want me to regenerate?

kba (approved):

kba:
Thanks, this is also useful because empty reading order groups are an XSD validation error.

> Do you want me to regenerate?

I can do that too, tomorrow.

bertsky:
> Thanks, this is also useful because empty reading order groups are an XSD validation error.

Yes, but this does not attack empty groups in general – only at the top level, so that `ReadingOrder` itself can be removed. This is enough to have validity in most cases. 

But the more processors do with reading order, the more can go wrong. So yes, we could try to traverse all groups recursively, and reduce each empty group to a mere `RegionRef(Indexed)` keeping everything else (the `@regionRef` and the chain).

kba:
> But the more processors do with reading order, the more can go wrong. So yes, we could try to traverse all groups recursively, and reduce each empty group to a mere `RegionRef(Indexed)` keeping everything else (the `@regionRef` and the chain).

Indeed. I want to make a new release of core and ocrd_all later today though, so I'll add that recursion in a separate PR.

PR #649, opened 2020-11-27
TaskSequence.parse: instantiate given class

bertsky:
I am doing a proof of concept for a workflow server – outside of core, just importing from it – and found this error: the class method should instantiate the passed class, not a fixed one, so subclasses can override properly.


kba (approved):
LGTM

> I am doing a proof of concept for a workflow server – outside of core, just importing from it

Awesome, looking forward to what you come up with!

PR #651, opened 2020-11-30
task_sequence: remove invalid refs to out/err

bertsky:
Since run_cli does not (capture and) return stdout and stderr anymore, these references are impossible and unnecessary.


kba (approved):
Thanks, that was an oversight.

PR #663, opened 2021-01-27
resmgr: No downloading on-demand

kba:


ocrd/ocrd/processor/base.py
```diff
@@ -154,25 +154,11 @@ def resolve_resource(self, val):
             val (string): resource value to resolve
         """
         executable = self.ocrd_tool['executable']
-        if exists(val):
-            return val
```
bertsky, line 158: Why'd you remove resolving absolute path names?

kba: :blush: accident. fixing.

kba: d5443c2 fixed

ocrd/ocrd/processor/base.py
```diff
@@ -153,26 +153,14 @@ def resolve_resource(self, val):
...
-            raise FileNotFoundError("Could not resolve %s resource '%s'" % (executable, val))
-        _, resdict = reslist[0]
-        return str(resmgr.download(
-            executable,
-            url=resdict['url'],
-            name=resdict['name'],
-            path_in_archive=resdict['path_in_archive'],
-            resource_type=resdict['type']
-        ))
+        raise FileNotFoundError("Could not find resource '%s' for executable '%s'. Try 'ocrd resmgr download %s %s' to download this resource." %
```
bertsky, line 162: If you feel an exception is better than logging an error plus exit 1 here, so be it. (Stack trace would be ugly, but maybe later we want to catch Processor's exceptions upwards in a certain way...)

bertsky (approved):

ocrd/ocrd/processor/base.py
kba: Nah, the stacktrace is just spammy, let's log and exit, no need to scare users for such expectable behavior.

kba: 5e2516d9

kba:
8abb08d remembers what `$PWD` was before the `chdir` call in the processor initialization, so models installed with `--location cwd` can be found by the processors' `self.resolve_resourcve`.

bertsky:
> [8abb08d](https://github.com/OCR-D/core/commit/8abb08d8c1c99d5c4bd335619bd28ab48ba82edb) remembers what `$PWD` was before the `chdir` call in the processor initialization, so models installed with `--location cwd` can be found by the processors' `self.resolve_resourcve`.

That was too fast for me. Is that really what we want `cwd` to mean? Until recently, relative paths were resolved w.r.t. the workspace directory. Therefore I assumed `cwd` would denote that, too.

But perhaps the old behaviour was bad anyway. After all, the user/caller does not necessarily _expect_ the chdir...

ocrd/ocrd/resource_manager.py
```diff
@@ -137,7 +137,7 @@ def default_resource_dir(self):
     def location_to_resource_dir(self, location):
         return '/usr/local/share/ocrd-resources' if location == 'system' else \
                 join(XDG_DATA_HOME, 'ocrd-resources') if location == 'data' else \
-                getcwd()
+                join(getcwd(), 'ocrd-resources')
```
bertsky, line 140: That is totally unexpected IMHO. CWD-relative path names should not get any prefix directory, neither `ocrd-resources` nor the processor name.

kba:
> But perhaps the old behaviour was bad anyway. After all, the user/caller does not necessarily expect the chdir...

Sry, I thought this was non-controversial. I constantly stumble over the fact that processors `chdir` to the workspace, I think that's counterintuitive and I didn't consider that you would call `ocrd resmgr download --location cwd` **inside** a workspace directory. For example, I am using that feature in ocrd_anybaseocr's CI, downloading the models to `/home/circleci/repo/ocrd-resources` and then call the processors from that directory but with full path to workspace `mets.xml` in `--mets`.

So I think this is the less bad solution but I will give you more time to react before mergin in the future.

ocrd/ocrd/resource_manager.py
bertsky: > With the 2.22.3 behavior, it consistently stores and looks in `$PWD/ocrd-resources>/<executable>/<resname>`.
> 
> And I honestly thought that's how it should be but indeed the specs say `$PWD/<resname>`.

And so I (strongly) feel they should :slightly_smiling_face: 

> We can fix this by `pushd_popd(self.old_pwd)` before we check whether `<resname>` can be resolved as-is. That would still allow you to drop resources flat into `$PWD`.

Whatever it takes!

> One reason I find `$PWD/ocrd-resources` useful is for caching in CI, because you don't need to list all the individual files like that.

I don't quite understand that TBH...

kba: > > With the 2.22.3 behavior, it consistently stores and looks in `$PWD/ocrd-resources>/<executable>/<resname>`.
> > And I honestly thought that's how it should be but indeed the specs say `$PWD/<resname>`.
> 
> And so I (strongly) feel they should 🙂
> 
> > We can fix this by `pushd_popd(self.old_pwd)` before we check whether `<resname>` can be resolved as-is. That would still allow you to drop resources flat into `$PWD`.
> 
> Whatever it takes!

OK, but that only solves half of the problem, the other question is whether `ocrd resmgr --location cwd` should download a resource `<res>` to `$PWD/<res>` or `$PWD/ocrd-resources/<res>`. IIUC you also want the download to the former, without `ocrd-resources`, correct?

> 
> > One reason I find `$PWD/ocrd-resources` useful is for caching in CI, because you don't need to list all the individual files like that.
> 
> I don't quite understand that TBH...

>> One reason I find $PWD/ocrd-resources useful is for caching in CI, because you don't need to list all the individual files like that.

> I don't quite understand that TBH...

We want to cache models across CI runs so we don't need to download gigabytes of model data every time. If all the models are in one directory `$PWD/ocrd-resources`, you can just cache that whole directory instead of listing explicitly every file in `$PWD` thart should be cached. It's a minor thing.

bertsky: > OK, but that only solves half of the problem, the other question is whether `ocrd resmgr --location cwd` should download a resource `<res>` to `$PWD/<res>` or `$PWD/ocrd-resources/<res>`. IIUC you also want the download to the former, without `ocrd-resources`, correct?

Yes.

> We want to cache models across CI runs so we don't need to download gigabytes of model data every time. If all the models are in one directory `$PWD/ocrd-resources`, you can just cache that whole directory instead of listing explicitly every file in `$PWD` thart should be cached. It's a minor thing.

Oh, I got it. But you could still have the entire CWD cached, if you placed the workspace elsewhere.

PR #664, opened 2021-01-28
workspace: fix docstrings

bertsky:
(according to https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html specs)

There's lots more to do in similar fashion, but `workspace` seemed like a good starting point.

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -1,5 +1,5 @@
 """
-Processor base class and helper functions
+Processor base class and helper functions.
 """
 
 __all__ = [
@@ -37,9 +37,10 @@ from .helpers import run_cli, run_processor, generate_processor_help # pylint: d
 
 class Processor():
     """
-    A processor is an OCR-D compliant command-line-interface for executing
-    a single workflow step on the workspace (represented by local METS). It
-    reads input files for all or requested physical pages of the input fileGrp(s),
+    A processor is a tool that implements the uniform OCR-D command-line interface
+    for run-time data processing. That is, it executes a single workflow step,
+    or a combination of workflow steps, on the workspace (represented by local METS).
+    It reads input files for all or requested physical pages of the input fileGrp(s),
     and writes output files for them into the output fileGrp(s). It may take 
     a number of optional or mandatory parameters.
     """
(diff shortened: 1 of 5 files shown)


kba:
Thanks for documenting and fixing the docstrings, much appreciated. CI failure is due to the changed resmgr `--location cwd` behavior.

I really need to clean up the API docs (https://ocr-d.de/core/) so people can appreciate those docstrings rendered.

bertsky:
> CI failure is due to the changed resmgr `--location cwd` behavior.

See https://github.com/OCR-D/core/pull/669/commits/c920fde9db32ebcb15184fc10bb8850d4299689e for a fix.

bertsky:
Sorry to say the generated apidoc does not always work with this formatting. For example, argument descriptions wrapped across multiple lines are broken. But [this](https://google.github.io/styleguide/pyguide.html) says:

> Args:
List each parameter by name. A description should follow the name, and be separated by a colon followed by either a space or newline. If the description is too long to fit on a single 80-character line, use a hanging indent of 2 or 4 spaces more than the parameter name (be consistent with the rest of the docstrings in the file). 

I have tried both, but this does not in fact work. The first word of the follow-up line gets picked up as a type annotation like so:

<dl class="field-list simple">
<dt class="field-odd">Parameters</dt>
<dd class="field-odd"><ul class="simple">
<li><p><strong>require_first</strong> (<em>bool</em>) – If true, then skip a page entirely</p></li>
<li><p><strong>it is not available in the first input fileGrp.</strong> (<em>whenever</em>) – </p></li>
<li><p><strong>mimetype</strong> (<em>str</em>) – If not None, filter by the specified MIME</p></li>
<li><p><strong>(</strong><strong>literal</strong><strong> or </strong><strong>regex prefixed by //.</strong> (<em>type</em>) – </p></li>
<li><p><strong>prefer PAGE</strong><strong> or </strong><strong>image.</strong> (<em>Otherwise</em>) – </p></li>
</ul>
</dd>
</dl>

instead of

<dl class="field-list simple">
<dt class="field-odd">Parameters</dt>
<dd class="field-odd"><ul class="simple">
<li><p><strong>require_first</strong> (<em>bool</em>) – If true, then skip a page entirely whenever it is not available in the first input <tt>fileGrp</tt>.</p></li>
<li><p><strong>mimetype</strong> (<em>str</em>) – If not <tt>None</tt>, filter by the specified MIME type (literal or regex prefixed by <tt>//.</tt>) Otherwise prefer PAGE or image.</p></li>
</ul>
</dd>
</dl>

(I also tried adding `napoleon_google_docstring = True` to `docs/conf.py`, but that does not change anything either.)

bertsky:
What's even more strange, cross-references appear to be randomly working or not working. For example, in `ocrd.workspace.Workspace`, the same string
```
:py:class:`ocrd_models.ocrd_file.OcrdFile`
```

renders as `span class="pre"` in `download_file` but as correct `a href=...`  in `download_url`!

PR #671, opened 2021-02-16
resmgr: --location cwd means cwd, no subdirs, #663

kba:
`ocrd resmgr --location cwd download ...` will now download resource `<res>` to `$PWD/<res>`, not `$PWD/ocrd-resources/<executable>/<res>` as per the spec.

One minor drawback is that `ocrd resmgr list-installed` will no longer pick up resources in `$PWD` because there would be too many false positives, files that are not resources.

### DIFF
--- a/ocrd/ocrd/cli/resmgr.py
+++ b/ocrd/ocrd/cli/resmgr.py
@@ -1,6 +1,4 @@
 import sys
-from os import getcwd
-from os.path import join
 from pathlib import Path
 from distutils.spawn import find_executable as which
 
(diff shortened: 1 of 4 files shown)


bertsky (approved):

bertsky:
> One minor drawback is that `ocrd resmgr list-installed` will no longer pick up resources in `$PWD` because there would be too many false positives, files that are not resources.

Indeed. (One could argue though, that _downloading_ to CWD is not really _installing_ in the true sense. You'll have to move/drag them along to whereever you go to when you start processing.)

PR #672, opened 2021-02-18
implement value range with .. operator for pageId

kba:
When processing works with a non-trivial amount of pages, or when a long process fails in the middle of a run, it would be useful to not only allow comma-separated page ID values but ranges of page IDs.

This PR implements this by introducing the syntax `--page-id <startid>..<endid>` which

* extracts the first contiguous match of numbers of both `<startid>` and `<endid>`
* counts from start to end number and replaces the number in <startid> with the zero-padded current number

E.g. you can do `ocrd-processor -g PHYS_0001..PHYS_0003` is equivalent to `ocrd-processor -g PHYS_0001,PHYS_0002,PHYS_0003` but shorter.

### DIFF
--- a/ocrd_utils/ocrd_utils/str.py
+++ b/ocrd_utils/ocrd_utils/str.py
@@ -5,7 +5,6 @@ Utility functions for strings, paths and URL.
 import re
 import json
 from .constants import REGEX_FILE_ID
-from .logging import getLogger
 
 __all__ = [
     'assert_file_grp_cardinality',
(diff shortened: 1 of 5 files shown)


bertsky (approved):
Very nice!

(Although I would personally find it more intuitive if 01..05 included 05. Most users will probably not be programmers. And even if they are, they might expect semantics as in shell brace expansion.)

kba:
> Although I would personally find it more intuitive if 01..05 included 05

That was my intention ("Generate a list of strings by incrementing the number part of ``start`` until including ``end``.") but I got wrong how `range` works myself :) fixing that.

kba:
Should we also support this for `--file-id`?

bertsky:
> Should we also support this for `--file-id`?

Not so sure. That's not part of the processor API, and you can still use the regex syntax...

BTW, the pageId range expression must go to the spec, too.

ocrd_utils/ocrd_utils/str.py
```diff
@@ -177,3 +178,17 @@ def safe_filename(url):
...
     return ret
+
+def generate_range(start, end):
+    """
+    Generate a list of strings by incrementing the number part of ``start`` until including ``end``.
+    """
+    ret = []
+    start_num, end_num = re.search('\d+', start), re.search('\d+', end)
+    if not (start_num and end_num):
+        getLogger('ocrd_utils.generate_range').error("Unable to generate range %s .. %s, could not detect number part" % (start, end))
```
bertsky, line 189: Just a thought: Maybe this should even be an exception. After all, it's likely the entire result is not what was intended afterwards. Better safe than sorry.

kba: Fine with me.

Issue #670, opened 2021-02-16
ocrd workspace merge


kba:
If users chose to parallelize processing by partitioning pages, they will end up with multiple `mets.xml` that need to be merged.

We should add a CLI `ocrd workspace -m mets1.xml merge mets2.xml mets3.xml ...` that iterates all the files in all the mets files passed to it and add them to the workspace.

bertsky:
Maybe also: If processing happened on a clone of the workspace (e.g. on faster storage or another host), it should be possible to merge new fileGrps.

kba:
> Maybe also: If processing happened on a clone of the workspace (e.g. on faster storage or another host), it should be possible to merge new fileGrps.

https://github.com/OCR-D/core/pull/582 would come in handy for merging, which could be just a subset of the clone functionality (with the clone being a local workspace not a remote METS). It's not in a mergeable state right now, so I'll implement the fileGrp filtering/mapping here and maybe later use the `OcrdMetsFilter` from #582

bertsky:
Oh, I didn't realize this had been solved already, even in the CLI. Would be wonderful to see this documented in the wiki or user guide, too. We had two likely scenarios:

1. splitting a large workspace (by cloning with pageId filter), processing in parallel and then merging again
2. cloning a workspace to fast storage, processing there, then merging into the original location

PR #673, opened 2021-02-19
Merging METS/workspaces

kba:
Implements #670


bertsky (approved):
Awesome!

(I assume you'll add some form of fileGrp filter and pageId filter later-on?)

kba:
> (I assume you'll add some form of fileGrp filter and pageId filter later-on?)

It's not exposed to the CLI yet but `OcrdMets.merge` passes all kwargs on to `OcrdMets.find_files`, so it is already possible in the API. Commit with that is upcoming.

PR #652, opened 2020-12-01
Workflow server

bertsky:
Implementation of the [workflow server](https://hackmd.io/23-JzLp_Q96cb6T0ttoFIA).

    ocrd --help

```
Commands:
  bashlib    Work with bash library
  log        Logging
  ocrd-tool  Work with ocrd-tool.json JSON_FILE
  validate   All the validation in one CLI
  process    Run processor CLIs in a series of tasks
  workflow   Process a series of tasks
  workspace  Working with workspace
  zip        Bag/Spill/Validate OCRD-ZIP bags
```

    ocrd workflow --help

```
Usage: ocrd workflow [OPTIONS] COMMAND [ARGS]...

  Process a series of tasks

Options:
  --help  Show this message and exit.

Commands:
  process  Run processor CLIs in a series of tasks alias for ``ocrd...
  server   Start server for a series of tasks to run processor CLIs or APIs...
```

    ocrd workflow server --help

```
Usage: ocrd workflow server [OPTIONS] TASKS...

  Start server for a series of tasks to run processor CLIs or APIs on
  workspaces

  Parse the given tasks and try to instantiate all Pythonic processors among
  them with the given parameters. Open a web server that listens on the
  given host and port for GET requests named ``process`` with the following
  (URL-encoded) arguments:

      mets (string): Path name (relative to the server's CWD,
      or absolute) of the workspace to process

      page_id (string): Comma-separated list of page IDs to process

      overwrite (bool): Remove output pages/images if they already exist

  The server will handle each request by running the tasks on the given
  workspace. Pythonic processors will be run via API (on those same
  instances).  Non-Pythonic processors (or those not directly accessible in
  the current venv) will be run via CLI normally, instantiating each time.
  Also, between each contiguous chain of Pythonic tasks in the overall
  series, no METS de/serialization will be performed.

  Stop the server by sending SIGINT (e.g. via ctrl+c on the terminal), or
  sending a GET request named ``shutdown``.

Options:
  -l, --log-level [OFF|ERROR|WARN|INFO|DEBUG|TRACE]
                                  Log level
  -h, --host TEXT                 host name/IP to listen at
  -p, --port INTEGER RANGE        TCP port to listen at
  --help                          Show this message and exit.
```

Example:

[...]

Issue #376, opened 2019-12-19
Additional parameter for custom resolution


M3ssman:
Hello,

please add DPI-Parameters to enable to enforce custom resolution when using tesseract

## Tesseract CLI

`--dpi 470`

## ocrd-tesserocr

`dpi: 470`

bertsky:
> Besides the manual DPI override, this would also allow supporting DPI meta-data validation with different levels of strictness.
> 
> Or supporting automatic workspace validation with different levels/sets of checks.
> 
> Or supporting processing with `--force/--overwrite`.
> 
> Or supporting processing on multiple CPUs/GPUs with given scalefactor.
> 
> Or supporting time constraints on different hierarchy levels.
> 
> Just saying!

Or enabling/disabling METS caching (or even METS server).

PR #678, opened 2021-03-04
Profile mem

kba:
When running processors with `ocrd process`, we already measure the wall and CPU time of a processor run. This adds basic tracking of memory usage with https://github.com/pythonprofilers/memory_profiler.

Currently, this just outputs a sparkline of memory usage and max/min memory, e.g.

```
16:35:54.593 INFO ocrd.process.profile - memory consumption: ▁▁▂▅▅▃▄▄▄▄▄▄▄▄▄▄▅▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▅▅▅▅▅▅▅▅▅▅▆▆▆▆▆▆▆▆▆▆▆▆▆▆▆▆▆██▅▅▅▅▅▅▆▆▆▆▆▆▆▆▆▆▆▆▆▆▆▆▆▆▅▅▅▅▆▆▆▆▆▆▆▆▆▆▆▆▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▆▄▄
max: 300.79 MiB min: 133.09 MiB
```


ocrd/ocrd/processor/helpers.py
```diff
@@ -69,7 +71,13 @@ def run_processor(
     log.debug("Processor instance %s (%s doing %s)", processor, name, otherrole)
     t0_wall = perf_counter()
     t0_cpu = process_time()
-    processor.process()
+    mem_usage = memory_usage(proc=processor.process, interval=.1, timeout=None, timestamps=True)
```
bertsky, line 74: Maybe you wanna add `multiprocess=True` in there for processors that use `multiprocessing` or similar, or that delegate to shell calls. Perhaps also `include_children=True` (not sure about the meaning though – does the sum then count shared/CoW pages repeatedly?).

bertsky (approved):
Excellent!

Not sure how to add that, but `memory_profiler` is based on `psutil`, which would also [give](https://psutil.readthedocs.io/en/latest/) us `.cpu_times()['iowait']` or `.disk_io_counters()['read_time']` / `write_time` / `busy_time` etc...

bertsky:
This will conflict with #652 (where that part of `run_processor` is delegated to the new `run_api`) – not sure what's the least painful way for merging.

Oh, since I saw this also uses `multiprocessing`, have you measured any CPU overhead for the memory profiling yet?

kba:
> This will conflict with #652 (where that part of run_processor is delegated to the new run_api) – not sure what's the least painful way for merging.

Is that just an issue of resolving conflicts or do you mean there's an incompatible API-wise now?

> Oh, since I saw this also uses multiprocessing, have you measured any CPU overhead for the memory profiling yet?

I haven't, yet. I did notice that when I profile the current process (`proc=-1`) instead of the `processor.process()` call only, then either it just offers a single frame of measurement (`max_iter=1`) or will hang until the timeout is reached (`timeout=100`)

ocrd/ocrd/processor/helpers.py
bertsky: > > Perhaps also `include_children=True` (not sure about the meaning though – does the sum then count shared/CoW pages repeatedly?).
> 
> ```python
>             if include_children:
>                 mem +=  sum(_get_child_memory(process, meminfo_attr))
> ```

I saw that, but the question remains: what is in that sum? (We probably don't want to count shared and CoW pages repeatedly. But it may be hard to calculate exactly. IIRC standard `ps` and `top` and even `time` do count repetitions, so they appear to large (don't "add up") for multiprocessed programs, whereas `smem` gets it right. [This](https://unix.stackexchange.com/questions/34795/correctly-determining-memory-usage-in-linux) tells me we are actually interested in the PSS, the proportional set size, rather than the RSS.)

bertsky:
> > This will conflict with #652 (where that part of run_processor is delegated to the new run_api) – not sure what's the least painful way for merging.
> 
> Is that just an issue of resolving conflicts or do you mean there's an incompatible API-wise now?

It _looks_ like this is only cosmetic (since the workflow server does not handle the retval either, it just adds exception handling, and does not use multiprocessing itself, at least in the current implementation). But I would have to check/test.

ocrd/ocrd/processor/helpers.py
```diff
@@ -69,7 +71,12 @@ def run_processor(
     log.debug("Processor instance %s (%s doing %s)", processor, name, otherrole)
     t0_wall = perf_counter()
     t0_cpu = process_time()
-    processor.process()
+    mem_usage = memory_usage(proc=processor.process, interval=.1, timeout=None, timestamps=True, multiprocess=True, include_children=True)
```
bertsky, line 74:
```suggestion
    mem_usage = memory_usage(proc=processor.process, 
                            interval=.1, timeout=None, timestamps=True, 
                            # include sub-processes
                            multiprocess=True, include_children=True, 
                            # get proportional set size instead of RSS
                            backend='psutil_pss')
```

bertsky:
I wonder if it would be much effort to incorporate [gputil](https://github.com/anderskm/gputil) for (avg/peak) GPU memory usage statistics. (Or perhaps one could add it to memory_profiler as an alternative `backend`?)

ocrd/ocrd/processor/helpers.py
bertsky: Ok, but PSS vs RSS is a more general problem/discussion. Any multiprocessing coder will know about that. And RSS is still a valuable information. So let's stick to your current implementation.

ocrd/ocrd/processor/helpers.py
bertsky: If it is really that simple – let's make this configurable (PSS vs RSS measuring). Did you try this?

bertsky:
> I wonder if it would be much effort to incorporate [gputil](https://github.com/anderskm/gputil) for (avg/peak) GPU memory usage statistics. (Or perhaps one could add it to memory_profiler as an alternative `backend`?)

I'm afraid it's not that simple at all. GPutil only gives snapshots, no aggregation/statistics there. And it looks like one would need to use the respective ML framework's facilities for process/session measurements anyway. (See [here](https://discuss.pytorch.org/t/measuring-peak-memory-usage-tracemalloc-for-pytorch/34067) and [there](https://discuss.pytorch.org/t/high-gpu-memory-usage-problem/34694) for Pytorch. TF now also has things like [get_memory_info](https://www.tensorflow.org/api_docs/python/tf/config/experimental/get_memory_info) / [get_memory_usage](https://www.tensorflow.org/api_docs/python/tf/config/experimental/get_memory_usage) / [reset_memory_stats](https://www.tensorflow.org/api_docs/python/tf/config/experimental/reset_memory_stats).)

Let's treat that off-topic for now!

ocrd/ocrd/processor/helpers.py
kba: Sure, however we first need a mechanism to toggle this behavior. Simple idea would be to add CLI flags `--mem-profile-rss` and `--mem-profile-pss`.

@bertsky prefers a way to have this configurable via the logging configuration - @M3ssman do you have a good idea how that might work? That way, we could keep the CLI arguments small and the `--help` output less confusing.

bertsky: Perhaps for the moment it could be as simple as 
```python
mem_prof = os.getenv('OCRD_PROFILE', '')
if mem_prof:
    ...
        backend='psutil_pss' if mem_prof == 'PSS' else 'psutil')
```

kba: I like environment variables in general, but when the choice is between environment variables and CLI flags, I prefer the latter, because at least it's consistent. Because otherwise you could get `OCRD_PROFILE=PSS ocrd-dummy --profile-file my.prof ...`.

M3ssman: @bertsky What-how do you mean by configurable via logging-conf?
 
But wait, looks like ocrd might lack a general configuration ini-file, doesn't it? 
I'm using config-file nowadays virtually in every python (client-)application, so thinking in terms of ocrd-as-framework (please don't mess up with the discussion ocrd(-core) as library vs. framework). Even standard python configuration API works pretty well, only have to work around if you want the ConfigParser also evaluate lists and dicts. 
Actually, with the latest python apps I tend to use CLI flags only to specify the path to the config-file :slightly_smiling_face:

M3ssman: @bertsky What-how do you mean by configurable via logging-conf?
 
But wait, looks like ocrd might lack a general configuration ini-file, doesn't it? 
I'm using config-file nowadays virtually in every python (client-)application, so thinking in terms of ocrd-as-framework (please don't mess up with the discussion ocrd(-core) as library vs. framework). Even standard python configuration API works pretty well, only have to work around if you want the ConfigParser also evaluate lists and dicts. 
Actually, with the latest python apps I tend to use CLI flags only to specify the path to the config-file :slightly_smiling_face:

kba: We tried to avoid a config file for the longest time to eliminate sources of trouble, or rather, force us to keep the interfaces as simple as possible and have sensible defaults instead of making everything configurable.

That being said, we do have `$HOME/.local/share/resource.yml` which is a kind of config file and we have the configuration file for the logging.

But if the CLI become to convoluted with lots of flags and there is consensus, I won't veto a configuration file. Syntax-wise, I'm a big fan of YAML combined with JSON-Schema but ini, TOML, dotenv etc. all are fine with me.

bertsky: @kba I agree, however I did not argue env var _plus_ cli arg, but only the former. Perhaps we could also move `--profile` and `--profile-file` into env var control entirely.

Or we go the other direction and put everything into a single CLI arg, just with multi-faceted value: `--profile mem=rss,cpu,file=foo.log` for example.

For the broader discussion of how to configure processors without adding to many options / parameters and what for, see [this old thread](https://github.com/OCR-D/core/issues/376#issuecomment-562349028)

ocrd/ocrd/processor/helpers.py
```diff
@@ -85,7 +87,12 @@ def run_processor(
...
```
kba, line 90: This causes a strange test failure in test_decorators, running the `processor.process` method 7 times instead of once.

```suggestion
    mem_usage = memory_usage(proc=processor.process, interval=.1, timeout=None, timestamps=True, multiprocess=True, include_children=True, max_iterations=1)
```

Seems to fix it but I am not sure I 100% understand how the function works.

ocrd/ocrd/processor/helpers.py
kba: OK, now that the wonky multiple-process-call behavior is fixed, I'll implement this via environment variable `OCRD_PROFILE` and `OCRD_PROFILE_FILE`

* `OCRD_PROFILE` not set or empty: no profiling
* `OCRD_PROFILE_FILE` set to a value (== current `--profile-file` flag):
  * if `OCRD_PROFILE` is defined but empty: no profiling
  * if `OCRD_PROFILE` is not defined: set `OCRD_PROFILE` to `CPU`
  * if `OCRD_PROFILE` is defined but does not contain `CPU`: raise error
* `OCRD_PROFILE` contains `CPU`, do CPU profiling (== current `--profile` flag)
* `OCRD_PROFILE` contains `RSS` do RSS memory profiling
* `OCRD_PROFILE` contains `PSS`: do proportionate mem profiling
* If `OCRD_PROFILE` contains both `PSS` and `RSS`: raise error

Did I catch everything?

kba: Implemented in 4fdfbf39b

MehmedGIT: Environment variables look good to me. 

In the readme, I would also suggest to reference an external link where the difference between `RSS` and `PSS` is explained or at least briefly described. May not be the best suggestion, but consider [this](https://elinux.org/Android_Memory_Usage).

bertsky: yes, sounds reasonable and complete AFAICS –  thx!

README.md
```diff
@@ -81,6 +82,19 @@ supported flags, options and arguments.
...
+
+Some parts of the software are configured via environement variables:
+
+* `OCRD_METS_CACHING`: If set to `true`, access to the METS file is cached, speeding in-memory search and modification.
+* `OCRD_PROFILE`: This variable configures the built-in CPU and memory profiling. If empty, no profiling is done. Otherwise expected to contain any of the following tokens:
+  * `CPU`: Enable CPU profiling of processor runs
+  * `RSS`: Enable RSS memory profiling
+  * `PSS`: Enable proportionate memory profiling
+* `OCRD_PROFILE_FILE`: If set, then the CPU profile is written to this file for later peruse with a analysis tools like [snakeviz](https://jiffyclub.github.io/snakeviz/)
+
```
bertsky, line 97:
```suggestion

Logging is configured via configuration files, see [module logging](./ocrd_utils#ocr-d-module-logging).
```

bertsky (approved):

PR #683, opened 2021-03-17
ocrd_page: id for Page (imageFilename) and PcGts (pcGtsId), fix #682

### LINKED ISSUE #682: ocrd_page: provide id attribute on page level
We have already amended `__hash__` for the entire segment hierarchy:

https://github.com/OCR-D/core/blob/1aabf82730610c091be1fd8d95f461a0e238c809/ocrd_models/ocrd_page_user_methods/__hash__.py#L1-L10

It would be nice if one could do a simple `.id` on `PcGtsType` and `PageType` as well. So the `__hash__` would then only delegate to that new `hash(self.id)`.

kba:


ocrd_models/ocrd_page_user_methods/id.py
```diff
@@ -0,0 +1,5 @@
+@property
+def id(self):
+    if hasattr(self, 'pcGtsId'):
+        return self.pcGtsId
```
bertsky, line 4: That can still fail, though: If no `@pcGtsId` is annotated, then this would return `None`, which is not hashable.

```suggestion
        return self.pcGtsId or ''
```

bertsky (approved):
Splendid!

PR #686, opened 2021-04-23
Fix image features last vs max

### LINKED ISSUE #684: image API: retrieve maximum feature set when underspecified
In the current implementation of `Workspace.image_from_page` and `Workspace.image_from_segment`, unless the combination of requested `feature_filter` and `feature_selector` arguments is already unique (or unsatisfiable), the **last** derived image is chosen (assuming this would be the most relevant for linear workflows).

However, that choice is not optimal for practical workflows. Since there is no contract with processors producing derived images to always base these on the last image or maximal feature set available, rather to the contrary, feature filters may yield combinations of a non-maximal feature set with the last operation, a feature might not "survive" the chain and might have to be re-retrieved by special processors.

For example, a cropper which needs `deskewed` (and therefore runs after `binarized`) might want to use the raw (instead of bitonal) image for its computation and annotation. Its resulting image would be added with `deskewed,cropped` but not `binarized`. Follow-up processors therefore won't get to "see" the bitonal image anymore.

If core always yields the maximal feature set (i.e. crop and deskew on the fly but keep `binarized` and other features it cannot recreate), then consumers will always get what they can.

bertsky:
Fixes #684 and ensures that masked areas do not influence the median for `fill='background'` estimation. (This is relevant for example during cropping if the area outside the polygon is mostly black in the binarized image, which then makes the median black instead of white, too, so the non-orthogonal margins become black instead of white.)


kba (approved):

kba:
076ef5c I've added a unit test to verify the changed behavior. The test data is bogus but enough to trigger the changed behavior afaict.

bertsky:
> [076ef5c](https://github.com/OCR-D/core/commit/076ef5c3b41590a80db4ca32c99cabb9fc56d640) I've added a unit test to verify the changed behavior. The test data is bogus but enough to trigger the changed behavior afaict.

Excellent, thanks!

PR #688, opened 2021-05-03
fix d6c51ca: ensure recrop actually applies to image

bertsky:


bertsky:
Looks like `test_image_feature_selectoro` needs to be updated to reflect `recropped` now appearing as image feature.

kba (approved):
It makes sense insofar as I can follow. We should add `recropped` to https://github.com/OCR-D/spec/blob/master/page.md#alternativeimage-classification

bertsky:
> We should add `recropped` to https://github.com/OCR-D/spec/blob/master/page.md#alternativeimage-classification

Or we always do recropping after deskewing (subsuming `recropped` under `deskewed`), and assume AlternativeImage producers abide by that rule (just as we are already relying on them to enlarge the canvas when rotating). Hard to enforce this though.

PR #692, opened 2021-06-17
add ALTO schema

bertsky:


bertsky:
Should probably differentiate ALTO versions, though.

Issue #693, opened 2021-06-19
processor decorator: working-dir not honoured


bertsky:
Currently, when running any Pythonic processor outside of a workspace by using the `--working-dir` option (but not the `--mets` option), the wrapper will fail with

```
Traceback (most recent call last):
  File "/local/ocr-d/ocrd_all/venv/local/sub-venv/headless-tf1/bin/ocrd-sbb-binarize", line 8, in <module>
    sys.exit(cli())
  File "/local/ocr-d/ocrd_all/venv/local/sub-venv/headless-tf1/lib/python3.6/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/local/ocr-d/ocrd_all/venv/local/sub-venv/headless-tf1/lib/python3.6/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/local/ocr-d/ocrd_all/venv/local/sub-venv/headless-tf1/lib/python3.6/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/local/ocr-d/ocrd_all/venv/local/sub-venv/headless-tf1/lib/python3.6/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/local/ocr-d/ocrd_all/venv/local/sub-venv/headless-tf1/lib/python3.6/site-packages/sbb_binarize/ocrd_cli.py", line 130, in cli
    return ocrd_cli_wrap_processor(SbbBinarizeProcessor, *args, **kwargs)
  File "/local/ocr-d/ocrd_all/venv/local/sub-venv/headless-tf1/lib/python3.6/site-packages/ocrd/decorators/__init__.py", line 65, in ocrd_cli_wrap_processor
    raise Exception(msg)
Exception: File does not exist: mets.xml
```

It does not even try to look under `working_dir`. 

This is the code responsible:
https://github.com/OCR-D/core/blob/b07ff2b2fb2f540f6c89cfdd1bbdbfc06224d8fc/ocrd/ocrd/decorators/__init__.py#L62-L65

IMHO the correct implementation would look very much like the existing code for `ocrd workspace`:
https://github.com/OCR-D/core/blob/b07ff2b2fb2f540f6c89cfdd1bbdbfc06224d8fc/ocrd/ocrd/cli/workspace.py#L27-L40

Related: #517 (where mets/workspace options have been fixed for `ocrd workspace`)

PR #696, opened 2021-06-22
Use the same logic for handling --mets/--mets-basename/--working-dir in processors and "ocrd workspace" CLI

kba:
See #693 . This PR factors the resolving logic for the CLI options of `ocrd workspace` out to `resolve_mets_arguments` and reuses that in the processor wrapper, so they should follow the same logic sans the `--mets-basename` which is not part of the processor CLI spec (should it be?).


ocrd_utils/ocrd_utils/os.py
```diff
@@ -120,3 +120,44 @@ def get_fileobject(self, **kwargs):
...
+        raise ValueError("Use either --mets or --mets-basename, not both")
+    elif not mets_basename and mets_url:
+        mets_basename = Path(mets_url).name
+    elif not mets_basename and not mets_url:
+        mets_basename = 'mets.xml'
+    else:
+        (log.warning if log else print)(DeprecationWarning("--mets-basename is deprecated. Use --mets/--directory instead"))
+
+    if directory and mets_url:
+        # XXX check whether mets_url has no parents, i.e. is actually the mets_basename
```
kba, line 139: Background: Due to how we set defaults in the click decorator, `--mets-url` will default to `mets.xml`. In that case (URL being a plain filename), behave as if `--mets-url` was omitted but `--mets-basename` was given. Not sure how robust that is though :grimacing:

ocrd/ocrd/cli/workspace.py
```diff
@@ -19,25 +19,7 @@ class WorkspaceCtx():
...
-            if mets_url.startswith('http') or mets_url.startswith('https:'):
-                raise ValueError("--mets is an http(s) URL but no --directory was given")
-            directory = dirname(abspath(mets_url)) or getcwd()
-        elif directory and not mets_url:
-            directory = abspath(directory)
-            mets_url = join(directory, mets_basename)
-        else:
-            directory = getcwd()
-            mets_url = join(directory, mets_basename)
+        directory ,mets_url, mets_basename = resolve_mets_arguments(directory, mets_url, mets_basename, self.log)
```
bertsky, line 22: Wouldn't it be better to keep the `DeprecationWarning` here, since it only applies to the workspace CLI?

ocrd/ocrd/cli/workspace.py
```diff
@@ -10,7 +10,7 @@
 import click
 
 from ocrd import Resolver, Workspace, WorkspaceValidator, WorkspaceBackupManager
-from ocrd_utils import getLogger, initLogging, pushd_popd, EXT_TO_MIME
+from ocrd_utils import getLogger, initLogging, pushd_popd, EXT_TO_MIME, resolve_mets_arguments
```
bertsky, line 13: Why not in `ocrd.resolver` or `ocrd.workspace`?

ocrd_utils/ocrd_utils/os.py
```diff
@@ -121,16 +121,19 @@ def atomic_write(fpath):
...
     """
     Resolve the ``--mets``, ``--mets-basename`` and `--directory`` argument
     into a coherent set of arguments according to https://github.com/OCR-D/core/issues/517
     """
     if mets_basename and mets_url:
         raise ValueError("Use either --mets or --mets-basename, not both")
     if mets_basename and not mets_url:
-        log.warning(DeprecationWarning("--mets-basename is deprecated. Use --mets/--directory instead"))
-    mets_basename = mets_basename if mets_basename else 'mets.xml'
+        (log.warning if log else print)(DeprecationWarning("--mets-basename is deprecated. Use --mets/--directory instead"))
```
bertsky: Wow, that's almost Lisp syntax. I didn't know Python is that expressive.

ocrd_utils/ocrd_utils/os.py
bertsky: I'd prefer if the exception / check could remain in place. Why don't you simply do the `abspath` thing to the processor decorator's `--mets` option as well?

ocrd_utils/ocrd_utils/os.py
```diff
@@ -128,25 +128,36 @@ def resolve_mets_arguments(directory, mets_url, mets_basename, log=None):
...
         mets_basename = 'mets.xml'
+    else:
+        (log.warning if log else print)(DeprecationWarning("--mets-basename is deprecated. Use --mets/--directory instead"))
+
     if directory and mets_url:
-        directory = abspath(directory)
-        if not abspath(mets_url).startswith(directory):
-            raise ValueError("--mets has a directory part inconsistent with --directory")
+        # XXX check whether mets_url has no parents, i.e. is actually the mets_basename
+        if Path(mets_url).parent == Path('.'):
```
bertsky, line 140: That's a strong assumption though. The user might run with `--mets data/mets.xml -w data`. The old solution was simpler AFAICS (see above).

ocrd_utils/ocrd_utils/os.py
kba: Because `--mets` might be `mets.xml` (as is the case when calling a processor with `--working-dir` but without explicit `--mets`.

So for example: `ocrd-dummy --working-dir /path/to/workspace`:
* `abspath(--directory) == /path/to/workspace`
* `abspath(--mets)  == /home/current/working/dir/mets.xml`

Which would be inconsistent. This is the case where I think it makes sense to treat `--mets` as `--mets-basename`.

ocrd/ocrd/cli/workspace.py
kba: The deprecation warning will still be raised, it's just been moved to `ocrd_utils.os.resolve_mets_arguments`:

```python
if mets_basename and mets_url:                                                                                           
    raise ValueError("Use either --mets or --mets-basename, not both")                                                   
elif not mets_basename and mets_url:                                                                                     
    mets_basename = Path(mets_url).name                                                                                  
elif not mets_basename and not mets_url:                                                                                 
    mets_basename = 'mets.xml'                                                                                           
else:                                                                                                                    
    (log.warning if log else print)(DeprecationWarning("--mets-basename is deprecated. Use --mets/--directory instead")) 
```

ocrd/ocrd/cli/workspace.py
kba: Mostly because functions are easier to test than methods. I was thinking of making it part of the `Resolver` since that has to be instantiated where the method is used. Would that be better?

ocrd_utils/ocrd_utils/os.py
kba: Lispier than https://github.com/OCR-D/core/blob/master/ocrd/ocrd/workspace.py#L573-L580 ? :D

```python
               if (all(feature in featureset
                        for feature in feature_selector.split(',') if feature) and
                    not any(feature in featureset
                            for feature in feature_filter.split(',') if feature) and
                    len(featureset.difference(auto_features)) >= \
                    len(best_features.difference(auto_features))):
                    best_features = featureset
                    best_image = alternative_image
```

ocrd_utils/ocrd_utils/os.py
kba: `--mets data/mets.xml`: Then `Path(mets_url).parent` would be `Path('data')`, not `Path(.)`

ocrd_utils/ocrd_utils/os.py
bertsky: Ah, I see, there **is** ambiguity in that CLI! So basically it does not make sense to provide `--mets` _and_ `--working-dir` – while it does to provide `--mets-basename` with `--working-dir` or `--mets` alone.

Couldn't we just make this implicit exclusiveness explicit by throwing an exception if both are used together (even if they do happen to contain the same directory prefix)? (Better simple and strict than complex and surprising...)

ocrd/ocrd/cli/workspace.py
bertsky: Yes, but _shouldn't_ it remain here (since there is nothing to deprecate elsewhere)?

ocrd/ocrd/cli/workspace.py
bertsky: Yes, that would look better IMHO.

ocrd_utils/ocrd_utils/os.py
bertsky: Yes, got it.

ocrd_utils/ocrd_utils/os.py
bertsky: But that's just predicating on a list of booleans. Yours is predicating on predicates (higher order). I guess subconsciously I was aware functions are first class citizens in Python, too. I just didn't know you can use the the parentheses syntax of standard lambda calculus for that. (In many other places, indentation is used instead of s-exp syntax...)

ocrd/ocrd/cli/workspace.py
kba: 56edc4c9 reintroduces the DeprecationWarning to `ocrd.cli.workspace` and replaces this call with an actual Python warning.

BTW I rue the day of thinking that `--mets-basename` was a good idea...

ocrd/ocrd/cli/workspace.py
kba: 56edc4c9 makes it a method of resolver and 187bc7fc makes sure the instantiation of the workspace uses the results of `Resolver.resolve_mets_arguments` as-is, without doing any `abspath` or `basename` calls.

ocrd_utils/ocrd_utils/os.py
kba: > (Better simple and strict)

Agreed, however we do need the combination of --directory and --mets-url for the case that --mets-url is an HTTP(S) URL (think: `ocrd workspace clone`), so I've changed the resolution logic to be based on that, which hopefully makes things less ambiguous...

kba:
> I'd prefer if the exception / check could remain in place. Why don't you simply do the abspath thing to the processor decorator's --mets option as well?

(cannot reply directly to comments on outdated code apparently) Can you have another look at the code pls? I think I have cleaned up the logic properly now. The reason we cannot easily do `abspath(--mets)` is because `--mets` might be a proper URL not a filename.

PR #699, opened 2021-06-30
ocrd_page with_etree and improved docstrings

bertsky:
This picks up on an old idea: #313 originally provided lxml etree nodes alongside the parsed OcrdPage DOM. This is helpful in processors and core for doing XPath queries on PAGE-XML (without the need to regenerate and reparse). It also allows some basic _mapping_ between DOM objects and lxml nodes:

```python
from ocrd_models.ocrd_page import parseEtree
pcgts, pcgts_root, pcgts_mapping, pcgts_revmap = parseEtree('page.xml')
line = pcgts.get_Page().get_TextRegion()[2].get_TextLine()[1]
assert pcgts_mapping[id(line)].attrib['id'] == line.id
```

The same can be achieved via:
```python
from ocrd_modelfactory import page_from_file, page_from_image
pcgts, pcgts_root, pcgts_mapping, pcgts_revmap = page_from_image(input_file, with_etree=True)
pcgts, pcgts_root, pcgts_mapping, pcgts_revmap = page_from_file(input_file, with_etree=True)
```

I have _not_ included shortcuts like `obj2node` or `node2id`, since (in the current implementation) one would need to pass the mappings along anyway. 

This is deliberate: This time, let's **not**:
1. monkey-patch the generated classes to store the mappings automatically, and 
2. take hold of implementation details to get a reverse mapping of memory objects from their `id`:
    ```python
    def di(ID):
        return _ctypes.PyObj_FromPtr(ID)
    ```

IIRC these hacks (which @cneud called _a path to the dark side_) were the culprits against adopting the old PR entirely, with the [main argument](https://github.com/OCR-D/core/pull/313#discussion_r370090562) being that we might loose performance _and_ stability, esp. when the DOM is changed and thus becomes inconsistent with the etree.

The reverse mapping in the present form might still be helpful to merely check against DOM object candidates already of interest.

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_page.py
+++ b/ocrd_models/ocrd_models/ocrd_page.py
@@ -7,6 +7,7 @@ __all__ = [
     'parse',
     'parseEtree',
     'parseString',
+    'OcrdPage',
 
     "AdvertRegionType",
     "AlternativeImageType",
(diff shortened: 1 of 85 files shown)


bertsky:
I also went to some length fixing the docstring / rST syntax to get a decent API doc browsing experience. I had to try out various forms for cross-referencing and line breaks – Sphinx/autodoc/napoleon documentation was not very helpful. It looks good in my generated HTML now, so I recommend doing `make docs-push` (but @kba please don't forget to include the `docs/build/html/_modules/` subdir – currently these files are not on the webserver, which is why all the source browsing links are 404ing).

(Oh, and I found a way to include an API doc of our `click` CLIs – via `sphinx_click`, it's really nice!)

bertsky:
CI failure:
```
E: AssertionError: 'invalid choice: foo' not found in "... Error: Invalid value for '-l' / '--log-level': 'foo' is not one of 'OFF', 'ERROR', 'WARN', 'INFO', 'DEBUG', 'TRACE'
```

Looks like a newer `click` version produces a different kind of error message?

kba:
> I also went to some length fixing the docstring / rST syntax to get a decent API doc browsing experience. I had to try out various forms for cross-referencing and line breaks – Sphinx/autodoc/napoleon documentation was not very helpful. It looks good in my generated HTML now, so I recommend doing `make docs-push` (but @kba please don't forget to include the `docs/build/html/_modules/` subdir – currently these files are not on the webserver, which is why all the source browsing links are 404ing).
> 
> (Oh, and I found a way to include an API doc of our `click` CLIs – via `sphinx_click`, it's really nice!)

Very good, thank you. I will have a look on the API docs and GT guidelines in the website this week.




> CI failure:
> 
> ```
> E: AssertionError: 'invalid choice: foo' not found in "... Error: Invalid value for '-l' / '--log-level': 'foo' is not one of 'OFF', 'ERROR', 'WARN', 'INFO', 'DEBUG', 'TRACE'
> ```
> 
> Looks like a newer `click` version produces a different kind of error message?

Indeed, I'll adapt the check.

ocrd_models/ocrd_models/ocrd_page.py
```diff
@@ -129,9 +130,56 @@
...
+        A tuple of
+         * The root object in the tree.
+         * The full node tree.
+         * A mapping from object IDs to tree nodes.
+         * A reverse mapping from tree nodes to object IDs.
+    """
+)
+
+# fix generated (malformed) docstrings
+parseString.__doc__ = (
```
kba, line 165: solid hack, will have to remember this :+1:

kba (approved):
Thanks, this is thorough and has none of the black magic that doomed #313.

kba:
Merged to master and released in https://github.com/OCR-D/core/releases/tag/v2.25.0

PR #700, opened 2021-06-30
patch ocrd_page_generateds for automatic inheritance

### LINKED ISSUE #698: OcrdPage: propagate inherited attributes and TextStyle
PAGE-XML features an implicit inheritance relation between various elements of the hierarchy:

- `Page/TextStyle → TextRegion*/TextStyle → TextLine/TextStyle → Word/TextStyle → Glyph/TextStyle`
- `Page/@primaryScript → TextRegion*/@primaryScript → TextLine/@primaryScript → Word/@primaryScript → Glyph/@script`
- `Page/@secondaryScript → TextRegion*/@secondaryScript → TextLine/@secondaryScript → Word/@secondaryScript → Glyph/@script`
- `Page/@primaryLanguage → TextRegion*/@primaryLanguage → TextLine/@primaryLanguage → Word/@language`
- `Page/@secondaryLanguage → TextRegion*/@secondaryLanguage → TextLine/@secondaryLanguage → Word/@language`
- `Page/@readingDirection → TextRegion*/@readingDirection → TextLine/@readingDirection → Word/@readingDirection`
- `Page/@textLineOrder → TextRegion*/@textLineOrder`

These relations are only documented and cannot be automatically implemented in a generated DOM. But these semantics are important, and it would make writing processors much easier if they would be implemented. 

For example, if I want to know if the current segment belongs to a certain script, I'd have to:
- check the element type, what kind of attribute name applies (`@script` or `@primaryScript` / `@secondaryScript`)
- check if that is set locally
- otherwise check the parent element's `@primaryScript` etc

This is very hard to achieve with XPath (because disjunction/unions are only possible on nodesets, not on predicates). And with the DOM it requires a lot of code.

But we could facilitate this by simply propagating all inherited features during `.build()` – we'd only have to patch `ocrd_page_generateds`.

We already have the user methods mechanism for patching, and we could simply use `buildChildren` to propagate all of the above attributes (going bottom up), because attributes of parents are build before those of children.

But for `TextStyle`, it's more complicated: on all hierarchy levels except the `Page` level, `TextStyle` sorts _after_ the logical children and thus is only built after they are built. Also, one would need to merge style attributes between levels (we usually have true, false and not set).

bertsky:
Fix #698


kba (approved):

kba (approved):
Looks good and patch applies fine. Proper support for inheritance of TextStyles will make life easier for e.g. the font recognition project. Any reason not to include in the next release together with #744 and #753?

bertsky:
> Any reason not to include in the next release together with #744 and #753?

Just to differentiate purely technical changes (the other PRs) from functional ones / new behaviour such as this. But release as you see fit.

kba:
further discussion in https://github.com/bertsky/core/pull/2

PR #713, opened 2021-09-02
log to STDERR not STDOUT, fix #667

### LINKED ISSUE #667: Log message about logging configuration file interferes with bashlib
With a logging configuration file at `$HOME/ocrd_logging.conf`, this happens:

```
ocrd workspace find -G OCR-D-IMG -k local_filename -k ID -k pageId                                
13:04:27.265 DEBUG ocrd.logging - Picked up logging config at /home/kba/ocrd_logging.conf
OCR-D-IMG/INPUT_0017.tif        INPUT_0017      PHYS_0017
OCR-D-IMG/INPUT_0020.tif        INPUT_0020      PHYS_0020
```

For bashlib processors like ocrd_im6convert, this will lead to an error like:

```
ocrd-im6convert: line 68: fields[1]: unbound variable
```

My first idea would be to make that "self-logging" behavior optional. But to make it optional, we'd have to rely on the configuration file. So if users want to debug logging itself, such as the configuration file not being found, they would have to configure that - in a configuration file :-?

kba:
The default logger will log to STDERR. As discussed in #667, this will eliminate the need for workarounds like completely disabling logging to ensure logging output does not interfere with parseable output.

The default configuration file has also been adapted accordingly.

### DIFF
--- a/ocrd_utils/ocrd_logging.conf
+++ b/ocrd_utils/ocrd_logging.conf
@@ -82,12 +82,12 @@ qualname=PIL
 
 
 #
-# handle stdout output
+# handle stderr output
 #
 [handler_consoleHandler]
 class=StreamHandler
 formatter=defaultFormatter
-args=(sys.stdout,)
+args=(sys.stderr,)
 
 #
 # example logfile handler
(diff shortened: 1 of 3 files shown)


bertsky (approved):

bertsky:
Oh, but 3 tests still need to be updated, too.

paulpestov (approved):

PR #715, opened 2021-09-08
workspace.merge: fix file-grp parameter

bertsky:
This fixes the following previously undetected bug:

```sh
ocrd workspace merge -g PHYS_0015..PHYS_0812 -G BIN-SBB --copy-files ../other/mets.xml
TypeError: merge() got an unexpected keyword argument 'file_grp'
```

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -539,7 +539,7 @@ def set_id(ctx, id):   # pylint: disable=redefined-builtin
 
 @workspace_cli.command('merge')
 @click.argument('METS_PATH')
-@click.option('--copy-files', is_flag=True, help="Copy files as well", default=True)
+@click.option('--copy-files/--no-copy-files', is_flag=True, help="Copy files as well", default=True, show_default=True)
 @click.option('--fileGrp-mapping', help="JSON object mapping src to dest fileGrp")
 @mets_find_options
 @pass_workspace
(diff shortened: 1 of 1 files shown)


bertsky:
Also, I wonder if we should improve the docstring/help text for `--copy-files`: Since it is `default=True`, it would be better to say that, and mention `--no-copy-files` for opt-out.

bertsky:
Plus, I have just observed that no files have been copied physically. Investigating ...

bertsky:
> Plus, I have just observed that no files have been copied physically. Investigating ...

It's strange. The src files exist (I checked with instrumentation), and I get no exceptions. Still, no files are written. More tomorrow...

bertsky:
> > Plus, I have just observed that no files have been copied physically. Investigating ...
> 
> It's strange. The src files exist (I checked with instrumentation), and I get no exceptions. Still, no files are written. More tomorrow...

The problem was with Click's handling of the `copy_files` option. Apparently, if you have a boolean flag and provide only a single option name (i.e. no `.../--no-copy-files`) and set `default=True`, then **the logic is inversed**, i.e. the option is used to _disable_ the variable.

So what should we do here? I think the UX is broken at this point – either we rename the option (`--no-copy-files` or `--keep-files`), or we provide both values (`--copy-files/--no-copy-files`).

kba:
> or we provide both values (--copy-files/--no-copy-files).

We should do that, I have been doing this for new CLI for a while now (always provide both options for booleans).

ocrd/ocrd/cli/workspace.py
```diff
@@ -539,7 +539,7 @@ def set_id(ctx, id):   # pylint: disable=redefined-builtin
 
 @workspace_cli.command('merge')
 @click.argument('METS_PATH')
-@click.option('--copy-files', is_flag=True, help="Copy files as well", default=True)
+@click.option('--copy-files/--no-copy-files', is_flag=True, help="Copy files as well", default=True, show_default=True)
```
kba, line 542: :+1:

Issue #719, opened 2021-09-15
`ocrd resmgr download ocrd-tesserocr-recognize <model>` is broken: tries to download from old master branch


twerkmeister:
Using latest `ocrd-all:maximum` docker container ✌️ 

To reproduce download some training data you don't already have:
`ocrd resmgr download ocrd-tesserocr-recognize Latin.traineddata`

you'll see in the log 
`INFO ocrd.resource_manager._download_impl - Downloading https://github.com/tesseract-ocr/tessdata_fast/raw/master/script/Latin.traineddata to /usr/local/share/ocrd-resources/ocrd-tesserocr-recognize/Latin.traineddata`

and just after that 

```
Traceback (most recent call last):
  File "/usr/bin/ocrd", line 33, in <module>
    sys.exit(load_entry_point('ocrd', 'console_scripts', 'ocrd')())
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/build/core/ocrd/ocrd/cli/resmgr.py", line 144, in download
    progress_cb=lambda delta: bar.update(delta)
  File "/build/core/ocrd/ocrd/resource_manager.py", line 210, in download
    self._download_impl(url, fpath, progress_cb)
  File "/build/core/ocrd/ocrd/resource_manager.py", line 161, in _download_impl
    total = size if size else int(r.headers.get('content-length'))
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
as the file didn't exist. Used
[...]

PR #726, opened 2021-10-13
ocrd_models requires six / tesseract resource URL change

hnesk:
Two small fixes (accidently in one PR, sorry)

1) ocrd_models/ocrd_page_generateds.py imports from `six`, but it is not listed in `requirements.txt`
....
in ocrd_models/ocrd_page_generateds.py line 27
    from six.moves import zip_longest
ModuleNotFoundError: No module named 'six'

2) The tesseract resource URLs changed (#719), this was fixed in 5f1c58974cb5e84aaf186be88c53771e0640131c except for the resource `configs` which still used the old branch name

### DIFF
--- a/ocrd/ocrd/resource_list.yml
+++ b/ocrd/ocrd/resource_list.yml
@@ -45,12 +45,12 @@ ocrd-tesserocr-recognize:
     parameter_usage: 'without-extension'
     description: Tesseract Latin model
     size: 89384811
-  - url: https://github.com/tesseract-ocr/tesseract/archive/master.tar.gz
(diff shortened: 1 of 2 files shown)


bertsky:
> I would prefer not to have to include `six` in the `requirements.txt` since we only support Python 3.6+. We could probably patch this out of the generateds code (`s,from six import zip_longest,from itertools import zip_longest`).

I agree.

kba:
https://github.com/OCR-D/core/pull/726/commits/2e2d01e4aacc7e9e04ae2c4d7e658c7545182d7c is fixed in #732. Thanks for contributing!

PR #732, opened 2021-11-03
Remove requirement on six

kba:
No need to be backwards-compatible with Python 2.x

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_page_generateds.py
+++ b/ocrd_models/ocrd_models/ocrd_page_generateds.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 #
-# Generated Wed Jun 30 17:57:54 2021 by generateDS.py version 2.35.20.
+# Generated Wed Nov  3 12:30:32 2021 by generateDS.py version 2.35.20.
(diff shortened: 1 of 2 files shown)


mikegerber:
I gave it a quick test using the dinglehopper test suite (includes a OCR-D test) and a manual test. It seems to fix my issue!

ocrd_models/ocrd_models/ocrd_page_generateds.py
```diff
@@ -1229,7 +1229,7 @@ def __hash__(self):
     @property
     def id(self):
         if hasattr(self, 'pcGtsId'):
-            return self.pcGtsId
+            return self.pcGtsId or ''
```
mikegerber, line 1232: Is this an accidental change? It does not seem related to six?

mikegerber: Ah, this is the generated file and behaviour seemed to have changed slightly. The question is if this is wanted or not

kba: Yes, this is intended, a change proposed by @bertsky to https://github.com/OCR-D/core/blob/master/ocrd_models/ocrd_page_user_methods/id.py which I forgot to regenerate the file for.

PR #733, opened 2021-11-03
resmgr.add_to_user_database: make sure resdict is always defined

### LINKED ISSUE #689: resmgr: reference before assignment
https://github.com/OCR-D/core/blob/b07ff2b2fb2f540f6c89cfdd1bbdbfc06224d8fc/ocrd/ocrd/resource_manager.py#L113

Here, `resdict` is not always defined.

```
  File "/usr/bin/ocrd", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 829, in __call__
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1259, in invoke
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1259, in invoke
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 1066, in invoke
  File "/usr/local/lib/python3.6/dist-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/ocrd/cli/resmgr.py", line 52, in list_installed
    for executable, reslist in resmgr.list_installed(executable):
  File "/usr/lib/python3.6/site-packages/ocrd/resource_manager.py", line 84, in list_installed
    resdict = [self.add_to_user_database(this_executable, res_filename)]
  File "/usr/lib/python3.6/site-packages/ocrd/resource_manager.py", line 113, in add_to_user_database
    return resdict
UnboundLocalError: local variable 'resdict' referenced before assignment
```

kba:
fix #689

### DIFF
--- a/ocrd/ocrd/resource_manager.py
+++ b/ocrd/ocrd/resource_manager.py
@@ -12,7 +12,6 @@ from yaml import safe_load, safe_dump
 
 from ocrd_validators import OcrdResourceListValidator
 from ocrd_utils import getLogger
-from ocrd_utils.constants import HOME, XDG_DATA_HOME, XDG_CONFIG_HOME
 from ocrd_utils.os import list_all_resources, pushd_popd
 
 from .constants import RESOURCE_LIST_FILENAME, RESOURCE_USER_LIST_COMMENT
@@ -23,10 +22,12 @@ class OcrdResourceManager():
     Managing processor resources
     """
     def __init__(self):
+        from ocrd_utils.constants import XDG_CONFIG_HOME
         self.log = getLogger('ocrd.resource_manager')
         self.database = {}
         self.load_resource_list(Path(RESOURCE_LIST_FILENAME))
         self.user_list = Path(XDG_CONFIG_HOME, 'ocrd', 'resources.yml')
+        print(self.user_list)
         if not self.user_list.exists():
             if not self.user_list.parent.exists():
                 self.user_list.parent.mkdir(parents=True)
@@ -63,6 +64,7 @@ class OcrdResourceManager():
         """
         List installed resources, matching with registry by ``name``
         """
+        from ocrd_utils.constants import XDG_DATA_HOME
         ret = []
         if executable:
             all_executables = [executable]
@@ -97,7 +99,8 @@ class OcrdResourceManager():
             user_database = safe_load(f) or {}
         if executable not in user_database:
             user_database[executable] = []
-        if not self.find_resources(executable=executable, name=res_name, database=user_database):
+        resources_found = self.find_resources(executable=executable, name=res_name, database=user_database)
+        if not resources_found:
             resdict = {
                 'name': res_name,
                 'url': url if url else '???',
@@ -106,6 +109,8 @@ class OcrdResourceManager():
                 'size': res_size
             }
             user_database[executable].append(resdict)
+        else:
+            resdict = resources_found[0]
         with open(self.user_list, 'w', encoding='utf-8') as f:
             f.write(RESOURCE_USER_LIST_COMMENT)
             f.write('\n')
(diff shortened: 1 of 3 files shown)


ocrd/ocrd/resource_manager.py
```diff
@@ -23,10 +22,12 @@ class OcrdResourceManager():
     Managing processor resources
     """
     def __init__(self):
+        from ocrd_utils.constants import XDG_CONFIG_HOME
         self.log = getLogger('ocrd.resource_manager')
         self.database = {}
         self.load_resource_list(Path(RESOURCE_LIST_FILENAME))
         self.user_list = Path(XDG_CONFIG_HOME, 'ocrd', 'resources.yml')
+        print(self.user_list)
```
bertsky, line 30:
```suggestion
```

ocrd/ocrd/resource_manager.py
```diff
@@ -12,7 +12,6 @@
 
 from ocrd_validators import OcrdResourceListValidator
 from ocrd_utils import getLogger
-from ocrd_utils.constants import HOME, XDG_DATA_HOME, XDG_CONFIG_HOME
```
bertsky, line 15: I am surprised to see you move imports from module level to functions here. What's the reason for this?

bertsky: Ah, apparently to be able to move away the resource config after `ocrd.resource_manager` has been loaded already in your new `test_resource_manager.py`. 

But couldn't you achieve the same result by moving L9 `from ocrd.resource_manager import OcrdResourceManager` to the test fixture instead in the test case itself?

bertsky (changes requested):

ocrd/ocrd/resource_manager.py
kba: I tried that, i.e. only instantiating `OcrdResourceManager` in the `monkey_patch_temp_xdg` method but that will only work once, because top-level import only happens once. Only solution I see that doesn't involve doing method-level imports in `OcrdResourceManager` is to squish all the tests into a single method.

kba: 31505067

tests/test_resource_manager.py
```diff
@@ -0,0 +1,43 @@
...
+    global _tempdir
+    with pushd_popd(tempdir=True) as tempdir:
+        old_config = ocrd_utils.constants.XDG_CONFIG_HOME
+        old_data = ocrd_utils.constants.XDG_DATA_HOME
+        ocrd_utils.constants.XDG_CONFIG_HOME = tempdir
+        ocrd_utils.constants.XDG_DATA_HOME = tempdir
+        from ocrd.resource_manager import OcrdResourceManager
+        yield tempdir, OcrdResourceManager()
+        ocrd_utils.constants.XDG_CONFIG_HOME = old_config
+        ocrd_utils.constants.XDG_DATA_HOME = old_data
```
bertsky, line 23: Wow, that's a great pattern – combining context manager and generator like this

bertsky (approved):

tests/test_resource_manager.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_config_created():
+    with monkey_patch_temp_xdg() as (tempdir, mgr):
+        f = Path(tempdir, 'ocrd', 'resources.yml')
+        assert f.exists()
```
bertsky, line 28: You probably meant 
```suggestion
        assert not f.exists()
```

tests/test_resource_manager.py
kba: I think that is just the standard contextmanager pattern, though you could technically use it as a real generator with multiple yields. I use that a lot for cases like this one where you want to do some setup and cleanup in a reusable way.

tests/test_resource_manager.py
kba: No, I was originally testing that the YAML file is created during instantiation. The problem is that any test or module that does `from ocrd import ...` will lead to the module being loaded before the XDG constant changing trickery happens. I.e. the test passes in isolation but not as part of the test suite. Only way to fix is to run the test separately, like we do for logging :/

bertsky: Oh, now I see

PR #734, opened 2021-11-03
workspace.remove-group: handle files not in subdirs gracefully

### LINKED ISSUE #728: workspace remove-group fails with FNFE
I happen to have a METS with files not under the conventional fileGrp directory. Upon trying `ocrd workspace remove-group -fr GRP` I then get:
```
  File "ocrd/workspace.py", line 241, in remove_file_group
    if Path(file_dir).is_dir() and not listdir(file_dir):
FileNotFoundError: [Errno 2] No such file or directory: ''
```

Digging into the code…
https://github.com/OCR-D/core/blob/061bae37f79aaac5aea64cf9afb8e1429b8243f5/ocrd/ocrd/workspace.py#L236-L253

…I believe the culprit is that `dirname` can give you empty strings when you have no directory component (unlike some `Path` functions, which will yield `.`). So the simplest fix would be to just add an extra check `file_dir and` in L252.

kba:
fix #728

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -238,7 +238,9 @@ class Workspace():
             for f in self.mets.find_files(fileGrp=USE):
                 self.remove_file(f, force=force, keep_file=keep_files, page_recursive=page_recursive, page_same_group=page_same_group)
                 if f.local_filename:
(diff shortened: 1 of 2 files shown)


bertsky (approved):
Oh, that's even better. Thanks!!

PR #736, opened 2021-11-04
Rename group

### LINKED ISSUE #729: workspace rename-group should also rename file IDs and file names if possible
In the current implementation, `ocrd workspace rename-group` does not rename the files themselves (and their `mets:file/@ID`), it only moves them from one directory (and `mets:fileGrp/@USE`) to another and redirects AlternativeImages.

This is not ideal, because most processors now use `ocrd_utils.make_file_id` for output file names (to avoid clashes and make file names more predictable and independent of `--page-id` runtime settings). But that function will not work with such files, because they no longer contain the fileGrp as part of their name.

kba:
* When setting an `OcrdFile.ID`, also remove `mets:fptr` to the old ID and add a `mets:fptr` with the updated ID
* When renaming filegroups, also
  * rename the file name to start with the new fileGrp, if the file name started with the old fileGrp
  * rename the file ID accordingly

Fix #729

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -276,12 +276,18 @@ class Workspace():
             url_replacements = {}
             log.info("Moving files")
             for mets_file in self.mets.find_files(fileGrp=old, local_only=True):
-                new_url = sub(r'^%s/' % old, '%s/' % new, mets_file.url)
+                new_url = mets_file.url
+                # Directory part
+                new_url = sub(r'^%s/' % old, r'%s/' % new, new_url)
+                # File part
+                new_url = sub(r'/%s' % old, r'/%s' % new, new_url)
                 url_replacements[mets_file.url] = new_url
                 # move file from ``old`` to ``new``
                 move(mets_file.url, new_url)
                 # change the url of ``mets:file``
                 mets_file.url = new_url
+                # change the file ID and update structMap
+                mets_file.ID = sub(r'^%s' % old, r'%s' % new, mets_file.ID)
             # change file paths in PAGE-XML imageFilename and filename attributes
             for page_file in self.mets.find_files(mimetype=MIMETYPE_PAGE, local_only=True):
                 log.info("Renaming file references in PAGE-XML %s" % page_file)
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -276,7 +276,7 @@ class Workspace():
             url_replacements = {}
             log.info("Moving files")
             for mets_file in self.mets.find_files(fileGrp=old, local_only=True):
-                new_url = mets_file.url
+                new_url = old_url = mets_file.url
                 # Directory part
                 new_url = sub(r'^%s/' % old, r'%s/' % new, new_url)
                 # File part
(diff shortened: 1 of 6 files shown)


ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -452,6 +452,22 @@ def remove_physical_page(self, ID):
         if mets_div:
             mets_div[0].getparent().remove(mets_div[0])
 
+    def remove_physical_page_fptr(self, fileId):
+        """
+        Delete all ``mets:fptr[@FILEID = fileId`` to ``mets:file[@ID == fileId]`` from alll entries in the physical ``mets:structMap`` ``mets:div`` entries) :py:attr:`ID`.
```
bertsky, line 457:
```suggestion
        Delete all ``mets:fptr[@FILEID = fileId]`` to ``mets:file[@ID == fileId]`` for :py:attr:`fileId` from all ``mets:div`` entries in the physical ``mets:structMap``.
```

ocrd_models/ocrd_models/ocrd_file.py
```diff
@@ -110,7 +110,13 @@ def ID(self, ID):
         """
         if ID is None:
             return
+        if self.mets is None:
+            raise Exception("OcrdFile %s has no member 'mets' pointing to parent OcrdMets" % self)
+        old_id = self.ID
         self._el.set('ID', ID)
+        # also update the references in the physical structmap
+        for pageId in self.mets.remove_physical_page_fptr(fileId=old_id):
+            self.pageId = pageId
```
bertsky, line 119: elegant!

ocrd/ocrd/workspace.py
```diff
@@ -276,12 +276,18 @@ def rename_file_group(self, old, new):
...
+                new_url = sub(r'^%s/' % old, r'%s/' % new, new_url)
+                # File part
+                new_url = sub(r'/%s' % old, r'/%s' % new, new_url)
                 url_replacements[mets_file.url] = new_url
                 # move file from ``old`` to ``new``
                 move(mets_file.url, new_url)
                 # change the url of ``mets:file``
                 mets_file.url = new_url
+                # change the file ID and update structMap
+                mets_file.ID = sub(r'^%s' % old, r'%s' % new, mets_file.ID)
```
bertsky, line 290: Just a precaution: maybe we should check whether `mets_file.mets` already contains the target ID – and then keep the old one. (For example, someone might have used `rename-group` earlier, and now tries to rename back with the new implementation...)

bertsky (approved):

ocrd/ocrd/workspace.py
kba: You mean something like this?

```python
# change the file ID and update structMap                                                                    
new_id = sub(r'^%s' % old, r'%s' % new, mets_file.ID)                                                        
try:                                                                                                         
    next(self.mets.find_files(ID=new_id))                                                                    
    log.warning("ID %s already exists, not changing ID while renaming %s -> %s" % (new_id, old_url, new_url))
except StopIteration:
    mets_file.ID = new_id                                                                                    
```

bertsky: yes, that's perfect!

kba: 5158e152

PR #744, opened 2021-11-12
make_file_id: no page_id number extraction

bertsky:
In `make_file_id`, if the input file's ID does not contain the input fileGrp, then do not attempt to extract the numerical part of the pageId (which might still clash).

But before fallback to purely numerical ID, additionally check if the input file's ID does already contain the pageId: in that case,
only append the output fileGrp to that ID (because it is sufficiently unique already).

(Often file IDs have two numbers, one of which will clash. In that case only the numerical fallback works. On the other hand, often the file IDs from non-OCRD data contain the pageId directly, in which case it's better to stick to that convention.)

### DIFF
--- a/ocrd_utils/ocrd_utils/str.py
+++ b/ocrd_utils/ocrd_utils/str.py
@@ -68,7 +68,7 @@ def make_file_id(ocrd_file, output_file_grp):
     """
     ret = ocrd_file.ID.replace(ocrd_file.fileGrp, output_file_grp)
     if ret == ocrd_file.ID:
-        if ocrd_file.pageId in ocrd_file.ID:
+        if ocrd_file.pageId and ocrd_file.pageId in ocrd_file.ID:
             # still sufficiently unique
             ret = ocrd_file.ID + '_' + output_file_grp
         else:
--- a/ocrd_utils/ocrd_utils/str.py
+++ b/ocrd_utils/ocrd_utils/str.py
@@ -62,24 +62,25 @@ def make_file_id(ocrd_file, output_file_grp):
     Derive a new file ID for an output file from an existing input file ``ocrd_file``
     and the name of the output file's ``fileGrp/@USE``, ``output_file_grp``.
     If ``ocrd_file``'s ID contains the input file's fileGrp name, then replace it by ``output_file_grp``.
+    Else if ``ocrd_file``'s ID contains the input file's pageId, then merely append ``output_file_grp``.
     Otherwise use ``output_file_grp`` together with the position of ``ocrd_file`` within the input fileGrp
-    (as a fallback counter). Increment counter until there is no more ID conflict.
+    (as a fallback counter), and increment counter until there is no more ID conflict.
     """
     ret = ocrd_file.ID.replace(ocrd_file.fileGrp, output_file_grp)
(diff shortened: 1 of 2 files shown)


bertsky:
> Often file IDs have two numbers, one of which will clash. In that case only the numerical fallback works.

To keep that option alive, it would probably also work to just strip any non-numerical content. Let me know if this is preferable (with lower priority, right before fallback), and I'll add that change.

ocrd_utils/ocrd_utils/str.py
```diff
@@ -62,24 +62,25 @@ def make_file_id(ocrd_file, output_file_grp):
...
+    (as a fallback counter), and increment counter until there is no more ID conflict.
     """
     ret = ocrd_file.ID.replace(ocrd_file.fileGrp, output_file_grp)
     if ret == ocrd_file.ID:
-        m = re.match(r'.*?(\d{3,}).*', ocrd_file.pageId or '')
-        if m:
-            n = int(m.group(1))
+        if ocrd_file.pageId and ocrd_file.pageId in ocrd_file.ID:
+            # still sufficiently unique
+            ret = ocrd_file.ID + '_' + output_file_grp
```
kba, line 73: It doesn't matter wrt to uniqueness but would it not be more consistent to do

```python
ret = output_file_grp + '_' + ocrd_file.ID
```

to imitate the OCR-D-internal convention for file IDs?

tests/test_utils.py
```diff
@@ -289,7 +289,7 @@ def test_make_file_id_570(self):
     def test_make_file_id_605(self):
         """https://github.com/OCR-D/core/pull/605"""
         mets = OcrdMets.empty_mets()
-        f = mets.add_file('1:!GRP', ID='FOO_0001', pageId='phys0001')
+        f = mets.add_file('2:!GRP', ID='FOO_0001', pageId='phys0001')
```
kba, line 292: We should probably disallow filegroups starting with a number because the resulting ID might lead to an invalid `xsd:ID` because they mustn't start with a number.

kba (approved):
> > Often file IDs have two numbers, one of which will clash. In that case only the numerical fallback works.
> 
> To keep that option alive, it would probably also work to just strip any non-numerical content. Let me know if this is preferable (with lower priority, right before fallback), and I'll add that change.

IIUC this seems unnecessary. Do you have an example of an ID where this might be pertinent?

ocrd_utils/ocrd_utils/str.py
bertsky: Right! Sry, I was too hasty there.

(Edited you comment as a suggestion and commited just now.)

tests/test_utils.py
bertsky: Yes, if #746 kicks in, such tests should all fail...

bertsky:
> Do you have an example of an ID where this might be pertinent?

I recently saw it in [Konzilsprotokolle GT](https://zenodo.org/record/215383#.Ya4WW1mYUUF), e.g. `1796-97_00000024_img`, and [Boernerianus](http://opacplus.bsb-muenchen.de/search?query=BV002197164), e.g. `img100-10`.

PR #747, opened 2021-11-24
run_processor: add mets:agent/mets:note with fileGrps and params

bertsky:
So far we only save the parameters into the PAGE-XML, but it's hard to reconstruct the exact workflow path from that alone. This adds the information which tool processed which fileGrps (and the parameters again – as I feel this redundancy is useful).

Note that the METS schema does allow attributes under `metsHdr/agent/note`, but only if they are from a foreign namespace – so I decided to just invent a (pseudo) OCR-D namespace.

Example output:
```XML
    <mets:agent TYPE="OTHER" OTHERTYPE="SOFTWARE" ROLE="OTHER" OTHERROLE="layout/segmentation/region">
      <mets:name>ocrd-cis-ocropy-segment v0.1.5</mets:name>
      <mets:note xmlns:ocrd="https://ocr-d.de" ocrd:option="input-file-grp">OCR-D-GT-SEG-PAGE-BINPAGE-sauvola-DENOISE-ocropy-DESKEW-tesseract-DESKEW-ocropy</mets:note>
      <mets:note xmlns:ocrd="https://ocr-d.de" ocrd:option="output-file-grp">OCR-D-SEG-BLOCK-ocropy</mets:note>
      <mets:note xmlns:ocrd="https://ocr-d.de" ocrd:option="parameter">{"level-of-operation": "page", "dpi": 0, "maxcolseps": 20, "maxseps": 20, "maximages": 10, "csminheight": 4, "hlminwidth": 10, "gap_height": 0.01, "gap_width": 1.5, "overwrite_order": true, "overwrite_separators": true, "overwrite_regions": true, "overwrite_lines": true, "spread": 2.4}</mets:note>
      <mets:note xmlns:ocrd="https://ocr-d.de" ocrd:option="page-id"/>
    </mets:agent>
```

### DIFF
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -88,20 +88,25 @@ def run_processor(
     processor.process()
     t1_wall = perf_counter() - t0_wall
     t1_cpu = process_time() - t0_cpu
-    logProfile.info("Executing processor '%s' took %fs (wall) %fs (CPU)( [--input-file-grp='%s' --output-file-grp='%s' --parameter='%s']" % (
+    logProfile.info("Executing processor '%s' took %fs (wall) %fs (CPU)( [--input-file-grp='%s' --output-file-grp='%s' --parameter='%s' --page-id='%s']" % (
         ocrd_tool['executable'],
         t1_wall,
         t1_cpu,
-        input_file_grp if input_file_grp else '',
-        output_file_grp if output_file_grp else '',
-        json.dumps(parameter) if parameter else {}
+        input_file_grp or '',
+        output_file_grp or '',
+        json.dumps(parameter) or '',
+        page_id or ''
     ))
     workspace.mets.add_agent(
         name=name,
         _type='OTHER',
         othertype='SOFTWARE',
         role='OTHER',
-        otherrole=otherrole
+        otherrole=otherrole,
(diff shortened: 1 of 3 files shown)


bertsky:
Not sure if this should also go to the spec, but it does not hurt having the implementation first (as proof of concept), and then discuss/modify.

kba (approved):
Good proposal, it always irked me that we have the provenance information (including parameters) only in the PAGE-XML and not the METS where it makes more sense to look for to actually reproduce a workflow.

Since we're not generating the METS API from XSD like we do for PAGE-XML I also see no problem with the custom `ocrd` XML namespace. In the interest of documentation and long-term support we should probably not only add this to the METS spec but also provide and publish a basic XSD that describes the option, so METS consumers know what `ocrd:option` is about.

bertsky:
> In the interest of documentation and long-term support we should probably not only add this to the METS spec but also provide and publish a basic XSD that describes the option, so METS consumers know what `ocrd:option` is about.

Agreed. So we basically translate the (processing part of the) CLI into an XSD under OCR-D namespace, and then document that our `metsHdr/agent/note` attributes refer to specific elements of that schema? (I am uncertain though whether a true validator which recursively retrieves schemas for each namespace would actually accept exchanging attributes for elements.) Or am I misreading your statement?

PR #753, opened 2021-12-02
bashlib: list-resources, show-resource, and input-files

bertsky:
Fixes #751

PR #759, opened 2021-12-07
Release 2.29.0

kba:
Preparing the next release, includes:

  * #758 
  *  #744 
  * #753

I would also like to include #700 if we can resolve the discussion on short notice, otherwise in the next release in a few weeks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 Changed:
 
-  * `ocrd_utils.make_file_id`: only increment number in ID if it is the page ID, #744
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Changed:
+
+  * `ocrd_utils.make_file_id`: only increment number in ID if it is the page ID, #744
```
bertsky, line 10: That's not what the change actually does.

```suggestion
  * `ocrd_utils.make_file_id`: combine with output fileGrp if input has pageId, but don't extract numbers, #744
```

bertsky (approved):
> I would also like to include #700 if we can resolve the discussion on short notice, otherwise in the next release in a few weeks.

Let's discuss tomorrow, then test some more, then release independently (as next minor).

Issue #587, opened 2020-09-03
Bundle tests/base.py


kba:
Related to #580 : OCR-D/core has a nifty `tests/base.py` test helper library with useful functionality beyond core. I sometimes copy that file to other projects which is redundant and leads to inconsistencies. It would be useful to have a central place from where to import it.

PR #762, opened 2021-12-09
Feat/rfct/pytest

M3ssman:
Refactoring
* tests/model/test_agent.py (fairly straight-forward-simple-example)
* tests/model/test_exif.py (use test params)
* tests/test_resolver.py (use mocks instead of requesting assets over network, reduces test execution drastically for poor networks, use params + fixtures as well)
* test/test_resolver_oai.py (just apply autopep8-formatting to file)

_all_ modules contain still the `main` - call at the end

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -8,7 +8,7 @@ recommonmark >= 0.5.0
 docstr-coverage
 pylint
 atomicwrites
-opencv-python
+opencv-python-headless
 deprecated
 click
 twine
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,3 +1,4 @@
+autopep8
 pytest >= 4.0.0
 generateDS == 2.35.20
 coverage >= 4.5.2
--- a/tests/model/test_agent.py
+++ b/tests/model/test_agent.py
@@ -1,12 +1,17 @@
 # -*- coding: utf-8 -*-
 
-from ocrd_models import OcrdAgent
+from ocrd_models import (
+    OcrdAgent
+)
+from tests.base import (
+    main
+)
 
 
 def test_init_role_no_name():
     ag = OcrdAgent(role='FOO')
-    assert 'FOO' == ag.role
-    assert None == ag.name
+    assert ag.role == 'FOO'
+    assert ag.name is None
 
 
 def test_init_role_as_string():
@@ -17,18 +22,22 @@ def test_init_role_as_string():
 
 def test_init_otherrole_and_othertype():
     ag = OcrdAgent(otherrole='BAR', othertype='x')
-    assert 'OTHER' == ag.role
-    assert 'BAR' == ag.otherrole
-    assert 'x' == ag.othertype
+    assert ag.role == 'OTHER'
+    assert ag.otherrole == 'BAR'
+    assert ag.othertype == 'x'
 
 
 def test_init_othertype():
     ag = OcrdAgent(othertype='foobar')
-    assert 'OTHER' == ag.type
+    assert ag.type == 'OTHER'
 
 
 def test_set_name():
     ag = OcrdAgent(name='foobar')
-    assert 'foobar' == ag.name
-    ag.name = 'barfoo'
-    assert 'barfoo' == ag.name
+    assert ag.name == 'foobar'
+    ag.name = 'barfoo' 
+    assert ag.name == 'barfoo'
+
+
+if __name__ == '__main__':
+    main(__file__)
--- a/tests/model/test_agent.py
+++ b/tests/model/test_agent.py
@@ -1,32 +1,34 @@
-from tests.base import TestCase, assets, main # pylint: disable=unused-import
+# -*- coding: utf-8 -*-
 
 from ocrd_models import OcrdAgent
 
-# pylint: disable=no-member
-class TestOcrdAgent(TestCase):
 
-    def test_basic1(self):
-        ag = OcrdAgent(role='FOO')
-        self.assertEqual(ag.role, 'FOO')
-        self.assertEqual(ag.name, None)
-        self.assertEqual(str(ag), '<OcrdAgent [type=---, othertype=---, role=FOO, otherrole=---, name=---]/>')
+def test_init_role_no_name():
+    ag = OcrdAgent(role='FOO')
+    assert 'FOO' == ag.role
+    assert None == ag.name
 
-    def test_basic2(self):
-        ag = OcrdAgent(otherrole='BAR', othertype='x')
-        self.assertEqual(ag.role, 'OTHER')
-        self.assertEqual(ag.otherrole, 'BAR')
-        self.assertEqual(ag.othertype, 'x')
 
-    def test_basic3(self):
-        ag = OcrdAgent(name='foobar')
-        self.assertEqual(ag.name, 'foobar')
-        ag.name = 'barfoo'
-        self.assertEqual(ag.name, 'barfoo')
+def test_init_role_as_string():
+    ag = OcrdAgent(role='FOO')
+    expected = '<OcrdAgent [type=---, othertype=---, role=FOO, otherrole=---, name=---]/>'
+    assert expected == str(ag)
 
-    def test_basic4(self):
-        ag = OcrdAgent(othertype='foobar')
-        self.assertEqual(ag.type, 'OTHER')
-        #  print(ag)
 
-if __name__ == '__main__':
-    main()
+def test_init_otherrole_and_othertype():
+    ag = OcrdAgent(otherrole='BAR', othertype='x')
+    assert 'OTHER' == ag.role
+    assert 'BAR' == ag.otherrole
+    assert 'x' == ag.othertype
+
+
+def test_init_othertype():
+    ag = OcrdAgent(othertype='foobar')
+    assert 'OTHER' == ag.type
+
+
+def test_set_name():
+    ag = OcrdAgent(name='foobar')
+    assert 'foobar' == ag.name
+    ag.name = 'barfoo'
+    assert 'barfoo' == ag.name
--- a/tests/model/test_exif.py
+++ b/tests/model/test_exif.py
@@ -55,7 +55,6 @@ def test_ocrd_exif(path, width, height, xResolution, yResolution, resolution, re
     assert ocrd_exif.compression == compression
 
 
-@pytest.mark.skipif(not sys.platform.startswith('linux'), reason="not platform-independent/stable")
 def test_ocrd_exif_serialize_xml():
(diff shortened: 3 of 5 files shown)


tests/model/test_exif.py
```diff
@@ -1,82 +1,78 @@
...
-        #      exif.to_xml(),
-        #      '<exif><width>2875</width><height>3749</height><photometricInterpretation>RGB</photometricInterpretation><compression>jpeg</compression><photometric_interpretation>None</photometric_interpretation><xResolution>300.0</xResolution><yResolution>300.0</yResolution><resolutionUnit>inches</resolutionUnit></exif>'
-        #  )
+pil_version_major, pil_version_minor = [
+    int(x) for x in pil_version.split('.')][:2]
+pil_below_83 = pil_version_major < 8 or (
+    pil_version_major == 8 and pil_version_minor < 3)
+
 
+@pytest.mark.parametrize("path,width,height,xResolution,yResolution,resolution,resolutionUnit,photometricInterpretation,compression", [
```
kba, line 24: very elegant and a useful pattern

tests/test_resolver.py
```diff
@@ -1,157 +1,281 @@
...
+
+
+def _get_kant_data(key):
+    if key in DATA_KANT.keys():
+        (path, mime) = DATA_KANT[key]
+        with open(path, mode='rb') as _file:
+            return (_file.read(), mime)
+
+
+def request_behavior(*args):
```
kba, line 45: This is also a candidate for reuse in a dedicated ocrd_testing package https://github.com/OCR-D/core/issues/587

tests/test_resolver_oai.py
```diff
@@ -10,13 +10,16 @@
 from ocrd_models.utils import extract_mets_from_oai_content
 from ocrd_utils import getLogger, initLogging, LOG_FORMAT
 
+
 @fixture(name="response_dir")
 def fixture_response_dir(tmpdir):
     src = './tests/data/response/oai_get_record_2200909.xml'
-    target_file = str(tmpdir.mkdir('responses').join('oai_get_record_2200909.xml'))
+    target_file = str(tmpdir.mkdir('responses').join(
```
kba, line 17: Just out of curiosity: Why the short line length? I know PEP8 and tools based on it call for 78 max line length, but I tend to increase that to at least 100-120 because I actually find it more readable if a method doesn't have so many lines. But I have been wondering what kind of problem this might cause.

kba (approved):
Great, this makes the tests much more concise and DRY.

requirements_test.txt
```diff
@@ -1,3 +1,4 @@
+autopep8
```
kba, line 1: I used pylint and black to catch issues and autopep8 to fix them in the past but we are not enforcing this because different developers have different aesthetic preferences - but maybe we should?

tests/model/test_agent.py
```diff
@@ -1,32 +1,43 @@
...
 
-    def test_basic3(self):
-        ag = OcrdAgent(name='foobar')
-        self.assertEqual(ag.name, 'foobar')
-        ag.name = 'barfoo'
-        self.assertEqual(ag.name, 'barfoo')
+def test_init_role_as_string():
+    ag = OcrdAgent(role='FOO')
+    expected = '<OcrdAgent [type=---, othertype=---, role=FOO, otherrole=---, name=---]/>'
+    assert expected == str(ag)
```
bertsky, line 20: Won't that take away the benefit of seeing the expected value along with the actual value whenever the assertion fails? (`pytest.assertEqual` shows both IIRC, but plain assertions only show the line of code, which does not even contain the literal value anymore here.)

tests/model/test_exif.py
```diff
@@ -1,82 +1,78 @@
...
-        self.assertEqual(exif.width, 944)
-        self.assertEqual(exif.height, 1472)
-        self.assertEqual(exif.xResolution, 1)
-        self.assertEqual(exif.yResolution, 1)
-        self.assertEqual(exif.resolution, 1)
-        self.assertEqual(exif.resolutionUnit, 'inches')
-        self.assertEqual(exif.photometricInterpretation, 'RGB')
-        self.assertEqual(exif.resolutionUnit, 'inches')
+@pytest.mark.skipif(not sys.platform.startswith('linux'), reason="not platform-independent/stable")
+def test_ocrd_exif_serialize_xml():
```
bertsky, line 59: IIUC this test is new – could you please elaborate / comment why this only works on Linux?

tests/test_resolver.py
```diff
@@ -1,157 +1,281 @@
...
+        resp.headers = {'Content-Type': mime}
+    return resp
+
+
+def test_workspace_from_url_bad():
+    with pytest.raises(Exception) as exc:
+        Resolver().workspace_from_url(None)
+
+    # check exception
+    assert "Must pass 'mets_url'" in str(exc)
```
bertsky, line 62: Interesting. More flexible than `with self.assertRaisesRegex(Exception, "Must pass 'mets_url'")`.

But I wonder: why is the `exc` binding checked outside of the scope of the context manager here?

tests/test_resolver.py
```diff
@@ -1,157 +1,281 @@
...
+
+
+@mock.patch("requests.get")
+def test_workspace_from_url_kant(mock_request, tmp_path):
```
bertsky, line 66: Neat!

tests/test_resolver.py
```diff
@@ -1,157 +1,281 @@
...
+
+
+@mock.patch("requests.get")
+def test_workspace_from_url_404(mock_request):
+    """Expected behavior when try create workspace from invalid online target
+    """
+
+    # arrange
+    url_404 = 'https://raw.githubusercontent.com/OCR-D/assets/master/data/kant_aufklaerung_1784/data/mets.xmlX'
+    mock_request.side_effect = Exception('HTTP request failed')
```
bertsky, line 145: Ah, now I get why this is called _side effect_ – it can be much more than just a function!

tests/test_resolver.py
```diff
@@ -1,157 +1,281 @@
...
+
+    ws2 = Resolver().workspace_from_nothing(tmp_path)
+    assert ws2.directory == tmp_path
+
+    with pytest.raises(Exception) as exc:
+        Resolver().workspace_from_nothing(tmp_path)
+
+    # assert
+    the_msg = "METS 'mets.xml' already exists in '%s' and clobber_mets not set" % tmp_path
+    assert the_msg in str(exc)
```
bertsky, line 230: Again, `exc` outside its scope... (?)

tests/test_resolver.py
bertsky: I still don't quite understand how the `tmp_path` arguments work here, though. Where is the `tempfile.TemporaryDirectory` being set up now?

tests/model/test_agent.py
kba: pytest does some black magic behind the scenes to turn "plain" assertions:

```python
    def test_foo():
        expected = [1,2,3]
        actual = {"foo": "bar"}
>       assert expected == actual
E       AssertionError: assert [1, 2, 3] == {'foo': 'bar'}
E         +[1, 2, 3]
E         -{'foo': 'bar'}
E         Full diff:
E         - {'foo': 'bar'}
E         + [1, 2, 3]
```

tests/test_resolver.py
kba: It's a builtin fixture in pytest, cf. https://docs.pytest.org/en/6.2.x/builtin.html. This means that if you use the particular names for fixtures (which can be builtin as in the case of `tmp_path` or looked up in the test suite for matching function names), pytest will notice this and call that fixture for both data and side-effects like setting up/tearing down temporary paths, like here.

tests/model/test_agent.py
bertsky: Ah, thanks!

tests/model/test_exif.py
bertsky: Understood. (Then perhaps just add a `todo` comment?)

tests/test_resolver.py
bertsky: Got it. _Builtin fixtures_ – or in other words: "Open sesame!"

kba: Yup, magic! Or rather clever use of code introspection and generation :-)

tests/test_resolver_oai.py
M3ssman: autopep8c might utilize custom configuration files (https://pypi.org/project/autopep8/#configuration), where such parameters can be set. I didn't include one, therefore it just takes the defaults. I don't have a personal favor about row length, it just puzzles me, when I hit formatting shortcut and the file gets literally rewritten :smile:

tests/model/test_exif.py
M3ssman: And anyway, used as example of the skipif-mark

tests/test_resolver.py
M3ssman: AFAIK it's regular Python Context Management with some extensions. Under the hood it's a Function Call that wraps the actual Expression which raises the Exception and produces an ExceptionInfo-object (the `exc` in the tests mentioned), which carries some of the Exception Attributes to the next lines. The line that raises the Exception must be in any case the final line in the pytest-Context Management, see: https://docs.pytest.org/en/6.2.x/reference.html#pytest.raises

M3ssman: Like `The execinfo is not part of the context manager scope (this is true for the context of any context manager), so it lives on after the context manager has exited, making this possible.` see: https://stackoverflow.com/questions/69510562/exception-value-is-not-being-caught-in-pytest-raises-context-manager

bertsky: That's an important detail. So you **must** do the assertion outside pytest.raises. Thanks for clarifying!

tests/test_resolver_oai.py
kba: Since we do have a [`pylintrc`](https://github.com/OCR-D/core/blob/master/.pylintrc) (which skips the `line-too-long` check entirely) in place, maybe we could add one for autopep8 and try to find common ground (i.e. argue 😆) about good code style conventions that work for all of us?

tests/model/test_exif.py
kba: I cannot now reproduce the original problem, though I have only tested it in Linux but with different python versions and no issue so far. I think the problem only happened in the CI. However, CI did not run for this pull request, I'll open a new one to verify.

kba: https://app.circleci.com/pipelines/github/OCR-D/core/1129/workflows/8969786f-c965-4c0f-bcab-5577839e0883/jobs/3540 no problems there either. I'll remove the `skipif` mark for now (but will definitely use that for tests that **are** platform-dependent in the future).

PR #767, opened 2021-12-20
Feat/rfct/pytest2

M3ssman:
Besides rather straight-forward re-construction some cases puzzle me (the testcases are marked and require reviews): 

* test_ocrd_mets.py # `test_add_file_ignore`
  when inserting the file with the ignore-flag, it looks like 2 file entries with same ID are created, which seems suspicious because this interferes with the regular IDREF-constraint of unique identifiers IMHO
* test_resolver.py # `test_resolve_image_grayscale`
  Tried to express and handle the deprecation warning appropriately, but failed to turn into modern usage within reasonable amount of time. Both methods mentioned in the deprecation warning (`workspace.image_from_page and ` and `workspace.image_from_segment`) are not drop-in replacements, the usage is IMHO completely different and not self-explanatory from the documentation. Looks like there are required completely different objects, API, etc.

### DIFF
--- a/tests/model/test_ocrd_file.py
+++ b/tests/model/test_ocrd_file.py
@@ -95,7 +95,7 @@ def test_create_ocrd_file_with_defaults_basename_wo_extension(local_filename, wo
     assert f.basename_without_extension == wo_extension
 
 
-@pytest.mark.xfail(reason="not possible anymore as of Fri Sep  3 13:11:00 CEST 2021")
+@pytest.mark.skip(reason="not possible anymore as of Fri Sep  3 13:11:00 CEST 2021")
 def test_file_group_wo_parent():
     with pytest.raises(ValueError) as val_err:
         OcrdFile(None)
@@ -105,11 +105,8 @@ def test_file_group_wo_parent():
 def test_file_group_wo_parent_new_version():
     """Test for new error message
     """
-    with pytest.raises(ValueError) as val_err:
+    with pytest.raises(ValueError, match=r"Must provide mets:file element this OcrdFile represent"):
         OcrdFile(None)
-    assert "Must provide mets:file element this OcrdFile represent" in str(
-        val_err.value)
-
 
 def test_ocrd_file_equality():
     mets = OcrdMets.empty_mets()
--- a/tests/model/test_ocrd_file.py
+++ b/tests/model/test_ocrd_file.py
@@ -1,97 +1,138 @@
-from tests.base import TestCase, main, create_ocrd_file, create_ocrd_file_with_defaults
-from ocrd_models import OcrdMets
+# -*- coding: utf-8 -*-
 
-class TestOcrdFile(TestCase):
+import pytest
 
-    def test_ocrd_file_without_id(self):
-        with self.assertRaisesRegex(ValueError, "set ID"):
-            create_ocrd_file('FOO')
+from tests.base import (
+    main, 
+    create_ocrd_file, 
+    create_ocrd_file_with_defaults
+)
+from ocrd_models import (
+    OcrdMets,
+    OcrdFile,
+)
 
-    def test_ocrd_file_without_filegrp(self):
-        with self.assertRaisesRegex(ValueError, "set fileGrp"):
-            create_ocrd_file(None, ID='foo')
 
-    def test_loctype(self):
-        f = create_ocrd_file_with_defaults()
-        self.assertEqual(f.loctype, 'OTHER')
-        self.assertEqual(f.otherloctype, 'FILE')
-        f.otherloctype = 'foo'
-        self.assertEqual(f.otherloctype, 'foo')
-        f.loctype = 'URN'
-        self.assertEqual(f.loctype, 'URN')
-        self.assertEqual(f.otherloctype, None)
-        f.otherloctype = 'foo'
-        self.assertEqual(f.loctype, 'OTHER')
+def test_ocrd_file_without_id():
+    with pytest.raises(ValueError) as val_err:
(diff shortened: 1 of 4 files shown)


kba:
> * test_ocrd_mets.py # `test_add_file_ignore`
>   when inserting the file with the ignore-flag, it looks like 2 file entries with same ID are created, which seems suspicious because this interferes with the regular IDREF-constraint of unique identifiers IMHO

This is expected behavior, as this is exactly what `ignore` does:

```python
"""
...
ignore (boolean): Do not look for existing files at all. Shift responsibility for preventing errors from duplicate ID to the user.
...
"""
# ...
mets_file = next(self.find_files(ID=ID), None)                  
if mets_file and not ignore:                                    
    if not force:                                               
        raise Exception("File with ID='%s' already exists" % ID)
```

kba:
> * Tried to express and handle the deprecation warning appropriately, but failed to turn into modern usage within reasonable amount of time. Both methods mentioned in the deprecation warning (`workspace.image_from_page and ` and `workspace.image_from_segment`) are not drop-in replacements, the usage is IMHO completely different and not self-explanatory from the documentation. Looks like there are required completely different objects, API, etc.

These tests were using the deprecated method deliberately. While the public method (`resolve_image_as_pil`) is deprecated, the private method (`_resolve_image_as_pil`) is not, it is actually used in the non-deprecated public method. Once we remove `resolve_image_as_pil` we should switch to the private method here.

I am not an expert on unit test conventions but it seemed important to test this method outside the context of its more complex cousins. If that's bad practice I can change it.

tests/model/test_ocrd_file.py
```diff
@@ -1,97 +1,138 @@
...
+                         )
+def test_create_ocrd_file_with_defaults_basename_wo_extension(local_filename, wo_extension):
+    """Behavior for ocrd_file_with_defaults
+    """
+
+    f = create_ocrd_file_with_defaults(local_filename=local_filename)
+    assert f.basename_without_extension == wo_extension
+
+
+@pytest.mark.xfail(reason="not possible anymore as of Fri Sep  3 13:11:00 CEST 2021")
```
kba, line 98: When do you use `xfail` and when `skip` and when remove the test altogether? I would probably just remove the test, since it's replaced with `test_file_group_wo_parent_new_version`.

tests/model/test_ocrd_mets.py
```diff
@@ -1,265 +1,364 @@
+# -*- coding: utf-8 -*-
```
kba, line 1: Is that necessary/does it help?

tests/model/test_ocrd_file.py
M3ssman: If it really does serve as documentation, it'll be best to keep it - otherwise drop it. But since it's already part of the repository history (and with the increase of entropy after all), don't be afraid to free yourself  :smile:

M3ssman: I'll try to place the col restriction with 150 in the config appropriately

tests/model/test_ocrd_mets.py
M3ssman: Well, I've found this in many places with Beginner Tutorials etc. It's because to get an alert on other OS where UTF-8 is not set as default charset (windows, for example).

tests/model/test_ocrd_file.py
M3ssman: Ok I've added to tox.ini, and at least my IDE respects it

M3ssman:
To proceed with `test_resolve_image_grayscale` , it should be renamed when it' about to test `` and/or `` with ending up with 2 additional tests, one for each way IMHO.

Testing protected stuff is no good practice, but if you've got very good reasons (and comments/docs to explain the motivations) it'll be okay.

With `test_add_file_ignore` - well, 2 identical IDREFs, I'm quite sure this is invalid METS-output. 
Why on earth should one wants to get this? If you can convince me, I'll put this in the tests comment an be ... fine ...

tests/model/test_ocrd_mets.py
kba: I always thought this was an emacs convention, like vim's modelines but was confused since I had you down as a VisualCode user :)

kba:
> Testing protected stuff is no good practice, but if you've got very good reasons (and comments/docs to explain the motivations) it'll be okay.

OK, then it is probably best to just remove the tests. If they are confusing you, they will confuse other potential developers and that's the oppposite of what we want from reading the tests.



> With `test_add_file_ignore` - well, 2 identical IDREFs, I'm quite sure this is invalid METS-output.
> Why on earth should one wants to get this? If you can convince me, I'll put this in the tests comment an be ... fine ...

It's about performance, really. If you **know** your data and the consequences of --ignore in `workspace.add_file` (which must be documented including the pitfalls about duplicate IDs obviously), then it is faster to not have to iterate through all the files for every add operation. The performance issue becomes more pronounced the more files you have in a workspace (for a book with 400 pages and 5 processors producing 1-2 files each, you can see how much lookups this can cause to preven duplicates. Then again, in such a case, one should probably use `workspace bulk-add` which does duplicate detection but saves time by not parsing/serializing METS for every add operation. Plus, parameter names like `--ignore` or `--force` are probably a bad design pattern because they are too generic/broad. If you are surprised by this behavior, so are others probably and we should not surprise/confuse users.

@bertsky What do you think?

M3ssman:
Renamed `test_resolve_image_grayscale` to `test_resolve_image_as_pil` to reflect the changing test subject

kba (approved):

PR #770, opened 2021-12-20
Feat/rfct/pytest3

M3ssman:
This refactoring of test_resource_manager is also meant as proposal to reduce the environmental and internal dependencies in the ocrd-modules.

As @kba pointed out, there's actually a strong dependency between the resource_manager-module to the time, when the constants `XDG_CONFIG_HOME` and `XDG_DATA_HOME` are loaded / evaluated.
The trouble with testing the resource_manager originates from the - somehow intransparent - relationship to the execution environment i.e. context - intransparent to the caller in the moment of execution, which makes it difficult to test any depending functionalities, since testing is a rather artificial execution. The rather plain "workaround" for this bias is the possibility to pass these this info to the objects which require them, as properties or at initialization time. I found myself often ending up in implementations like this, which only exist to ease testing purposes and nothing else.

This way there's no need to run the three resource_manager tests separate from regular unit tests.

Personally, I tried even to go further this line, which simply results in having no "constant" values for `HOME` or the xdg-stuff, but rather evaluated both only when they are *really* required, but this interferes with other modules - os.py and bashlib.py, the latter which I do not yet understand.

### DIFF
--- a/ocrd/ocrd/resource_manager.py
+++ b/ocrd/ocrd/resource_manager.py
@@ -20,18 +20,16 @@ class OcrdResourceManager():
     """
     Managing processor resources
     """
-    def __init__(self, xdb_config_home=None, xdb_data_home=None):
+    def __init__(self, xdg_config_home=None):
         self.log = getLogger('ocrd.resource_manager')
         self.database = {}
 
-        self.xdb_data_home = xdb_data_home
-        if not self.xdb_data_home:
-            self._init_data_home()
-        self.xdb_config_home = xdb_config_home
-        if not self.xdb_config_home:
-            self._init_config_home()
+        self._xdg_data_home = None
+        self.xdg_config_home = xdg_config_home
+        if not self.xdg_config_home:
+            self._calculate_xdg_config_home()
         self.load_resource_list(Path(RESOURCE_LIST_FILENAME))
-        self.user_list = Path(self.xdb_config_home, 'ocrd', 'resources.yml')
+        self.user_list = Path(self.xdg_config_home, 'ocrd', 'resources.yml')
         if not self.user_list.exists():
             if not self.user_list.parent.exists():
                 self.user_list.parent.mkdir(parents=True)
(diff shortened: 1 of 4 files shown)


tests/test_resource_manager.py
```diff
@@ -1,40 +1,71 @@
...
+    assert f == mgr.user_list
+    assert mgr.add_to_user_database('ocrd-foo', f)
+    mgr.list_installed()
+    proc = 'ocrd-anybaseocr-layout-analysis'
+    fpath = mgr.download(proc, CONST_RESOURCE_URL_LAYOUT, mgr.location_to_resource_dir('data'))
+    assert fpath.exists()
+    assert mgr.add_to_user_database(proc, fpath)
+
+
+def test_resources_manager_from_environment(tmp_path, monkeypatch):
```
kba, line 32: elegant

ocrd/ocrd/resource_manager.py
```diff
@@ -13,27 +13,46 @@
...
     Managing processor resources
     """
-    def __init__(self):
+    def __init__(self, xdb_config_home=None, xdb_data_home=None):
         self.log = getLogger('ocrd.resource_manager')
         self.database = {}
+
+        self.xdb_data_home = xdb_data_home
+        if not self.xdb_data_home:
+            self._init_data_home()
```
kba, line 29: This could probably be solved in a property:

```
@property
def xdg_home(self):
  if not gettattr(self, '_xdg_home', None):
        the_home = path.expanduser('~')
        if 'HOME' in environ and environ['HOME'] != path.expanduser('~'):
            the_home = environ['HOME']
        self._xdg_home = the_home
  return self._xdg_home
```

etc.

ocrd/ocrd/resource_manager.py
```diff
@@ -13,27 +13,46 @@
...
-
 from .constants import RESOURCE_LIST_FILENAME, RESOURCE_USER_LIST_COMMENT
 
 class OcrdResourceManager():
 
     """
     Managing processor resources
     """
-    def __init__(self):
+    def __init__(self, xdb_config_home=None, xdb_data_home=None):
```
kba, line 23: `s,xdb,xdg,g` :)

kba (changes requested):
Thanks again! You're probably right, it's best to inject these into the object rather than relying on "constants" that need to be changed for testing. The final testing code is much more readable like this. I also discovered `monkeypatch` today, this is extremely useful, e.g. for mocking `sys.stdin` etc. Looking forward to using that once we completely migrated to pytest :)

ocrd/ocrd/resource_manager.py
M3ssman: If you like to, it'll we done this way

kba: If I can choose, then yes please :) I really like `@property`, they combine the advantages of flat attributes with  methods and are great for encapsulating initialization.

M3ssman: Turned `xdg_data_home` into a property, but not `xdg_config_home` since this information is already required by the resource_manager at initialization time for user_list.

kba: OK, but why not use the property in the constructor? For consistency I added `userdir` (your `the_home` variable) and `xdg_config_home` to the constructor kwargs. Spotted a bug already (database not reloaded after adding to it) - hooray for testing :)

kba (approved):

bertsky:
@M3ssman @kba I'm not sure if it was this or one of the other merged pytest PRs, but I am having a regression when testing on a local system with `~/ocrd_logging.conf` in place. (Related to #679 I think...)

M3ssman:
Must confess I usually do not have an account-global logging configuration, I pass it straight into the working directory for each OCR-D-Workflow. Please can you provide some more details on how the regression turns out and how one might reproduce this?

bertsky:
Of course: take the default `ocrd_utils/ocrd_logging.conf` and modify it in some places, say by defining a logger for `qualname=PIL` and `level=ERROR` (which will fail at https://github.com/OCR-D/core/blob/9567190aaceba7f6d24d169633d698382065533a/tests/test_decorators.py#L64), or by setting the default `consoleHandler` to `sys.stdout` instead of stderr.

M3ssman:
Thanks! Plz give me some time to take a closer look to the mentioned test module `test_decorators.py`.

Issue #769, opened 2021-12-20
ocrd workspace bulk-add --help: code formatting broken


kba:
```
  \b Example:     ocrd workspace bulk-add \\             --regex
  '^.*/(?P<fileGrp>[^/]+)/page_(?P<pageid>.*)\.(?P<ext>[^\.]*)$' \\
  --file-id 'FILE_{{ fileGrp }}_{{ pageid }}' \\             --page-id
  'PHYS_{{ pageid }}' \\             --file-grp "{{ fileGrp }}" \\
  --url '{{ fileGrp }}/FILE_{{ pageid }}.{{ ext }}' \\
  path/to/files/*/*.*
```

is hard to read

bertsky:
duplicate of #641 

(I wonder why our `\b` does not work here BTW...)

kba:
Because the docstring has a `r` prefix, fixed in 3bfcc7ee as part of fixing #754

Issue #754, opened 2021-12-02
workspace bulk-add: allow parsing stdin instead of filesystem paths


bertsky:
The real problem here is that `bulk-add` needs strict file path names, though. But which page ID and output file ID you get depends on lots of factors (including naming convention on the input side and potential clashes with existing IDs). It is not realistic to reconstruct all information (file ID and page ID) from the file path name alone.

It would be nice if we had a function in core where you can pipe these parameter sets via stdin. Say:

    { echo PHYS_0001 BIN FILE_0001_BIN.IMG-wolf BIN/FILE_0001_BIN.IMG-wolf.png image/png; \
      echo PHYS_0001 BIN FILE_0001_BIN BIN/FILE_0001_BIN.xml application/vnd.prima.page+xml; \
      echo PHYS_0002 BIN FILE_0002_BIN.IMG-wolf BIN/FILE_0002_BIN.IMG-wolf.png image/png; \
      echo PHYS_0002 BIN FILE_0002_BIN BIN/FILE_0002_BIN.xml application/vnd.prima.page+xml; \
    } | ocrd workspace bulk-add -r "(?P<pageid>.*) (?P<filegrp>.*) (?P<fileid>.*) (?P<url>.*) (?P<mimetype>.*)" \
      -G "{{ filegrp }}" -g "{{ pageid }}" -i "{{ fileid }}" -m "{{ mimetype }}" -u "{{ url }}" -

_Originally posted by @bertsky in https://github.com/OCR-D/ocrd_olena/issues/77#issuecomment-984986141_

bertsky:
Related to #641 (we need both, and the implementation must be generic enough to allow for that).

Issue #641, opened 2020-11-12
bulk-add: improve usability


bertsky:
IMHO `ocrd workspace bulk-add` can be made more user-friendly:

1. Do _not_ convert the `FILE_GLOB` argument (which could be an actual glob, or a shell-expanded list of files) from relative path names to absolute. With absolute path names there is always the need to remove the right prefix within `--regex`. But with relative path names the regex could just start with it, which is more intuitive and less error-prone (as no directory prefix needs to be matched).
1. Do _not_ make the `--url` option required. For the special case when the `add_file` should not copy files into the workspace physically, but only add them to the METS wherever they are, this would save time and avoid errors (because when the URL does not match exactly, files will be copied, which might be surprising and leave a mess in the directory).
1. Do _not_ require a `--file-id` argument. Often it is good enough to just generate some unique ID automatically.
1. Fix the docstring such that newlines don't get removed. It is currently printed as:
```
 \b Example:     ocrd workspace bulk-add \\             --regex
  '^.*/(?P<fileGrp>[^/]+)/page_(?P<pageid>.*)\.(?P<ext>[^\.]*)$' \\
  --file-id 'FILE_{{ fileGrp }}_{{ pageid }}' \\             --page-id
  'PHYS_{{ pageid }}' \\             --file-grp "{{ fileGrp }}" \\
  --url '{{ fileGrp }}/FILE_{{ pageid }}.{{ ext }}' \\
  path/to/files/*/*.*
```

So all in all, this example could be simplified to:
```
 Example:
   ocrd workspace bulk-add \\
      --regex '^(?P<fileGrp>[^/]+)/page_(?P<pageid>.*)\.(?P<ext>[^\.]*)$' \\
      --page-id 'PHYS_{{ pageid }}' \\
      --file-grp "{{ fileGrp }}" \\
   OCR-D-IMG/page_*.*
```

kba:
Agree on all your proposed changes.

bertsky:
Also, there should be some sensible error message for non-matches. Currently, there is no telling what happened:
```
  File "/home/scott/src/github/OCR-D/ocrd_all/venv/lib/python3.6/site-packages/ocrd/cli/workspace.py", line 300, in workspace_cli_bulk_add
    file_dict[param_name] = file_dict[param_name].replace('{{ %s }}' % group_name, group_dict[group_name])
AttributeError: 'NoneType' object has no attribute 'replace'
```

bertsky:
[...]

PR #776, opened 2021-12-21
ocrd workspace bulk-add improvements

kba:
This implements the behaviors discussed in #641, #754 and #769:

* Support parsing the "file path" as a more complex list of values so that the "file path" can actually be "file path + url + mimetype + ..."
* If `FILE_GLOB` argument is `-` (single dash), read the data from STDIN
* `--file-id` is now optional, generated from the file path if omitted
* If `--url` is omitted, assume that the source file path is to be the URL of the OcrdFIle
* Provide more descriptive Exception when encountering empty/unset CLI options
* Do not convert any file paths to absolute to simplify regexes

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -241,16 +241,18 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, ignore, check_
 @click.option('-u', '--url', help="local filesystem path in the workspace directory (copied from source file if different)", required=True)
 @click.option('-G', '--file-grp', help="File group USE of the file", required=True)
 @click.option('-n', '--dry-run', help="Don't actually do anything to the METS or filesystem, just preview", default=False, is_flag=True)
+@click.option('-S', '--source-path', help="File path to copy from", default='{{ src }}')
 @click.option('-I', '--ignore', help="Disable checking for existing file entries (faster)", default=False, is_flag=True)
 @click.option('-f', '--force', help="Replace existing file entries with the same ID (no effect when --ignore is set, too)", default=False, is_flag=True)
 @click.option('-s', '--skip', help="Skip files not matching --regex (instead of failing)", default=False, is_flag=True)
 @click.argument('file_glob', nargs=-1, required=True)
 @pass_workspace
-def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp, dry_run, file_glob, ignore, force, skip):
+def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp, dry_run, file_glob, source_path, ignore, force, skip):
     """
     Add files in bulk to an OCR-D workspace.
 
-    FILE_GLOB can either be a shell glob expression or a list of files.
+    FILE_GLOB can either be a shell glob expression or a list of files or '-'
+    in which case file paths are read from STDIN.
 
     --regex is applied to the absolute path of every file in FILE_GLOB and can
     define named groups that can be used in --page-id, --file-id, --mimetype, --url and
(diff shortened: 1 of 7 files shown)


ocrd/ocrd/cli/workspace.py
```diff
@@ -229,42 +229,57 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, ignore, check_
...
+    """
     Add files in bulk to an OCR-D workspace.
 
-    FILE_GLOB can either be a shell glob expression or a list of files.
+    FILE_GLOB can either be a shell glob expression or a list of files or '-'
+    in which case file paths are read from STDIN.
 
     --regex is applied to the absolute path of every file in FILE_GLOB and can
     define named groups that can be used in --page-id, --file-id, --mimetype, --url and
     --file-grp by referencing the named group 'grp' in the regex as '{{ grp }}'.
```
bertsky, line 259:
```suggestion
    --file-grp options, e.g. by referencing the group name 'grp' from the regex as '{{ grp }}'.
```

ocrd/ocrd/cli/workspace.py
```diff
@@ -229,42 +229,57 @@ def workspace_add_file(ctx, file_grp, file_id, mimetype, page_id, ignore, check_
...
 
+    If the file paths aren't really file paths but CSV data that *contain* a
+    filename, the part that contains the actual filename should be in the named
+    group 'src'. If not explicitly defined, 'src' is equal to the file path.
```
bertsky, line 263:
```suggestion
    If the FILE_GLOB expressions do not denote the file names themselves
    (but arbitrary strings for --regex matching), then use --source-path to set
    the actual file paths to use. (This could involve fixed strings or group references.)
```

ocrd/ocrd/cli/workspace.py
```diff
@@ -302,20 +331,36 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
...
                 file_dict[param_name] = file_dict[param_name].replace('{{ %s }}' % group_name, group_dict[group_name])
 
-        # copy files
-        if file_dict['url']:
-            urlpath = Path(workspace.directory, file_dict['url'])
-            if not urlpath.exists():
-                log.info("cp '%s' '%s'", file_path, urlpath)
+        # Where to copy from
+        if 'src' in group_dict:
+            srcpath = Path(group_dict['src'])
```
bertsky, line 349:
```suggestion
        if file_dict['src']:
            srcpath = Path(file_dict['src'])
```

bertsky (changes requested):
Great stuff!

There was a discrepancy between your solution based on an implicit `src` group name vs explicit `--source-path` setting. I assumed you started with the former but then decided for the latter. In hindsight, I'm not sure which way around. (But I like the latter option better, hence my suggestions to make it work. If you do prefer the implicit `src` group name, then we should delete the new `--source-path` option and adapt the help texts.)

ocrd/ocrd/cli/workspace.py
```diff
@@ -277,8 +292,16 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
         sys.exit(1)
 
     file_paths = []
-    for fglob in file_glob:
-        file_paths += [Path(x).resolve() for x in glob(fglob)]
+    from_stdin = file_glob == ('-',)
+    if from_stdin:
+        file_paths += [Path(x.strip()) for x in sys.stdin.readlines()]
```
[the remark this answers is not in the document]
kba: Trailing whitespace in filenames? Wild. :-) But okay, these probably do exist.

ocrd/ocrd/cli/workspace.py
```diff
@@ -288,10 +309,16 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
...
-            log.error("File not matched by regex: '%s'" % file_path)
+            log.error("File '%s' not matched by regex: '%s'" % (file_path, regex))
             sys.exit(1)
         group_dict = m.groupdict()
 
+        # derive --file-id from filename if not --file-id not explicitly set
+        if not file_id:
+            file_id = safe_filename(str(file_path))
+
+        # if not url
```
mweidling, line 320: Is this a placeholder for a TODO?

kba: good catch, looks like an artefact of refactoring, should be removed.

ocrd/ocrd/cli/workspace.py
kba: This won't work because `src` is not a key in the original definition of `file_dict`, so it won't be set by the loop before this. I am also missing the commit that wired the `source_path` kwarg, apparently I forgot to push it and/or wasn't on a branch when I did it :(

kba:
> There was a discrepancy between your solution based on an implicit src group name vs explicit --source-path setting. I assumed you started with the former but then decided for the latter. In hindsight, I'm not sure which way around. (But I like the latter option better, hence my suggestions to make it work. If you do prefer the implicit src group name, then we should delete the new --source-path option and adapt the help texts.)

Yeah, it should be the latter, I only notice now that this wasn't pushed. Redoing it.

Issue #750, opened 2021-12-01
list_all_resources does not handle subdirectories well


bertsky:
In `ocrd_utils.list_all_resources` (used by a processor's `--list-resources` and resmgr's `--list-installed`), entries that are themselves directories do not get traversed recursively. 

That's okay if the processor takes directory resources (e.g. ocrd-calamari-recognize's h5/json checkpoint dirs), but not if its resources are files in subdirectories (e.g. ocrd-tesserocr-recognize's `script/*.traineddata` or `configs/*`).

So IMO a better behaviour would be to list all such directories recursively.

bertsky:
On the other hand, it will be very difficult to discern the two cases, especially during `ResourceManager.add_to_user_database` (i.e. unknown resources). 

Perhaps we should entirely rule out Tesseract-style subdirectory resources by definition (and prevent such installations during `make install-models`)?

bertsky:
But even then, the current implementation of `--show-resource` does not handle the directory case at all. It insists on `is_file=True`, and cannot find any. (Files are not handled correctly, too – but that's another story, see #752.)

I think we should not dictate the file status a priori, only its existence. Then, for directories, one could list their contents.

krvoigt:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/OCR-D-619ce2730fdbba0010685240/issues/OCR-D/core/750?planning-poker) with ZenHub @kba @lena-hinrichsen

kba:
> So IMO a better behaviour would be to list all such directories recursively.

Would we really need full recursion to support tesseract-like subdirectories? From what I have seen, there isn't any deeper nesting than 1, e.g. `script/Fraktur` or `configs/alto`?

kba:
One way to reliably solve this and not output folders for processors that only accept files and vice versa would be to add a `resources_are_files` boolean to the ocrd-tool.json (which would be `false` e.g. for calamari or eynollah)` and prune the candidates based on that. Or are there any processors (or possible fututure processors) that require both? It won't help to add this info to the resource entries in the `resource_list.yml`, since we do not know about the resources before we're searching for them (or I might overlook a solution here).

kba:
> But even then, the current implementation of `--show-resource` does not handle the directory case at all. It insists on `is_file=True`, and cannot find any

What should we show when calling `--show-resource` on a directory? Tar-gzipped folder contents?

bertsky:
[...]

PR #777, opened 2021-12-21
Show resource subdir

kba:
This partially implements #750,. `ocrd-processor --list-resources` will show directories as well as files and will traverse one level down from the processor data location to support subdirectories like in tesseract (`script/Fraktur.traineddata` etc.)

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -104,7 +104,7 @@ class Processor():
                 print(res)
             return
         if show_resource:
-            res_fname = list_resource_candidates(ocrd_tool['executable'], show_resource, is_file=True)
+            res_fname = list_resource_candidates(ocrd_tool['executable'], show_resource)
             if not res_fname:
                 initLogging()
                 logger = getLogger('ocrd.%s.__init__' % ocrd_tool['executable'])
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -27,6 +27,7 @@ from ocrd_utils import (
     list_resource_candidates,
     pushd_popd,
     list_all_resources,
+    are_processor_resources_directories,
 )
 from ocrd_validators import ParameterValidator
 from ocrd_models.ocrd_page import MetadataItemType, LabelType, LabelsType
@@ -99,16 +100,11 @@ class Processor():
             print(json.dumps(ocrd_tool, indent=True))
             return
         if list_resources:
-            for res in list_all_resources(ocrd_tool['executable']):
+            for res in list_all_resources(ocrd_tool['executable'], is_dir=are_processor_resources_directories(None, ocrd_tool)):
                 print(res)
             return
         if show_resource:
-            # XXX if a processor has at least one parameter with
-            # `content-type == text/directory`, assume ALL file parameters
-            # point to directories, not files
-            has_at_least_one_directory_parameter = next((True for p in ocrd_tool['parameters'].values()
-                if 'content-type' in p and p['content-type'] == 'text/directory'), False)
-            is_dir = has_at_least_one_directory_parameter
+            is_dir = are_processor_resources_directories(None, ocrd_tool)
             is_file = not(is_dir)
             res_fname = list_resource_candidates(ocrd_tool['executable'], show_resource, is_file=is_file, is_dir=is_dir)
             if not res_fname:
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -116,8 +116,16 @@ class Processor():
                 logger = getLogger('ocrd.%s.__init__' % ocrd_tool['executable'])
                 logger.error("Failed to resolve %s for processor %s" % (show_resource, ocrd_tool['executable']))
             else:
-                with open(res_fname[0], 'rb') as f:
-                    copyfileobj(f, sys.stdout.buffer)
+                fpath = Path(res_fname[0])
+                if fpath.is_dir():
+                    with pushd_popd(fpath):
+                        fileobj = io.BytesIO()
+                        with tarfile.open(fileobj=fileobj, mode='w:gz') as tarball:
+                            tarball.add('.')
+                        fileobj.seek(0)
+                        copyfileobj(fileobj, sys.stdout.buffer)
+                else:
+                    sys.stdout.buffer.write(fpath.read_bytes())
             return
         self.ocrd_tool = ocrd_tool
         if show_help:
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -27,7 +27,7 @@ from ocrd_utils import (
     list_resource_candidates,
     pushd_popd,
     list_all_resources,
-    are_processor_resources_directories,
+    get_processor_resource_types
 )
 from ocrd_validators import ParameterValidator
 from ocrd_models.ocrd_page import MetadataItemType, LabelType, LabelsType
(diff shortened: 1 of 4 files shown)


ocrd/ocrd/processor/base.py
```diff
@@ -104,7 +104,7 @@ def __init__(
                 print(res)
             return
         if show_resource:
-            res_fname = list_resource_candidates(ocrd_tool['executable'], show_resource, is_file=True)
+            res_fname = list_resource_candidates(ocrd_tool['executable'], show_resource)
```
bertsky, line 107: This does not work, though. Just removing `is_file` will yield _empty_ results, because we still have:
https://github.com/OCR-D/core/blob/abaa6b2ad5eff549341fdc650ac2c952f7b20e2f/ocrd_utils/ocrd_utils/os.py#L74-L78

bertsky (changes requested):
IMO we should start the fix by implementing a proper file vs directory distinction in ocrd-tool.json, then we can rest on that information everywhere else.

ocrd/ocrd/processor/base.py
kba: Right, that was too hasty :/

kba: Please have another look, I've added a function (`are_processor_resources_directories`) that checks whether at least one of the resource parameters of a processor has `content-type == 'text/directory` and use that to set the `is_dir` flag to `list_all_resources`.

I am still not 100% happy with this but at least we now 

*  get only directories for `ocrd-eynollah-segment -L` (with https://github.com/qurator-spk/eynollah/pull/61)
* get only directories for `ocrd-tesserocr-recognize -L`, including in subdirectories
* have `ocrd-eynollah-segment -C default` print a tarball of the directory to stdout

ocrd_utils/ocrd_utils/os.py
```diff
@@ -90,15 +94,36 @@ def list_all_resources(executable):
...
-    return [x.path for x in candidates]
+            if Path(processor_path).is_dir():
+                candidates += Path(processor_path).iterdir()
+    datadir = Path(XDG_DATA_HOME, 'ocrd-resources', executable)
+    if datadir.is_dir():
+        candidates += datadir.iterdir()
+    systemdir = Path('/usr/local/share/ocrd-resources', executable)
+    if systemdir.is_dir():
+        candidates += systemdir.iterdir()
+    # recurse once
```
mweidling, line 105: Why? (I don't get why this comment is useful, but this might be just me not understanding the code in depth.)

kba: I always find it useful to comment where recursion is happening because that is often a source of unexpected behavior.

In this case, we recurse only one level to support tesseract-style `scripts/Fraktur.traineddata` subdirectories. If we later decide we need full recursion, this is the place to adapt.

ocrd_utils/ocrd_utils/os.py
```diff
@@ -90,15 +94,36 @@ def list_all_resources(executable):
...
+    for parent in candidates:
+        if parent.is_dir():
+            candidates += parent.iterdir()
+    if is_dir:
+        candidates = [x for x in candidates if x.is_dir()]
+    else:
+        candidates = [x for x in candidates if not x.is_dir()]
```
bertsky, line 112: Like I said, I would find it better to check the file type on a candidate by candidate basis (so we don't need to assume file and directory parameters never co-occur in a tool's params – as they do in ocrd-calamari-recognize's `checkpoint` vs `checkpoint_dir` params). The candidates could be returned on the condition that `x.exists()` here, and in `show_resource`, one would only have to check `is_dir` again to discern the two cases.

bertsky: Couldn't we look into the tool json once, and get back with a boolean tuple `dir_res, file_res` (so either or both or none can be true), and then list accordingly? (Since you have already separated the question of whether to recurse...)

kba: You mean "only output directories iff all file parameters are directories"? This would effectively mean the status quo (list all files and folders) if a procesor has both file and directory resources. Which affects only ocrd_calamari AFAICT.

But sure, it's at least less hacky than checking if any parameter is a directory parameter and potentially not listing relevant resources.

bertsky: No, I meant
- "output directories under resource paths if any file parameters are directories"
- "output files under resource paths if any file parameters are files"
(either of which could be empty or none or both)

ocrd_utils/ocrd_utils/os.py
mweidling: Ah, okay. Do you think this is a kind of implicit knowledge that should be added?

kba: Sure, feel free to propose a better wording.

ocrd_utils/ocrd_utils/os.py
kba: Gotcha 229925d5 replaces `are_processor_resources_directories` with `get_processor_resource_types` which returns a tuple `has_dirs, has_files` and then we filter based on that after getting the list of candidates.

This works nicely but has the drawback that due to the selection criteria (not all file parameters have `content-type`), it breaks e.g. `ocrd-tesserocr-recognize -L`. Adding a `content-type` to `ocrd-tesserocr-recognize`'s `model` parameter fixes that.

But since we will have to do work on the various ocrd-tool.json to decentralize resource lists anyway, I think we should really fix the processors in this regard, not work around it in core.

bertsky: Agreed!

ocrd_utils/ocrd_utils/os.py
```diff
@@ -106,24 +102,24 @@ def list_all_resources(executable, is_dir=False):
...
     if not ocrd_tool:
+        # if the processor in question is not installed, assume both files and directories
         if not which(executable):
-            return False
+            return (True, True)
         result = run([executable, '--dump-json'], stdout=PIPE, check=True, universal_newlines=True)
         ocrd_tool = loads(result.stdout)
-    return next((True for p in ocrd_tool['parameters'].values() if 'content-type' in p and p['content-type'] == 'text/directory'), False)
+    has_dirs = next((True for p in ocrd_tool['parameters'].values() if 'content-type' in p and p['content-type'] == 'text/directory'), False)
+    has_files = next((True for p in ocrd_tool['parameters'].values() if 'content-type' in p and p['content-type'] != 'text/directory'), False)
```
bertsky, line 121: Perhaps we should reformulate this so that lazy processors that do not use `content-type` for its file resources at all are assumed to be file-only:
```suggestion
    has_dirs = next((True for p in ocrd_tool['parameters'].values() if p.get('content-type', None) == 'text/directory'), False)
    has_files = next((True for p in ocrd_tool['parameters'].values() if p.get('content-type', None) != 'text/directory'), False)
```

kba: No, wait that does not work, because then `has_files` is basically always true because the absence of a `content-type` on **any** parameters makes it true.

kba: 86c9e290

ocrd_utils/ocrd_utils/os.py
```diff
@@ -118,7 +118,7 @@ def get_processor_resource_types(executable, ocrd_tool=None):
         result = run([executable, '--dump-json'], stdout=PIPE, check=True, universal_newlines=True)
         ocrd_tool = loads(result.stdout)
     has_dirs = next((True for p in ocrd_tool['parameters'].values() if p.get('content-type', None) == 'text/directory'), False)
-    has_files = next((True for p in ocrd_tool['parameters'].values() if p.get('content-type', None) != 'text/directory'), False)
+    has_files = next((True for p in ocrd_tool['parameters'].values() if 'content-type' in p and p['content-type'] != 'text/directory'), False)
```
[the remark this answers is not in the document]
kba: Boolean logic is surprisingly hard at times :) But your proposal won't help either because the `format: uri` (or `format: file`) are not what is required. As the spec states

> To mark a parameter as expecting the address of a file, it must declare the content-type property as a valid media type.

`content-type` actually is the best indicator that this is a resource parameter. So no way around fixing the ocrd-tool.json of the processors AFAICT. Do you see another solution?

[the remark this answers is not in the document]
kba: E.g. in ocrd-tesserocr-recognize, `model` is just

```
  "model": {
   "type": "string",
   "description": "The tessdata text recognition model to apply (an ISO 639-3 language specification or some other basename, e.g. deu-frak or Fraktur)."
  },
```

(so we did not eat our own dog food and now it's coming back to bite us).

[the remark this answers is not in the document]
kba: 834b4d16 I really want this done, so I added a workaround:

```
if not next((True for p in ocrd_tool['parameters'].values() if 'content-type' in p), False):
    # None of the parameters for this processor are resources (or not
    # the resource parametrs are not properly declared, so output both
    # directories and files
    return (True, True)
```

This should ensure that the current behavior of listing all the files/directories is retained if a processor does not define resource parameters. That should not be the case, because there should be no files then but better to be safe. Once we fix the ocrd-tool.json, the result will be properly filtered.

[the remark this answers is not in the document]
bertsky: Sounds like a reasonable compromise for now and way forward. And since it's already merged, I won't puzzle over its ramifications, but observe :-)

PR #779, opened 2021-12-29
[test][rfct] propagate pytest

M3ssman:
Here are some more refactorings for both `test_ocrd_page` and `test_workspace` modules.
The major challenge was to get where paths-n-files actually are located, which is a quite hard for someone not familiar with the `pushd_popd` functionalities. Further, I divided test cases where beside regular assertions also exceptions were risen to have the exception handling being placed in it's own, separate test case.

One point that still puzzles me is the behavior of the `ocrd_workspace` `remove_...`- calls with the `force` parameter for invalid data (i.e., data which doesn't exist at all). I know we had already discussions about this, but still, I don't get it, since one might check the returned value from these calls, which in will result into 'None' for invalid data but are considered success. 

Please take into account all the `TODO` marks I've placed. They can be dropped if the reasons clear out.


kba:
> One point that still puzzles me is the behavior of the `ocrd_workspace` `remove_...`- calls with the `force` parameter for invalid data (i.e., data which doesn't exist at all). I know we had already discussions about this, but still, I don't get it, since one might check the returned value from these calls, which in will result into 'None' for invalid data but are considered success.

There's different conditions here that are relevant for `remove_file`:

1. Does the `mets:file` referenced by the kwargs of `remove_file` exist or not in the METS XML?
2. Does the local file in the file system exist

With `force=True`, the method will succeed if either of these is `False`.

And for `remove_file_group`:

1. Does the fileGrp exist or not in the METS-XML
2. Do all the files match the conditions for `remove_file`?

With `force=True`, will succeed if either of these is `False`.

Without `force`, you might be left with inconsistent data, e.g. an expected file to be deleted was not there (which might be because using the wrong `mets:file/@ID` for removal) or one of the AlternativeImages was not available. And for `remove_file`, without `force`, might lead to a half-deleted file group because one of the files in it was not found.

Would separating these aspects of concerns into different (and less generically named) kwargs help?

M3ssman:
If there's a fight on two fronts behind the scenes, I guess this can be reflected by a *attenzione:just proposing* returned tuple of two numbers, where the first counts deleted Metadata-structures an the 2nd counts deleted physical files - both should match in an ideal world.

The use case with the half-life files because of a broken removal and therefore missing or ghost items sounds very reasonable. When processing large batches in automated fashion, this is a rather familiar scenario. 

Instead of modifying the API, I prefer to return more meaningful responses. 
If you can get along with this, too, I'll do an update.

M3ssman:
This holds for removals of complete `mets:fileGrp` as well, where the behavior could even reflect _how many_ single files belonging to this group were dropped. One might even think of an inconsistent situation, where x files are referenced in the METS, but actually y files do exists physically.

kba:
I am still not 100% sure wether I understand how conditions 2) and 4) might arise - if a file is not referenced in the metadata, it will never be deleted. We don't delete files not related to the workspace, at least we should not.

But you seem to have a clear idea on how you'd go about implementing it and having a more expressive return value instead of a void method is indeed backwards-compatible, so by all means, go ahead, it probably becomes clearer with a (proof-of-concept) implementation :+1:

tests/model/test_ocrd_page.py
```diff
@@ -1,4 +1,8 @@
-from tests.base import TestCase, main, assets, create_ocrd_file_with_defaults
+# -*- coding: utf-8 -*-
```
kba, line 1: UTF-8 source encoding is the [default in Python 3](https://www.python.org/dev/peps/pep-3120/), so this is not necessary anymore.

tests/model/test_ocrd_page.py
```diff
@@ -49,297 +53,415 @@
...
+    https://github.com/OCR-D/core/pull/479
+    https://github.com/OCR-D/core/issues/240#issuecomment-493135797
+    """
+
+    # arrange
+    with open(assets.path_to('gutachten/data/TEMP1/PAGE_TEMP1.xml'), 'r') as f:
+        pg = parseString(f.read().encode('utf8'), silence=True).get_Page()
+
+    # assert
+    assert len(pg.get_AllRegions(order='reading-order-only')), 40
```
kba, line 293:
```suggestion
    assert len(pg.get_AllRegions(order='reading-order-only')) == 40
```

Or is this another pytest-specific syntax for `assert`?

tests/test_workspace.py
```diff
@@ -25,334 +34,510 @@
...
+    added_path = Path(join(plain_workspace.directory, added_res))
+
+    # assert
+    assert added_path.exists()
+    plain_workspace.remove_file_group('FOO', recursive=True)
+
+
+@pytest.fixture(name='kant_complex_workspace')
+def _fixture_kant_complex(tmp_path):
+    shutil.copytree(assets.path_to('kant_aufklaerung_1784-complex/data'), str(tmp_path), dirs_exist_ok=True)
```
kba, line 344: `dirs_exist_ok` was introduced in Python 3.8, but we need to support 3.7 for a while longer.

kba (changes requested):
🎉 Great work! Let's just make sure that all asserts are actually asserting and it runs in Python 3.7.

tests/test_workspace.py
kba: And we should phase out 3.6 as soon as possible, that has reached end of life now.

kba: I added a workaround in ced76d05

tests/model/test_ocrd_page.py
kba: corrected in 55fd25a5

kba (approved):

PR #784, opened 2022-01-24
bashlib parse_argv: must validate with params

bertsky:
This is needed for processors that have _required_ parameters – otherwise the ocrd__wrapped `ocrd validate tasks` will fail.

### DIFF
--- a/ocrd/bashlib/src/parse_argv.bash
+++ b/ocrd/bashlib/src/parse_argv.bash
@@ -78,7 +78,7 @@ ocrd__parse_argv () {
     if [[ -n "${ocrd__argv[page_id]:-}" ]]; then
         _valopts+=( --page-id "${ocrd__argv[page_id]}" )
     fi
-    _valopts+=( "${OCRD_TOOL_NAME#ocrd-} -I ${ocrd__argv[input_file_grp]} -O ${ocrd__argv[output_file_grp]} ${__parameters[*]} ${__parameter_overrides[*]}" )
+    _valopts+=( "${OCRD_TOOL_NAME#ocrd-} -I ${ocrd__argv[input_file_grp]} -O ${ocrd__argv[output_file_grp]} ${__parameters[*]@Q} ${__parameter_overrides[*]@Q}" )
     ocrd validate tasks "${_valopts[@]}" || exit $?
 
     # check parameters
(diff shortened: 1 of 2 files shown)


bertsky:
@kba for example, the new `ocrd-page-transform -L` (from https://github.com/bertsky/workflow-configuration/commit/003ccfb372be7965186ebfa3c1d79d726765f0b1) would fail without this.

ocrd/bashlib/src/parse_argv.bash
```diff
@@ -78,7 +78,7 @@ ocrd__parse_argv () {
     if [[ -n "${ocrd__argv[page_id]:-}" ]]; then
         _valopts+=( --page-id "${ocrd__argv[page_id]}" )
     fi
-    _valopts+=( "${OCRD_TOOL_NAME#ocrd-} -I ${ocrd__argv[input_file_grp]} -O ${ocrd__argv[output_file_grp]}" )
+    _valopts+=( "${OCRD_TOOL_NAME#ocrd-} -I ${ocrd__argv[input_file_grp]} -O ${ocrd__argv[output_file_grp]} ${__parameters[*]} ${__parameter_overrides[*]}" )
```
bertsky, line 81: I am wondering whether I should not quote the parameter variables here, though: Parameter values might themselves contain white-space, which would break in the `task_sequence.parse` because the `-p` or `-P` needs exactly two arguments.

bertsky: But quoting would clog all the `-p` / `-P` arguments into one. Perhaps the only adequate solution is to use 
```suggestion
    _valopts+=( "${OCRD_TOOL_NAME#ocrd-} -I ${ocrd__argv[input_file_grp]} -O ${ocrd__argv[output_file_grp]} ${__parameters[*]@Q} ${__parameter_overrides[*]@Q}" )
```

kba:
```
Q
The expansion is a string that is the value of parameter quoted in a format that can be reused as input.
```

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

bertsky: spoiler: it does work.

Issue #742, opened 2021-11-11
bashlib: fix make lib.bash dependencies


bertsky:
IMHO it's crazy we need (a recent version of) nodejs and its packagers to rebuild bashlib. If we do require non-bash functionality, then why not rely on Python here?

Also, the recipe https://github.com/OCR-D/core/blob/d9f660ee727c5235813e7f1534e26f2bebe483d3/ocrd/bashlib/Makefile#L16-L17 is dangerous, as it creates an (empty) result even in the case of errors (and thus prevents further updates to the target). (If using a shell redirect, the makefile should at least contain `.DELETE_ON_ERROR`.)

kba:
When I started on bashlib, I was doing a lot of Node.JS and was very enamoured with my shell-based tools (shinclude, shrender, tesht etc.) and OCR-D/core was still to be a reference implementation that people might or might not use, so I didn't really take into account how easy it is for others to work with it.

I think we should just drop the shinclude mechanism and have a single big `lib.bash` file that we work on directly.

As for Node.JS: You only need that for regenerating the `make help` output (by way of `@kba/makefile-parser`) and for the continuous building of lib.bash on file change - which we wouldn't need anymore.

PR #785, opened 2022-01-28
bashlib: drop building from fragments, #742

kba:
## Background

`bashlib` (in `ocrd/ocrd/lib.bash`) is a bash library that exposes/emulates much of the Python API for use in shell scripts, so developers can create OCR-D processors with relatively few lines of shell script to wrap non-Python-based processor backends, such as the OLENA C++ tools, ImageMagick's `identify` or any other command line tool.

## Current situation

As discussed in #742, the build process for `ocrd/ocrd/lib.bash` is over-engineered and adds unnecessary burden with setting up tooling (including node.js) on developers while offering little in terms of clarity or compartmentalization (because, without the shinclude tooling properly set up, one needs to edit both the source and the built library by hand).

## Solution

So, in this PR, we get rid of the build process entirely. `ocrd/ocrd/lib.bash` is to be edited directly and the README for bashlib is now part of the top-level README.

## Testing

This change is completely transparent for end users, nothing changes for them. OCR-D/core developers working on bashlib however will not need to set up shinclude and Node.JS anymore or call `make bashlib` to regenerate `ocrd/ocrd/lib.bash`, they can edit it directly.


bertsky (commented):
IMHO keeping the `make bashlib` and only exchanging its recipe (via simple `cat` instead of `shinclude`) would have been better, because we get to keep the distinct smaller files and their git history – but I can live with that, too.

Note: this would clash with https://github.com/OCR-D/core/pull/784

mweidling (approved):

PR #792, opened 2022-02-02
resmgr download: separate name and url

### LINKED ISSUE #780: resmgr download: allow specifying name independent of URL
```
ocrd resmgr download -n ocrd-detectron2-segment ...
ocrd resmgr list-installed -e ocrd-detectron2-segment
- All_X152.yaml @ data (https://layoutlm.blob.core.windows.net/tablebank/model_zoo/detection/All_X152/All_X152.yaml)
  Found at data on 2022-01-19 12:44:17.192397
- model_final_01ca85.pkl @ data (???)
  Found at data on 2022-01-20 23:29:42.287291
- model_final.pth @ data (https://layoutlm.blob.core.windows.net/tablebank/model_zoo/detection/All_X152/model_final.pth)
  Found at data on 2022-01-19 13:13:18.793976
- DLA_mask_rcnn_R_101_FPN_3x.yaml @ data (https://github.com/hpanwar08/detectron2/raw/master/configs/DLA_mask_rcnn_R_101_FPN_3x.yaml)
  Found at data on 2022-01-20 21:23:24.611257
```

This is not helpful. I would like to be able to set a `name` on the CLI, just as I can in the central `resource_list.yml`.

bertsky:
Fixes #780

### DIFF
--- a/ocrd/ocrd/cli/ocrd_tool.py
+++ b/ocrd/ocrd/cli/ocrd_tool.py
@@ -9,7 +9,7 @@ OCR-D CLI: ocrd-tool.json management
 from json import dumps
 import codecs
 import sys
-
+import os
 import click
 
 from ocrd.decorators import parameter_option, parameter_override_option
@@ -96,15 +96,23 @@ def ocrd_tool_tool_description(ctx):
 @ocrd_tool_tool.command('list-resources', help="List tool's file resources")
 @pass_ocrd_tool
 def ocrd_tool_tool_list_resources(ctx):
-    Processor(None, ocrd_tool=ctx.json['tools'][ctx.tool_name],
-              list_resources=True)
+    class BashProcessor(Processor):
+        @property
+        def moduledir(self):
+            return os.path.dirname(ctx.filename)
+    BashProcessor(None, ocrd_tool=ctx.json['tools'][ctx.tool_name],
+                  list_resources=True)
 
 @ocrd_tool_tool.command('show-resource', help="Dump a tool's file resource")
 @click.argument('res_name')
 @pass_ocrd_tool
 def ocrd_tool_tool_show_resource(ctx, res_name):
-    Processor(None, ocrd_tool=ctx.json['tools'][ctx.tool_name],
-              show_resource=res_name)
+    class BashProcessor(Processor):
+        @property
+        def moduledir(self):
+            return os.path.dirname(ctx.filename)
+    BashProcessor(None, ocrd_tool=ctx.json['tools'][ctx.tool_name],
+                  show_resource=res_name)
 
 @ocrd_tool_tool.command('help', help="Generate help for processors")
 @pass_ocrd_tool
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -9,6 +9,7 @@ __all__ = [
     'run_processor'
 ]
 
+from pkg_resources import resource_filename
 from os.path import exists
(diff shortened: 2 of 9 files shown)


bertsky:
We should add a test for the `-n URL TOOL NAME` use-case as well (both as an unregistered resource and as URL-override).

ocrd/ocrd/cli/ocrd_tool.py
```diff
@@ -96,15 +96,23 @@ def ocrd_tool_tool_description(ctx):
 @ocrd_tool_tool.command('list-resources', help="List tool's file resources")
 @pass_ocrd_tool
 def ocrd_tool_tool_list_resources(ctx):
-    Processor(None, ocrd_tool=ctx.json['tools'][ctx.tool_name],
-              list_resources=True)
+    class BashProcessor(Processor):
+        @property
+        def moduledir(self):
+            return os.path.dirname(ctx.filename)
```
kba, line 102: But will this work with the `SHAREDIR=...` convention we use for `make install` of bashlib processors? In that case, the executable might be installed to `$HOME/.local/bin`, but the resources to `$HOME/.local/share/<projectname>`?

ocrd/ocrd/processor/base.py
```diff
@@ -213,19 +205,50 @@ def resolve_resource(self, val):
...
-        List all resources found in the filesystem
+        List all resources found in the filesystem and matching content-type by filename suffix
+        """
+        mimetypes = get_processor_resource_types(None, self.ocrd_tool)
+        for res in list_all_resources(self.ocrd_tool['executable'], moduled=self.moduledir):
+            res = Path(res)
+            if not '*/*' in mimetypes:
+                if res.is_dir() and not 'text/directory' in mimetypes:
+                    continue
+                # if we do not know all MIME types, then keep the file, otherwise require suffix match
```
kba, line 233: :+1:

ocrd_utils/ocrd_utils/os.py
```diff
@@ -98,33 +100,45 @@ def list_all_resources(executable):
...
 
 def get_processor_resource_types(executable, ocrd_tool=None):
     """
-    Determine whether a processor has resource parameters that represent
-    directories (``has_dirs``), files (``has_files``) or neither.
+    Determine what type of resource parameters a processor needs.
 
-    Returns a pair ``(has_dir, has_files)``
+    Return a list of MIME types (with the special value `*/*` to
+    designate that arbitrary files or directories are allowed).
```
kba, line 127: That is much better than `has_dirs, has_files`.

kba (approved):

ocrd/ocrd/cli/ocrd_tool.py
bertsky: Yes, it does work – as long as the resources are in the same directory as the ocrd-tool.json file (which is the case for all existing bashlib processors we have). It does not matter where the executable itself is, or which directory the resources go to (venv or XDG or whatever `make install` does).

ocrd_utils/ocrd_utils/os.py
bertsky: It's been quite a chase to get here, but yes, I think this makes sense now.

bertsky:
Note: this warrants a new minor release, because the `resmgr download` CLI has changed slightly, and so has `ocrd_utils.get_processor_resource_types`. 

Documentation will have to be updated in various places:
- changelog here
- https://github.com/OCR-D/spec/pull/190
- models.md in spec

PR #676, opened 2021-02-26
OcrdExif: replace resolution logic with identify callout

kba:
For many operations, it is critical to have an accurate number for the pixel density of an image.

The capabilities of Pillow vary with different formats and requires a lot of checking in different places for numbers and often fails.

This PR replaces the PIL-based logic to get the DPI from a CLI call to ImageMagick's `identify` tool. This means that having IM installed will be a prerequisite for all OCR-D software. Getting this running in a native Windows environment will probably more difficult. But the IMHO huge advantage is that we have reliable DPI that is at least consistent and doesn't produce as many bogus `resolution 1` results.

I also tried to implement the behavior with wand, a dedicated IM library for python in https://github.com/OCR-D/core/commit/344de3eb17cd91f19bbdbae44e4e06b6f4c88a3f That has the advantage of being less platform-dependent and cleaner than parsing `identify` output. The main drawback is that it didn't produce much more accurate results than the PIL method. But this might be because Pillow's `PIL.save()` is not only slow even when writing to an in-memory object but will not correctly serialize the pixel density. Though this could probably be circumvented by reading the file from disk as is implemented in this PR.

PR #796, opened 2022-02-10
OcrdExif: reintroduce PIL method as fallback

kba:
When we merged #676 the assumption was that OCR-D users will always have ImageMagick installed. Which is still reasonable for end users but we did not consider usage of OCR-D/core as a library in other projects. And since ImageMagick is a system dependency and cannot be installed via PyPI, we effectively require library users, at least those using the OcrdExif class, to also install ImageMagick with their system's package manager.

This PR rectifies that by re-adding the old PIL-based method as a fallback if `identify` is not in `$PATH`.

Also adds `imagemagick` as a `deps-ubuntu` dependency and also runs `make deps-ubuntu` in the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,7 @@ RUN apt-get update && apt-get -y install --no-install-recommends \
     curl \
     sudo \
     git \
+    && make deps-ubuntu \
     && pip3 install --upgrade pip setuptools \
     && make install \
     && $FIXUP \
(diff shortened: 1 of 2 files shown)


kba:
> > and also runs `make deps-ubuntu` in the Dockerfile.
> 
> Seems like you forgot to commit that latter part. Otherwise fine IMO.

Indeed I did, now added.

PR #797, opened 2022-02-13
resmgr: properly implement --overwrite, fix #690

### LINKED ISSUE #690: resmgr: download --overwrite does not work
I accidentally ran `calamari-predict` from the top-level (instead of a sub-venv), which tried to update my GT4HistOCR model to the more recent TF – overwriting the checkpoint JSON.

Now, I tried to download that model again:
```shell
ocrd resmgr download --overwrite ocrd-calamari-recognize qurator-gt4histocr-1.0
INFO ocrd.cli.resmgr - Downloading resource {'url': 'https://qurator-data.de/calamari-models/GT4HistOCR/2019-12-11T11_10+0100/model.tar.xz', 'type': 'tarball', 'name': 'qurator-gt4histocr-1.0', 'description': 'Calamari model trained with GT4HistOCR', 'size': 90275264, 'path_in_archive': '.', 'version_range': '>= 1.0.0', 'parameter_usage': 'as-is'}
INFO ocrd.resource_manager._download_impl - Downloading https://qurator-data.de/calamari-models/GT4HistOCR/2019-12-11T11_10+0100/model.tar.xz to download.tar.xx
  [####################################]  100%          INFO ocrd.resource_manager.download - Extracting tarball
INFO ocrd.resource_manager.download - Copying '.' from tarball to /.local/share/ocrd-resources/ocrd-calamari-recognize/qurator-gt4histocr-1.0

  File "/venv/bin/ocrd", line 8, in <module>
    sys.exit(cli())
  File "/venv/lib/python3.6/site-packages/click/core.py", line 829, in __call__
  File "/venv/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/venv/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
  File "/venv/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
  File "/venv/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
  File "/venv/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/venv/lib/python3.6/site-packages/ocrd/cli/resmgr.py", line 137, in download
    progress_cb=lambda delta: bar.update(delta)
  File "/venv/lib/python3.6/site-packages/ocrd/resource_manager.py", line 225, in download
    copytree(path_in_archive, str(fpath))
  File "/usr/lib/python3.6/shutil.py", line 321, in copytree
    os.makedirs(dst)
  File "/usr/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/.local/share/ocrd-resources/ocrd-calamari-recognize/qurator-gt4histocr-1.0'
```

Looks like `overwrite` is not implemented properly.

kba:
Until now, the `ocrd resmgr download --overwrite` flag only had an effect if it was not set: Download would abort because the file already exists.

This PR adds an `overwrite` kwarg to the different methods handling download/copying/extraction. When set, the target directory/file are removed if they exist before data is written.


bertsky (commented):
Wouldn't it be simpler to just `rmtree(fpath)` after `resource_manager.download`'s L251 `if fpath.exists() and not overwrite:`?

kba:
> Wouldn't it be simpler to just `rmtree(fpath)` after `resource_manager.download`'s L251 `if fpath.exists() and not overwrite:`?

Simpler, yes, but I thought it would be less surprising if the `overwrite` case is handled by the download/copy method backends, because we have that pattern in other places too. You're right though, that all `overwrite=True` does is basically `rm -rf dst` before `cp/download src dst`.

bertsky:
> but I thought it would be less surprising if the `overwrite` case is handled by the download/copy method backends, because we have that pattern in other places too

I can only see downloading via `requests` in `Resolver.download_to_directory`, and in `Workspace.download_file` / `download_url`, which delegates to the former.

But even if you are planning on streamlining this into `OcrdResourceManager.download` in the future, it would still have to pass the top-level `fpath.exists()` test.

kba:
> But even if you are planning on streamlining this into `OcrdResourceManager.download` in the future, it would still have to pass the top-level `fpath.exists()` test.

9bc3aed85 Now it is done at the first `fpath.exists()` check, `overwrite` kwarg is unnecessarily convoluted.

PR #798, opened 2022-02-13
Assets change

kba:
Update the assets to include the latest fixes in the test data. This means that the SBB-sample now has fewer dangling files.

It also became apparent that the `ocrd workspace merge` functionality needs an `--overwrite` flag too in case both workspaces contain files named identically.


bertsky (approved):
Good catch!

kba:
Merged into #800

PR #805, opened 2022-02-17
[test][fix] adjust test parameter ocrd png

M3ssman:
Just adjust test parameters

### DIFF
--- a/tests/model/test_exif.py
+++ b/tests/model/test_exif.py
@@ -18,7 +18,7 @@ from ocrd_models import OcrdExif
     ('SBB0000F29300010000/data/OCR-D-IMG/FILE_0001_IMAGE.tif',
      2875, 3749, 300, 300, 300, 'inches', 'RGB', 'jpeg'),
     ('kant_aufklaerung_1784-binarized/data/OCR-D-IMG-BIN/BIN_0020.png',
-     1457, 2084, 116, 116, 116 , 'cm', '1', None),
+     1457, 2084, 1, 1, 1 , 'inches', '1', None),
     ('scribo-test/data/OCR-D-PRE-BIN-SAUVOLA/OCR-D-PRE-BIN-SAUVOLA_0001-BIN_sauvola.png',
      2097, 3062, 1, 1, 1, 'inches', '1', None),
     ('leptonica_samples/data/OCR-D-IMG/OCR-D-IMG_1555_007.jpg',


kba (approved):

kba (changes requested):
I am confused, why does this happen for you? It does not for me.

What's your ImageMagick installation? Mine:

```sh
$ identify --version

Version: ImageMagick 6.9.7-4 Q16 x86_64 20170114 http://www.imagemagick.org
Copyright: © 1999-2017 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Features: Cipher DPC Modules OpenMP 
Delegates (built-in): bzlib djvu fftw fontconfig freetype jbig jng jpeg lcms lqr ltdl lzma openexr pangocairo png tiff wmf x xml zlib
```

M3ssman:
Oh my, I wasn't aware. Just run it on 2 different machines, both Ubuntu 20.04 LTS Desktop, but:

* ImageMagick 6.9.10-23 => PR fail
* ImageMagick 7.0.9-12   => PR success, but missing png Delegate

I will investigate further ...

M3ssman:
I've got a "solution" :zany_face: 

Problem is _not_ ImageMagick 7+ itself, but I've must have installed it ages ago from source without png Delegate, which is only installed if `libpng-dev` is also available (so must be `libtiff-dev`, `libopenjp2-7-dev`,  ... ). 
Since I didn't use IM in this context, it didn't bother.

Although one could think of something like: 

```
@pytest.mark.skipif(
    'png' not in subprocess.run(['identify', '-version'], stdout=subprocess.PIPE).stdout.decode('UTF-8'), 
    reason='no PNG delegate for ImageMagick'
)
def test_ocr_exif_missing_delegate_png():
    assert ...
```
but IMHO this is overkill, since it applies to _all_ IM delegates as well.

Sorry for the inconvenience! Feel free to drop this PR! 
The delegate dependencies that might arise from custom builds should be documented in a proper place.

kba:
> I've got a "solution" 🤪
> 
> Problem is _not_ ImageMagick 7+ itself, but I've must have installed it ages ago from source without png Delegate, which is only installed if `libpng-dev` is also available (so must be `libtiff-dev`, `libopenjp2-7-dev`, ... ). Since I didn't use IM in this context, it didn't bother.
> 
> Although one could think of something like:
> 
> ```
> @pytest.mark.skipif(
>     'png' not in subprocess.run(['identify', '-version'], stdout=subprocess.PIPE).stdout.decode('UTF-8'), 
>     reason='no PNG delegate for ImageMagick'
> )
> def test_ocr_exif_missing_delegate_png():
>     assert ...
> ```
> 
> but IMHO this is overkill, since it applies to _all_ IM delegates as well.

I think unless you do `apt install --no-install-recommends`, the PNG and TIFF (but not openjp2, at least not in 18.04) should be installed together with imagemagick. Thanks for the pytest snippet though, it will come in handy later for another use case I'm sure :)

> 
> The delegate dependencies that might arise from custom builds should be documented in a proper place.

Good point. Where best should we document this you think?

M3ssman:
I guess best place is the central README. 
There's already a section regarding installation, I'll open a PR to add a subsection called preconditions or prerequisities or setup or whatever you like. There should also the interference between local log configuration and test execution stated.

kba:
> I guess best place is the central README.

Sure, PR welcome! Thank you!

PR #813, opened 2022-03-04
Remove unneeded dependency python3-pip

stweil:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ PIP_INSTALL = pip install
 
 # Dependencies for deployment in an ubuntu/debian linux
 deps-ubuntu:
-	apt-get install -y python3 python3-pip python3-venv
+	apt-get install -y python3 python3-venv
 
 # Install test python deps via pip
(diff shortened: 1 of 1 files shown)


kba (approved):

kba:
We probably even only need `python3-venv` because it implies `python3` doesn't it?

stweil:
That's right, but the superfluous `python3` won't cause installation of unneeded software (while `python3-pip` would, well, only about 1 .2 MB).

bertsky:
Wait, can you show me proof / documentation that's true? (I often see the case that `pip` or `pip3` in a venv points to the preinstalled version from the host system, not the venv.)

stweil:
I haven't installed `python3-pip` on any of our systems where I have built OCR-D (Linux, macOS, Debian, Ubuntu, x86_64, arm, aarch64, Python 3.6, 3.7, 3.8, 3.9). `pip` or `pip3` is always a small script with less than 300 bytes .

Issue #803, opened 2022-02-15
Improve usability of ocrd-dummy


kba:

kba:
> There should be a parameter disabling the copying of files, so that only the generated PAGE-XML is written out.

This has been implemented in #814 and released in 2.45.0

> ocrd-dummy is a bad name, we should call it something more intuitive like ocrd-copy or provide the non-copying-PAGE-generating functionality in a separate processor ocrd-page-from-images.

These are still open. Thinking some more about it, I think it would be best to "namespace" such builtin processors as `ocrd-util-*`. So, in this case `ocrd-util-copy`, the [post-processing/cleanup processor](https://github.com/OCR-D/core/issues/544) could be `ocrd-util-cleanup`.

PR #814, opened 2022-03-07
ocrd-dummy: make copying optional and disable by default, #803

kba:


bertsky (changes requested):
Something is rotten in the state of Denmark.

The behaviour used to be:
- for each input file representing a physical page (which could be PAGE-XML, a single image file or a single file of some other type),
  - if PAGE-XML, just update the pcGtsId and append to Metadata, writing a copy in the output fileGrp
  - otherwise, make a byte-by-byte copy in the output fileGrp; further, if it _was_ an image file, create a new PAGE-XML referencing that image file as original

We were dissatisfied with that, because copying images is generally not necessary and also not typical for OCR-D. (Images are always re-referenced by new PAGE-XML. Only new images must be added – if the processor operates on images. But ocrd-dummy is not such a processor. You can use ocrd-preprocess-image for that kind of thing.)

IMHO all that was necessary was to add a parameter `copy_files=false` and change the non-PAGE case such that the copying is dependent on that parameter (plus the image reference filename is, because you must use the new filename iff you created a new one).

However, the new behaviour here becomes:
- for each input file representing a physical page (which could be PAGE-XML, a single image file or a single file of some other type),
  - if PAGE-XML
    - **if `copy_files=true`**, then update the pcGtsId and append to Metadata, writing a copy in the output fileGrp
    - **if `copy_files=false`**, then do nothing (!)
  - otherwise,
    - **if `copy_files=true`**, then make a byte-by-byte copy in the output fileGrp
    - **if `copy_files=false`**, then do nothing
    - further, if it was an image file, create a new PAGE-XML referencing that image file as original (but always the _old_ image file!)

Why would you want to do this?

kba:
OK, I have revised the logic. Now:

* If input file is PAGE-XML:
  * set new pcGtsId
  * write output filegroup
* if input file is anything else except PAGE-XML
  * if `copy_files=true`:
    * copy input file to output filegroup, byte-by-byte
  * if input file is an image:
    * create PAGE-XML for it and reference either the original image (if `copy_files=false`), otherwise the copy we just created, in `imageFilename`.

I think that should be what you proposed and what @stefanCCS wants.

kba:
Failing CI is because macos, test passes for linux.

PR #819, opened 2022-03-16
change ocrd-anybaseocr-layout-analysis model

kba:
OCR-D/ocrd_anybaseocr#89

Since #800 still needs tests, for now in the central resource list.


bertsky (changes requested):
Note: the new model format is necessary for two processors, layout-analysis **and tiseg**. In https://github.com/OCR-D/ocrd_anybaseocr/pull/89 I have uploaded new versions for both.

Issue #687, opened 2021-05-03
image API: control order of image operations


bertsky:
Naturally, with operations like deskewing/rotation and cropping the resulting image size and relative coordinates depend on the relative order of the operations. We already respect that by following the exact order of image features in `AlternativeImage/@comments` when reconstructing coordinates.

But reconstructibility not always enough. One might need to enforce a certain ordering for producing _new_ images.

Consider for example the following realistic workflow:
1. binarization (binarizes the raw image)
2. deskewing (rotates 1.)
3. cropping (crops 2.)
4. some consumer that needs _both_ a binarized image (from 3.) and a _raw_ image, and naturally expects them to be in the same coordinate system

Here we get a problem: In 4., the request `image_from_page(page, feature_filter='binarized')` will make core produce a new image by applying cropping and deskewing to the raw image. (This is the _default ordering_ of core.) But having a different ordering, that second image will be larger and in a different coordinate system. So the consumer gets an inconsistent pair of images (which it would have to check and compensate for by looking at the two returned coordinate transforms).

There is no cheap solution here I'm afraid:
- We cannot just change the default ordering, because other workflows might be using another ordering, too.
- We cannot just look into the existing derived images and adopt their ordering, because (depending on the workflow) we would not know which of the images is the right one (and they could derive from differently ordered sub-workflows). That is, at the time the second `image_from_page` request is made, we don't know what was in the other.
- Even if we allow passing a parameter to make the caller control the ordering, the caller might forget to do that (or not be aware of the necessity, rightly expecting to get the same kind of result automatically)

So what the heck are we gonna do about this?

bertsky:
On the other hand – forgive my confusion – since d6c51ca5f40090e2a6ac305b034f124b05f4edd9 we should actually not see any difference, because we already re-crop after rotation whenever possible.

bertsky:
> On the other hand – forgive my confusion – since [d6c51ca](https://github.com/OCR-D/core/commit/d6c51ca5f40090e2a6ac305b034f124b05f4edd9) we should actually not see any difference, because we already re-crop after rotation whenever possible.

Darn! I think that commit was (well intentioned but) still flawed: when re-cropping via `_crop()`, which is the general function, it would first check whether `cropped` already applied on the image (which it is, but on an earlier state, now only half-way), and in that case only apply the corresponding coordinate transform, but not the actual image operation. So we need to bypass that feature check here.

bertsky:
[...]

PR #820, opened 2022-03-16
Fix recropping

bertsky:
Attempt to expose [the recropping problem](https://github.com/OCR-D/core/issues/687#issuecomment-1067920406) and then solve it. 

(Currently the test still succeeds, i.e. it does not correctly capture the problem. You can only see the warning `ocrd_utils.crop_image:image.py:468 crop coordinates ((0, 25, 932, 407)) exceed image (932x382)` if you run `python -m pytest -rpP tests/test_workspace.py::test_deskewing`.)

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -998,7 +998,9 @@ def _crop(log, name, segment, parent_image, parent_coords, op='cropped', **kwarg
     # crop, if (still) necessary:
     if (not isinstance(segment, BorderType) or # always crop below page level
         not op in parent_coords['features']):
-        if isinstance(segment, BorderType):
+        if op == 'recropped':
+            log.info("Recropping %s", name)
+        elif isinstance(segment, BorderType):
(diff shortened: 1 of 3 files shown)


bertsky:
> Attempt to expose [the recropping problem](https://github.com/OCR-D/core/issues/687#issuecomment-1067920406)

e615277

> and then solve it.

ac4b7f3

bertsky:
Just as an illustration of the `test_deskewing` case:
- region after deskewing but before recropping:
tmpr3ojx9hv
- region after deskewing and recropping:  tmp07p8vo3d
- region after deskewing and double recropping (which never gets exposed as image anyway, but was causing the warning): 
tmpthjwrd_s
tmpypgec_vj

kba (approved):

kba:
> Just as an illustration of the `test_deskewing` case:

These are very helpful to understand the operations - maybe we can reference them in the docstrings so it is available in the API docs? Or at least reference this issue so it will still be findable once this is merged.

PR #843, opened 2022-04-25
Remove assets server

mweidling:


kba (approved):
And we should also remove it from https://github.com/OCR-D/assets/blob/master/Makefile#L38-L41

bertsky:
What's the rationale for removing that feature?

kba:
> What's the rationale for removing that feature?

The main purpose of it was to test the on-demand downloading, I haven't used it in years, was not aware anyone was using it and it's using the old python2 SimpleHTTPServer.

bertsky:
Understood, thanks!

PR #845, opened 2022-04-28
Workspace.image_from_*: allow filtering by filename

bertsky:
Motivation for this: since we have deprecated `Workspace.resolve_image_as_pil` (because it cannot calculate the coordinate transform that would belong to the image; and we don't want processors to search for AlternativeImages from scratch), we lost a way to deal with the use-case that the caller does _not_ want to create new images dynamically, but instead already knows the desired AlternativeImage (and therefore its path).

Here we therefore (re)introduce that possibility with a kwarg `filename` for the image retrieval functions. (So the difference is that now you have to pass an image path that belongs to an existing AlternativeImage, but then you do get the coordinate transform along with the image.)

@kba I can see no harm here and have already tested locally.

### DIFF
--- a/ocrd/ocrd/workspace.py
+++ b/ocrd/ocrd/workspace.py
@@ -489,7 +489,7 @@ class Workspace():
 
     def image_from_page(self, page, page_id,
                         fill='background', transparency=False,
-                        feature_selector='', feature_filter=''):
+                        feature_selector='', feature_filter='', filename=''):
         """Extract an image for a PAGE-XML page from the workspace.
 
         Args:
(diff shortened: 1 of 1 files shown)


kba (approved):
Makes sense and since it's just an additional filter, I see no issue with compatibility.

ocrd/ocrd/workspace.py
```diff
@@ -688,6 +695,10 @@ def image_from_page(self, page, page_id,
                     fill=fill, transparency=transparency)
 
         # verify constraints again:
+        if filename and not getattr(page_image, 'filename', '').endswith(filename):
```
hnesk, line 698: page_image.filename is always empty for me, so this check always fails

bertsky: > page_image.filename is always empty for me, so this check always fails

then that AlternativeImage is not **maximal**, i.e. there are operations (derotation, cropping) that are possible from the description of the segment but have not been applied to the image yet – so the API does it automatically before returning an image. Any PIL.Image operation will invalidate the `filename` attribute, though.

Can you try retrieving the image with additional filters like `feature_filter=deskewed`?

Could you share the snipped of the segment/page that shows the AlternativeImage elements with their comments?

bertsky: I agree it's somewhat unexpected, but we have no other choice from the API point of view. Other callers would be surprised to _not_ get an image that corresponds with the features of the segment.

But we could clarify that in the docstring...

PR #856, opened 2022-05-12
OcrdMets.find_files: allow pageId regex, precompile all regexes

### LINKED ISSUE #855: OcrdMets.find_files: allow regex for pageId, or orderlabel ranges
If you want to process a range of pages, but the physical page IDs are not completely homogeneous (i.e. you have variation in both the alphabetic and numeric parts of the identifiers), then the current numerical page range syntax does not help.

But `find_files` does not allow the `//` prefix notation for regular expressions either:
```
find_files does not support regex search for pageId
```

Why not, actually?

As an alternative, one could allow passing the purely numerical index or the structMap's `@ORDERLABEL`, too.

bertsky:
Fixes #855

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+Changed:
+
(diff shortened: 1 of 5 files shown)


kba (approved):

kba:
LGTM and also good to precompile all regexes, not sure whether Python does this dynamically in all cases.

Issue #825, opened 2022-03-22
make_file_id: not correct for --overwrite


bertsky:
If the page ID is not directly contained in the file ID of the input file, then `make_file_id` determines the index of that file in the input fileGrp, and calculates a new ID based on that index and the output fileGrp. If that ID already exists, the index is incremented until a free one becomes available.

Alas,
1. the list returned by `OcrdMets.find_files` is not sorted by page ID (but `mets:file` element order)
2. the increment strategy is wrong in combination with `--overwrite`, because it will create multiple files for the same page ID – only the first one of which will be considered by follow-up processors (so nothing is actually overwritten; and the new files will be ignored entirely)

To address both issues, I suggest calculating the output ID based on the (output fileGrp and) page ID of the input file.

bertsky:
@kba this is a very nasty bug that prevents `--overwrite` for me in a lot of cases (and makes repairing the METS afterwards very hard). RFC

kba:
> RFC

Sry, this one got lost in the shuffle.

> To address both issues, I suggest calculating the output ID based on the (output fileGrp and) page ID of the input file.

Agreed in general, though we need a fallback for the case that a file has no pageId - which should not happen in real life data but is not strictly required.

I'll prepare a PR.

PR #860, opened 2022-05-13
make_file_id: use pageId for generating file ID if possible

kba:
Implements #825


ocrd_utils/ocrd_utils/str.py
```diff
@@ -61,16 +61,19 @@ def make_file_id(ocrd_file, output_file_grp):
...
+    is not unique, append a counter and increment it until it is unique.
+    Otherwise: Use the position of the ``ocrd_file`` in its fileGrp as a counter and increment it until the
+    concatenation of ``output_file_grp`` with the counter is unique
     """
     ret = ocrd_file.ID.replace(ocrd_file.fileGrp, output_file_grp)
     if ret == ocrd_file.ID:
-        if ocrd_file.pageId and ocrd_file.pageId in ocrd_file.ID:
-            # still sufficiently unique
-            ret = output_file_grp + '_' + ocrd_file.ID
+        if ocrd_file.pageId:
```
bertsky, line 71:
```suggestion
        elif ocrd_file.pageId:
```

bertsky (changes requested):
This does not address the issue, though: If you run on an already existing fileGrp, regardless of `--overwrite` or not, you will get new file IDs (for the same pageId) which will be henceforth be ignored (in the default `skip` strategy of `Processor.input_files`).

See #861 for a different (if unfinished) attempt.

But we probably also have to enforce this _outside_ of `make_file_id`, probably in `Workspace.add_file`: When adding a PAGE-XML file for a pageId that already exists in that fileGrp,
- if `overwrite`, then re-use that fileId
- otherwise raise an exception

bertsky:
So can this be closed @kba?

kba:
> So can this be closed @kba?

Yes, I just merged #861

PR #861, opened 2022-05-13
fix make_file_id

bertsky:
First we need to have a test for `--overwrite` that fails…


ocrd_utils/ocrd_utils/str.py
```diff
@@ -62,25 +62,29 @@ def make_file_id(ocrd_file, output_file_grp):
...
+        elif ocrd_file.pageId:
+            ret = output_file_grp + '_' + ocrd_file.pageId
         else:
             ids = [f.ID for f in ocrd_file.mets.find_files(fileGrp=ocrd_file.fileGrp, mimetype=ocrd_file.mimetype)]
             try:
                 n = ids.index(ocrd_file.ID) + 1
             except ValueError:
                 n = len(ids)
             ret = concat_padded(output_file_grp, n)
-            while next(ocrd_file.mets.find_files(ID=ret), None):
```
kba, line 81: Why did you remove the while loop here? This leads to ID clashes. Which is fine for the `--overwrite` case but not in a general setting, i.e.

```python
f = mets.add_file('ABC', ID="BAR_7", mimetype="image/tiff")     
self.assertEqual(make_file_id(f, 'FOO'), 'FOO_0001')            
mets.remove_file(fileGrp='FOO')                                 
self.assertEqual(make_file_id(f, 'FOO'), 'FOO_0001')            
mets.add_file('FOO', ID="FOO_0001", mimetype="image/tiff")      
self.assertEqual(make_file_id(f, 'FOO'), 'FOO_0002')            
```

fails because `make_file_id` returns `FOO_0001` instead of `FOO_0002`.

kba (commented):
If I grok this correctly, the problem is that `make_file_id` is not aware of the `overwrite_mode` of the workspace and should behave differently. It should not try to avoid clashes because that would lead to inconsistent state, i.e. more files with different IDs instead of overwriting existing files.

One solution could be to add a new kwarg `overwrite` to the `make_file_id` method. Not ideal because that requires all the procesors to be adapted and can lead to easy-to-miss inconsistencies across processors.

Since `overwrite` is a state of the `OcrdWorkspace` class and there is no link back from `OcrdMets` to the workspace it represents, we cannot go `OcrdFile` -> `OcrdMets` -> `OcrdWorkspace` - but maybe this should be possible. We could add a reference from OcrdMets to OcrdWorkspace in the latter's `__init__`.

Then we could check `OcrdFile.mets.workspace.overwrite_mode` and skip the loop to create unique IDs.

bertsky:
> If I grok this correctly, the problem is that `make_file_id` is not aware of the `overwrite_mode` of the workspace and should behave differently. It should not try to avoid clashes because that would lead to inconsistent state, i.e. more files with different IDs instead of overwriting existing files.

Yes. But also, conversely, running against an existing grp without `--overwrite` should produce clashes when it involves the existing page IDs.

(Of course, that's nothing we can ensure at the level of `make_file_id` anymore – see [my mirror review](https://github.com/OCR-D/core/pull/860#pullrequestreview-973649972)

> One solution could be to add a new kwarg `overwrite` to the `make_file_id` method. Not ideal because that requires all the procesors to be adapted and can lead to easy-to-miss inconsistencies across processors.

Yes. Possible but not a good design.

> Since `overwrite` is a state of the `OcrdWorkspace` class and there is no link back from `OcrdMets` to the workspace it represents, we cannot go `OcrdFile` -> `OcrdMets` -> `OcrdWorkspace` - but maybe this should be possible. We could add a reference from OcrdMets to OcrdWorkspace in the latter's `__init__`.
> 
> Then we could check `OcrdFile.mets.workspace.overwrite_mode` and skip the loop to create unique IDs.

Yes, but should `OcrdMets` really depend on `OcrdWorkspace`?

My best idea for the moment still is:
> But we probably also have to enforce this outside of `make_file_id`, probably in `Workspace.add_file`: When adding a PAGE-XML file for a pageId that already exists in that fileGrp,
> - if `overwrite`, then re-use that fileId
> - otherwise raise an exception

bertsky:
> IIUC you would replace
> 
> ```python
> mets_file = next(self.find_files(ID=ID), None)
> ```
> 
> with
> 
> ```python
> mets_file = next(self.find_files(fileGrp=fileGrp, pageId=pageId)
> ```
> 
> and if that is found and `force` is `True`, then change the found file accordingly but retain its old ID.

No. Like I said, keep `OcrdMets.add_file` unchanged, but before delegating to it, in `Workspace.add_file`, do:
```python
    if mimetype == MIMETYPE_PAGE:
        # only one PAGE per page
        file_ = next(self.mets.find_files(fileGrp=file_grp, pageId=kwargs['page_id'], mimetype=mimetype), None)
        if file_:
            kwargs['ID'] = file_.ID # let overwrite_mode decide what to do (re-use / raise)
    ret = self.mets.add_file(file_grp, **kwargs)
```
I'm not sure about `file_.local_filename` though: `OcrdMets.add_file` will force re-using the existing file path, but (unless we provide the `content=` anyway and don't care about the exact filename afterwards) we want our new file path to replace it.

bertsky:
Can we get this moving @kba? #825 is still quite a blow you know...

ocrd_utils/ocrd_utils/str.py
```diff
@@ -62,25 +62,29 @@ def make_file_id(ocrd_file, output_file_grp):
...
+    Else if ``ocrd_file`` has a pageId at all, then merely append it to ``output_file_grp``.
+    Otherwise use ``output_file_grp`` together with the position of ``ocrd_file`` within the input fileGrp.
+    """
+    # considerations for this behaviour:
+    # - uniqueness (in spite of different METS and processor conventions)
+    # - predictability (i.e. output name can be anticipated from the input name)
+    # - stability (i.e. output at least as much sorted and consistent as the input)
+    # ... and all this in spite of --page-id selection and --overwrite
     ret = ocrd_file.ID.replace(ocrd_file.fileGrp, output_file_grp)
     if ret == ocrd_file.ID:
```
bertsky, line 75: @kba, BTW we could add another feature easily here:
```suggestion
    if ret == ocrd_file.ID and ocrd_file.fileGrp != output_file_grp:
```

This change looks harmless, but it will allow processing _in-place_ on demand by simple passing the input fileGrp as the output fileGrp together with `--overwrite`. (That's something we don't support officially, but sometimes it is really helpful to avoid creating too many fileGrps, esp. during GT production.)

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -310,19 +310,36 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
+                        mets_file = None
+                    else:
+                        # XXX explicitly DO NOT set the ID but reuse the existing ID
+                        # XXX https://github.com/OCR-D/core/pull/861
+                        # mets_file.ID = ID
+                        mets_file.url = url
+                        mets_file.local_filename = local_filename
+                        mets_file.mimetype = mimetype
+            if not mets_file:
+                mets_file = next(self.find_files(ID=ID), None)
```
bertsky, line 334: Do we really want to steal an OcrdFile for another page or MIME or fileGrp, just because it clashes with our intended ID?

IMO this should give us an exception unconditionally.

ocrd_utils/ocrd_utils/str.py
```diff
@@ -85,6 +85,9 @@ def make_file_id(ocrd_file, output_file_grp):
             except ValueError:
                 n = len(ids)
             ret = concat_padded(output_file_grp, n)
+            while next(ocrd_file.mets.find_files(ID=ret), None):
+                n += 1
+                ret = concat_padded(output_file_grp, n)
```
bertsky, line 90: I thought we agreed we don't want to evade clashes by counting anymore?

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -310,19 +310,36 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
-            mets_file.pageId = pageId
-            mets_file.local_filename = local_filename
-        else:
+        mets_file = None
+        if not ignore:
+            if pageId:
+                mets_file = next(self.find_files(fileGrp=fileGrp, mimetype=mimetype, pageId=pageId), None)
+                if mets_file:
+                    if not force:
+                        # XXX should this be an exception?
```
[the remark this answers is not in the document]
kba: But if it is an exception here, we cannot use the second find query as a fallback. Unless we change the logic to not extend any existing OcrdFile like you propose above.

ocrd_utils/ocrd_utils/str.py
kba: OK, I think I (finally) found out why this behavior was confusing me: As a fallback for the files without pageID or where `pageId not in ocrd_file.ID`, I have used the IDs in `ocrd_file.fileGrp`, not the `output_file_grp`.

I've dropped the clash-evading clause you mention, replaced it with a FileExistsError and adapted the test in ac1cc0805, now it does make sense to me.

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
bertsky: That fallback is illegitimate just the same, though: If the ID already exists, it may be the file we want to overwrite, or it may not be (i.e. a clash by chance).

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -310,19 +310,36 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
-            mets_file.url = url
-            mets_file.mimetype = mimetype
-            mets_file.ID = ID
-            mets_file.pageId = pageId
-            mets_file.local_filename = local_filename
-        else:
+        mets_file = None
+        if not ignore:
+            if pageId:
+                mets_file = next(self.find_files(fileGrp=fileGrp, mimetype=mimetype, pageId=pageId), None)
```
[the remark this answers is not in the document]
bertsky: I gave it some more thought. The insight was right, but I am not convinced anymore that it implies we must abolish the `add_file` resolution. Here's why:

In all cases except PAGE-XML files, we can have multiple files per page/grp/type. So there we'd always lay a trap: if we intervene, we might effectively create an artificial clash (e.g. by disallowing multiple textline images); if we don't, we might miss an actual clash (e.g. by allowing two competing ALTO-XML outputs).

So page/grp/type clashes cannot be decided.

Then how about ID clashes? In the benign case, the same kind of process that is responsible for the existing ID is now repeating. So we should be fine just checking `force/overwrite` mode (hence, reuse vs. raise). In the malignant case, some other process created the same ID, despite being unrelated by the matter (i.e. different page/grp/type). Suppose that we _can_ distinguish that case, how do we proceed? 

We cannot control the process' `make_file_id` (as we are just in a subsequent/subordinate call), and we cannot rename the existing ID either (as it might cause inconsistencies elsewhere). So I'd say we are forced to capitulate (i.e. raise).

That would amount to a much simplified decision tree. Am I missing something?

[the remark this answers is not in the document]
kba: > In all cases except PAGE-XML files, we can have multiple files per page/grp/type. So there we'd always lay a trap: if we intervene, we might effectively create an artificial clash (e.g. by disallowing multiple textline images); if we don't, we might miss an actual clash (e.g. by allowing two competing ALTO-XML outputs).

> So page/grp/type clashes cannot be decided.

91e5bd48a

First of all, I moved the ID clash check in add_file to the top. If ID clashes and neither `force` nor `ignore` are active, refuse to add outright.

I also got rid of the "reuse existing OcrdFile" mechanism altogether because it offers no benefit and can lead to hard-to-debug errors. Instead, in case of clashes and with `force` active, existing files will be removed first, before being added. No change in behavior for non-clashing files.

AFAICS there are two sets of files where straightforward clash detection won't work: a) AlternativeImages (which might be many with the same pageId, mimetype and filegrp) and b) document-wide documents like TEI or PDF that do not have a pageId.

To accomodate a), I've added a new kwarg `is_alternative_image` that, if set, will ignore the "grp/mimetype/pageId" check outright. Unless the IDs are clashing, it's perfectly valid to have many files with the same mimetype in the same fileGrp for the same pageId.

The files in b) have in common that they lack a pageId. So, in case there is a "grp/mimetype" clash without `pageId` being set, nothing is removed. This does prevent e.g. having two vartiants of TEI of the whole document in the same fileGrp. So if that is a realistic prospect, it needs another round of adaptions.

I **think** the decision tree in `OcrdMets.add_file` should largely reflect your proposal, with the added complication of handling `force`, `ignore`, `is_alternative_image` and "is a file without pageId" as described.

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
kba: Yes, that was not the right solution, please have another look at the changed behavior.

ocrd_models/ocrd_models/ocrd_mets.py
kba: No, we should not reuse existing OcrdFiles and should raise immediately if ID clashes, cf 91e5bd48a

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -310,19 +310,36 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
-                raise Exception("File with ID='%s' already exists" % ID)
-            mets_file.url = url
-            mets_file.mimetype = mimetype
-            mets_file.ID = ID
-            mets_file.pageId = pageId
-            mets_file.local_filename = local_filename
-        else:
+        mets_file = None
+        if not ignore:
+            if pageId:
```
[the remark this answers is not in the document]
kba: I removed but then kind of reintroduced the pageId check but only after searching to detect whether a "grp/mimetype" clash happens for a file without pageId, in which case it is acceptable to have multiple files of the same mimetype in the same filegrp.

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
bertsky: > AFAICS there are two sets of files where straightforward clash detection won't work: a) AlternativeImages (which might be many with the same pageId, mimetype and filegrp) and b) document-wide documents like TEI or PDF that do not have a pageId.
> 
> To accomodate a), I've added a new kwarg `is_alternative_image` that, if set, will ignore the "grp/mimetype/pageId" check outright. Unless the IDs are clashing, it's perfectly valid to have many files with the same mimetype in the same fileGrp for the same pageId.

I see two problems with that:
1. you force the caller to decide in advance what the file is for. Hard to decide if you are in a shell script. (Basically, this extra arg has to be dragged through everywhere, with lots of places to adapt.)
2. there may be other multi-file-per-page cases besides derived images. Take custom formats like JSON or txt per region/line/word/glyph for example. We already have these cases in existing processors (e.g. ocrd-segment-extract-lines, ocrd-segment-replace-text).

Like I said, in all cases except PAGE-XML we simply do not know in advance whether multi-file semantics is ok or not. My formulation was more general and should not need that distinction. However, it hinges on the assumption that the malign case (ID clash from another, unrelated process) does not coincide with a grp/mime/page clash. (Which in my opinion is safe to make.)

> The files in b) have in common that they lack a pageId. So, in case there is a "grp/mimetype" clash without `pageId` being set, nothing is removed. This does prevent e.g. having two vartiants of TEI of the whole document in the same fileGrp. So if that is a realistic prospect, it needs another round of adaptions.

Yes, I agree we can assume that `pageId=None` is just as unique as any single pageId. But then you don't need the extra check. Notice this is already covered by my pseudocode (which makes for fewer conditionals).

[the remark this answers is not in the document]
bertsky: > I **think** the decision tree in `OcrdMets.add_file` should largely reflect your proposal, with the added complication of handling `force`, `ignore`, `is_alternative_image` and "is a file without pageId" as described.

If my analysis is right, then the only complication to my proposal would be the outer `if not ignore:`.

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
kba: Behavior is now much simpler, [as outline below](https://github.com/OCR-D/core/pull/861#discussion_r979116008)

ocrd_utils/ocrd_utils/str.py
kba: Implemented

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
kba: Resolved, no need for (or rather, too much potential errors with) `is_alternative_image`, implemented largely as proposed.

ocrd_utils/ocrd_utils/str.py
kba: No more "`make_file_id` trying to be smart", the file IDs it generates will now always be either `{file_grp}_{file_id}` or `{file_grp}_{page_id}` (with mitigation for invalid `xs:ID`). It's up to the caller to ensure that the ID is actually unique, e.g. by appending a string. If it is not unique that means either bad input data (which we shouldn't try to work around but rather fail explicitly) or a bug in the processor (which should be fixed).

ocrd_utils/ocrd_utils/str.py
```diff
@@ -62,25 +62,24 @@ def make_file_id(ocrd_file, output_file_grp):
...
     If ``ocrd_file``'s ID contains the input file's fileGrp name, then replace it by ``output_file_grp``.
-    Else if ``ocrd_file``'s ID contains the input file's pageId, then merely append ``output_file_grp``.
-    Otherwise use ``output_file_grp`` together with the position of ``ocrd_file`` within the input fileGrp
-    (as a fallback counter), and increment counter until there is no more ID conflict.
+    Else if ``ocrd_file`` has a ``pageId`` but it is not contained in the ``ocrd_file.ID``, concatenate ``output_file_grp`` and ``ocrd_file.pageId``.
+    Otherwise concatenate ``output_file_grp`` with the ``ocrd_file.ID``.
+
+    Note: ``make_file_id`` cannot guarantee that the ID is unique within an actual ``OcrdMets`` but will ensure uniqueness in a workflow with processors that consistently use ``make_file_id`` for ID generation.
+
+    Note: ``make_file_id`` generates page-specific IDs. For IDs referencing sub-page-level files or ``pc:AlternativeImage``, the output of ``make_file_id`` might need to be concatenated with a unique string, such as a growing index number.
```
bertsky, line 70:
```suggestion
    Note: ``make_file_id`` generates page-specific IDs. For IDs representing page segments
    or ``pc:AlternativeImage`` files, the output of ``make_file_id`` may need to be concatenated
    with a unique string for that sub-page element, such as `".IMG"` or the segment ID.
```

ocrd_utils/ocrd_utils/str.py
```diff
@@ -62,25 +62,24 @@ def make_file_id(ocrd_file, output_file_grp):
...
     """
+    # considerations for this behaviour:
+    # - uniqueness (in spite of different METS and processor conventions)
+    # - predictability (i.e. output name can be anticipated from the input name)
+    # - stability (i.e. output at least as much sorted and consistent as the input)
+    # ... and all this in spite of --page-id selection and --overwrite
```
bertsky, line 76:
```suggestion
    # ... and all this in spite of --page-id selection and --overwrite
    # (i.e. --overwrite should target the existing ID, and input vs output
    #  IDs should be different, except when overwriting the input fileGrp)
```

ocrd/ocrd/workspace.py
```diff
@@ -1032,6 +1032,7 @@ def save_image_file(self, image,
             local_filename=file_path,
             mimetype=mimetype,
             content=image_bytes.getvalue(),
+            is_alternative_image=True,
```
bertsky, line 1035:
```suggestion
```

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -485,13 +485,14 @@ def remove_physical_page_fptr(self, fileId):
...
 
-    def merge(self, other_mets, fileGrp_mapping=None, fileId_mapping=None, pageId_mapping=None, after_add_cb=None, **kwargs):
+    def merge(self, other_mets, force=False, fileGrp_mapping=None, fileId_mapping=None, pageId_mapping=None, after_add_cb=None, **kwargs):
         """
         Add all files from other_mets.
 
         Accepts the same kwargs as :py:func:`find_files`
 
         Keyword Args:
+            force (boolean): Whether to add_files with force (overwriting existing mets:file)
```
bertsky, line 495:
```suggestion
            force (boolean): Whether to :py:meth:`add_file`s with force (overwriting existing ``mets:file``s)
```

tests/model/test_ocrd_mets.py
```diff
@@ -116,12 +115,16 @@ def test_add_group():
...
     mets = OcrdMets.empty_mets()
     assert len(mets.file_groups) == 0, '0 file groups'
     assert len(list(mets.find_all_files(fileGrp='OUTPUT'))) == 0, '0 files in "OUTPUT"'
     f = mets.add_file('OUTPUT', ID="foo123", mimetype="bla/quux", pageId="foobar")
-    f2 = mets.add_file('OUTPUT', ID="foo1232", mimetype="bla/quux", pageId="foobar")
+    # TODO unless pageId/mimetype/fileGrp match raises exception this won't work
+    # with pytest.raises(Exception) as exc:
+    #     f2 = mets.add_file('OUTPUT', ID="foo1232", mimetype="bla/quux", pageId="foobar")
+    # assert str(exc.value) == "Exception: File with pageId='foobar' already exists in fileGrp 'OUTPUTx'"
+    f2 = mets.add_file('OUTPUT', ID="foo1232", mimetype="bla/quux", pageId="foobar", is_alternative_image=True)
```
bertsky, line 127:
```suggestion
    f2 = mets.add_file('OUTPUT', ID="foo1232", mimetype="bla/quux", pageId="foobar")
```

tests/processor/test_processor.py
```diff
@@ -127,7 +165,7 @@ class ZipTestProcessor(Processor): pass
             ws = self.resolver.workspace_from_nothing(directory=tempdir)
             ws.add_file('GRP1', mimetype=MIMETYPE_PAGE, file_id='foobar1', page_id='phys_0001')
             ws.add_file('GRP1', mimetype='image/png', file_id='foobar1img1', page_id='phys_0001')
-            ws.add_file('GRP1', mimetype='image/png', file_id='foobar1img2', page_id='phys_0001')
+            ws.add_file('GRP1', mimetype='image/png', file_id='foobar1img2', page_id='phys_0001', is_alternative_image=True)
```
bertsky, line 168:
```suggestion
            ws.add_file('GRP1', mimetype='image/png', file_id='foobar1img2', page_id='phys_0001')
```

tests/processor/test_processor.py
```diff
@@ -141,7 +179,7 @@ class ZipTestProcessor(Processor): pass
                     print("PAGE-filtered")
                     tuples = [(one.ID, two) for one, two in proc.zip_input_files(mimetype=MIMETYPE_PAGE)]
                     assert ('foobar3', None) in tuples
-            ws.add_file('GRP2', mimetype='image/tiff', file_id='foobar4dup', page_id='phys_0002')
+            ws.add_file('GRP2', mimetype='image/tiff', file_id='foobar4dup', page_id='phys_0002', is_alternative_image=True)
```
bertsky, line 182:
```suggestion
            ws.add_file('GRP2', mimetype='image/tiff', file_id='foobar4dup', page_id='phys_0002')
```

tests/processor/test_processor.py
```diff
@@ -152,7 +190,7 @@ class ZipTestProcessor(Processor): pass
                     assert ('foobar3', None) in tuples
                     with self.assertRaisesRegex(Exception, "No PAGE-XML for page .* in fileGrp .* but multiple matches."):
                         tuples = proc.zip_input_files(on_error='abort')
-            ws.add_file('GRP2', mimetype=MIMETYPE_PAGE, file_id='foobar2dup', page_id='phys_0001')
+            ws.add_file('GRP2', mimetype=MIMETYPE_PAGE, file_id='foobar2dup', page_id='phys_0001', is_alternative_image=True)
```
bertsky, line 193:
```suggestion
            ws.add_file('GRP2', mimetype=MIMETYPE_PAGE, file_id='foobar2dup', page_id='phys_0001')
```

bertsky:
You said re-review, which I did, but now you already merged without my suggestions…

ocrd/ocrd/workspace.py
kba: f6eb0419e

tests/model/test_ocrd_mets.py
kba: f6eb0419e

tests/processor/test_processor.py
kba: f6eb0419e

tests/processor/test_processor.py
kba: f6eb0419e

tests/processor/test_processor.py
kba: f6eb0419e

ocrd_models/ocrd_models/ocrd_mets.py
kba: e8c189c21

ocrd_utils/ocrd_utils/str.py
```diff
@@ -62,25 +62,24 @@ def make_file_id(ocrd_file, output_file_grp):
     Derive a new file ID for an output file from an existing input file ``ocrd_file``
     and the name of the output file's ``fileGrp/@USE``, ``output_file_grp``.
     If ``ocrd_file``'s ID contains the input file's fileGrp name, then replace it by ``output_file_grp``.
-    Else if ``ocrd_file``'s ID contains the input file's pageId, then merely append ``output_file_grp``.
-    Otherwise use ``output_file_grp`` together with the position of ``ocrd_file`` within the input fileGrp
-    (as a fallback counter), and increment counter until there is no more ID conflict.
+    Else if ``ocrd_file`` has a ``pageId`` but it is not contained in the ``ocrd_file.ID``, concatenate ``output_file_grp`` and ``ocrd_file.pageId``.
```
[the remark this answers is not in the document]
kba: c9de689f9

ocrd_utils/ocrd_utils/str.py
```diff
@@ -62,25 +62,24 @@ def make_file_id(ocrd_file, output_file_grp):
...
+    Otherwise concatenate ``output_file_grp`` with the ``ocrd_file.ID``.
+
+    Note: ``make_file_id`` cannot guarantee that the ID is unique within an actual ``OcrdMets`` but will ensure uniqueness in a workflow with processors that consistently use ``make_file_id`` for ID generation.
```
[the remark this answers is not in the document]
kba: c9de689f9

ocrd_utils/ocrd_utils/str.py
kba: c9de689f9

kba:
> > You said re-review, which I did, but now you already merged without my suggestions…
> 
> I'll amend.

Done in https://github.com/OCR-D/core/tree/861-re-review. Many thanks for the documentation fixes.

PR #864, opened 2022-05-18
bulk-add: 1-based logging counter

bertsky:


kba (approved):
Agreed, `[    0/10]` always looked weird. Maybe we should also drop the padding or at least use it consistency for `i` and `len(file_paths)`.

Issue #854, opened 2022-05-11
ocrd resmgr list-available -e ocrd-keraslm-rate raises KeyError


hnesk:
`ocrd resmgr list-available -e ocrd-keraslm-rate`

raises a KeyError for all executables without resources. 

> Traceback (most recent call last):
>   File "{ocrd-env}/bin/ocrd", line 8, in <module>
>     sys.exit(cli())
>   File "{ocrd-env}/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
>     return self.main(*args, **kwargs)
>   File "{ocrd-env}/lib/python3.8/site-packages/click/core.py", line 1055, in main
>     rv = self.invoke(ctx)
>   File "{ocrd-env}/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "{ocrd-env}/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "{ocrd-env}/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>   File "{ocrd-env}/lib/python3.8/site-packages/click/core.py", line 760, in invoke
>     return __callback(*args, **kwargs)
>   File "{ocrd-env}/lib/python3.8/site-packages/ocrd/cli/resmgr.py", line 48, in list_available
>     for executable, reslist in resmgr.list_available(executable):
>   File "{ocrd-env}/lib/python3.8/site-packages/ocrd/resource_manager.py", line 87, in list_available
>     return [(executable, self.database[executable])]
> KeyError: 'ocrd-keraslm-rate'

I would have expected a single line response: 
> ocrd-keraslm-rate

or a short error message

PR #865, opened 2022-05-19
Resmgr list available nonexisting

joschrew:
Fix for https://github.com/OCR-D/core/issues/854

### DIFF
--- a/ocrd/ocrd/cli/resmgr.py
+++ b/ocrd/ocrd/cli/resmgr.py
@@ -45,11 +45,7 @@ def list_available(executable=None):
     List available resources
     """
     resmgr = OcrdResourceManager()
-    resources = resmgr.list_available(executable)
-    if executable and not resources:
-        print("Error: No resources available for '%s'" % executable)
-        sys.exit(1)
-    for executable, reslist in resources:
+    for executable, reslist in resmgr.list_available(executable):
         print_resources(executable, reslist, resmgr)
 
 @resmgr_cli.command('list-installed')
--- a/ocrd/ocrd/cli/resmgr.py
+++ b/ocrd/ocrd/cli/resmgr.py
@@ -45,7 +45,11 @@ def list_available(executable=None):
     List available resources
     """
     resmgr = OcrdResourceManager()
-    for executable, reslist in resmgr.list_available(executable):
+    resources = resmgr.list_available(executable)
+    if executable and not resources:
+        print("Error: No resources available for '%s'" % executable)
+        sys.exit(1)
+    for executable, reslist in resources:
         print_resources(executable, reslist, resmgr)
 
 @resmgr_cli.command('list-installed')
--- a/ocrd/ocrd/resource_manager.py
+++ b/ocrd/ocrd/resource_manager.py
@@ -87,7 +87,6 @@ class OcrdResourceManager():
             if executable not in self.database:
                 return []
             return [(executable, self.database[executable])]
-
         return self.database.items()
 
     def list_installed(self, executable=None):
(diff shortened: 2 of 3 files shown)


ocrd/ocrd/cli/resmgr.py
```diff
@@ -45,7 +45,11 @@ def list_available(executable=None):
     List available resources
     """
     resmgr = OcrdResourceManager()
-    for executable, reslist in resmgr.list_available(executable):
+    resources = resmgr.list_available(executable)
+    if executable and not resources:
+        print("Error: No resources available for '%s'" % executable)
+        sys.exit(1)
+    for executable, reslist in resources:
```
bertsky, line 52: That hunk should be reverted for several reasons:
- we cannot just `print`, that would mix expected output (usually to be parsed by another tool) with messages (which should use logging and thus end up in a file or on stderr)
- we don't want a failure, just an empty result in that case
- it's not necessary, your fix in `OcrdResourceManager.list_available` is already sufficient

```suggestion
    for executable, reslist in resmgr.list_available(executable):
```

bertsky (changes requested):
Otherwise fine – thanks!

ocrd/ocrd/cli/resmgr.py
kba: Yes, I understand the @joschrew's intention to give a better error messages. But I agree with @bertsky this is not the right place, or rather it is not necessary.

If we want to make it explicit that there are no resources for an executable, the right place to adapt ist the `print_resources` function.

ocrd/ocrd/resource_manager.py
```diff
@@ -84,6 +84,8 @@ def list_available(self, executable=None):
         List models available for download by processor
         """
         if executable:
+            if executable not in self.database:
+                return []
             return [(executable, self.database[executable])]
```
kba, line 89:
```suggestion
            return [(executable, self.database.get(executable, []])]
```

We need to return the same format, a list of tuples `[(executable, list_of_resources)]`, returning just an empty list is not enough.

tests/test_resource_manager.py
```diff
@@ -104,5 +104,16 @@ def test_default_resource_dir(tmp_path):
...
 
+def test_list_available(tmp_path):
+    mgr = OcrdResourceManager(xdg_data_home=tmp_path)
+    res = mgr.list_available()
+    assert len(res) > 0
+
+def test_list_available_with_unknown_executable(tmp_path):
+    mgr = OcrdResourceManager(xdg_data_home=tmp_path)
+    res = mgr.list_available(executable="ocrd-non-existing-processor")
+    assert len(res) == 0
```
kba, line 115:
```suggestion
    assert len(res[0][1])  == 0
```

kba (changes requested):
Thanks for the PR, it solves the issue, just a few adaptions and then we can merge!

ocrd/ocrd/resource_manager.py
bertsky: @kba I don't think that's true. If the returned list is empty, then nothing will get printed (which is probably what you want when using `ocrd resmgr list-available -e ...` for other tools). In your variant `ocrd.cli.print_resources` will print the name of the processor, followed by an empty line.

kba: Yes, that's why I meant that we should adapt `print_resources` to print something like

```
ocrd-foo-bar: NONE
```

or similar.

joschrew:
ok, thanks for all your support. I think (hope) I changed it to that what kba initially showed me and it is working now.

kba (approved):

PR #868, opened 2022-05-20
resource_list: update eynollah URL

bertsky:
Note: I did not include the 2022 version because the code [cannot use it yet](https://github.com/qurator-spk/eynollah/issues/78).


bertsky:
It should not take that long to just update the resource URL. People are _not_ able to download eynollah models currently.

kba (approved):
Yes, I wanted to finalize the decentralization before but it's taking too long, agreed.

PR #870, opened 2022-05-20
ocrd log: read from stdin if passed "-", fix #852

### LINKED ISSUE #852: bashlib: allow piping into ocrd log
To make it callable like util-linux' `logger`, wrapping output from programs, e.g.

    some-command |& ocrd log -n my-processor warning -

kba:
So you can do

```
echo "foo bar
quux" | ocrd log warning -
```

and get

```
13:08:48.815 WARNING root - foo bar
13:08:48.815 WARNING root - quux
```


bertsky (approved):
Great stuff!

Issue #850, opened 2022-05-09
Tests failed for OS X


tdoan2010:
Running `make test` leads to 8 tests failed. My current `pillow` version is 9.1.0

```
Name: Pillow
Version: 9.1.0
Summary: Python Imaging Library (Fork)
Home-page: https://python-pillow.org
Author: Alex Clark (PIL Fork Author)
Author-email: aclark@python-pillow.org
License: HPND
Location: /Users/tdoan/projects-gwdg/ocr-d/ocrd-core/.venv/lib/python3.10/site-packages
Requires: 
Required-by: ocrd-utils
```

The short test summary info looks like this:

```
FAILED tests/test_utils.py::TestUtils::test_pushd_popd_newcwd - AssertionError: '/private/tmp' != '/tmp'
FAILED tests/test_utils.py::TestUtils::test_pushd_popd_tempdir - AssertionError: '/private/tmp/tmp6lvhxl2m' != '/tmp/tmp6lvhxl2m'
FAILED tests/cli/test_workspace.py::TestCli::test_add_existing_checked - AssertionError: 1 != 0
FAILED tests/cli/test_workspace.py::TestCli::test_add_remove_force - AssertionError: True is not false
FAILED tests/cli/test_workspace.py::TestCli::test_mets_basename_and_not_mets - AssertionError: '/private/tmp/tmpdkb8w8ja/foo\n' != '/tmp/tmpdkb8w8ja/foo\n'
FAILED tests/cli/test_workspace.py::TestCli::test_remove_file_group - AssertionError: '/private/tmp/tmpx7eqrp93/ws' != '/tmp/tmpx7eqrp93/ws'
FAILED tests/validator/test_workspace_validator.py::TestWorkspaceValidator::test_dimensions - AssertionError: "PAGE 'PAGE_0017_PAGE': @imageHeight != image's actual height (1234 != 2083)" not found in []
FAILED tests/validator/test_workspace_validator.py::TestWorkspaceValidator::test_pcgtsid - AssertionError: 'pc:PcGts/@pcGtsId differs from mets:file/@ID: "foo" !== "PAGE_0017_PAGE"' not found in []
```

kba:
Hmm I remember it looked like a Pillow issue to me when you shared your screen but the error messages don't really reflect that. I've installed python 3.10 and pillow 9.1.0 and `make test` works fine for me, albeit with a few DeprecationWarnings on distutils.

Does this still occur for you? In that case, let's schedule a call and debug.

[...]

PR #873, opened 2022-05-23
try testing OSX, #850

kba:


tdoan2010:
The `test-macos` on CircleCI is still failed. Should it be fixed first?

kba (commented):
LGTM, thanks for fixing these.

Issue #723, opened 2021-09-28
performance of input_files on large workspaces


bertsky:
By implementing #635 to properly handle all cases of PAGE-XML file matching per pageId, we have lost sight of the severe performance penalty that this comes with. In effect, we are now nearly as slow as before #482 on workspaces with lots of pages and fileGrps.

Here's a typical scenario:
1. ocrd-cis-ocropy-dewarp creates an image file for each text line, and references it under the pageId and fileGrp which it belongs to – under an image mimetype (this creates 29.000 files for me in a workspace with 500 pages)
2. ocrd-tesserocr-recognize runs afterwards and queries its `self.input_files`:
https://github.com/OCR-D/core/blob/9069a6581f37ec1c189e8cfaa62692fb66004964/ocrd/ocrd/processor/base.py#L294-L299
3. this searches through all `mets:file` entries, matching them for `fileGrp` (which is reasonably fast, it only gets a little inefficient when additionally filtering by `pageId`):
https://github.com/OCR-D/core/blob/9069a6581f37ec1c189e8cfaa62692fb66004964/ocrd_models/ocrd_models/ocrd_mets.py#L176-L208
4. Then in line 298 (and again further below) it queries `OcrdFile.pageId`:
https://github.com/OCR-D/core/blob/9069a6581f37ec1c189e8cfaa62692fb66004964/ocrd_models/ocrd_models/ocrd_file.py#L116-L122
5. This in turn needs to query the whole structMap via XPath (which on a workspace with 500 files and 25 fileGrps and 200.000 takes about 0.2sec per file, i.e. needs more than 1h just for the computation of `input_files`):
https://github.com/OCR-D/core/blob/9069a6581f37ec1c189e8cfaa62692fb66004964/ocrd_models/ocrd_models/ocrd_mets.py#L434-L441

A little cosmetics like turning `OcrdFile.pageId` into a `functools.cached_property` won't help here, the problem is bigger. METS with its mutually related fileGrp and pageId mappings is inherently expensive to parse. I know we have in the past decided against in-memory representations like dicts because that looked like memory leaks or seemed to expensive on very large workspaces. But have we really weighed the cost of that memory-cputime tradeoff carefully (and considering the necessity for pageId/mimetype filtering) yet? Is there any existing code attempting to cache fileGrp and pageId mappings to avoid reparsing the METS again and again, which I could tamper with?

kba:
[...]

PR #875, opened 2022-05-31
Cache functionality

MehmedGIT:


ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -50,20 +51,91 @@ def empty_mets(now=None):
...
+        if el_fileGrp_list is None or len(el_fileGrp_list) == 0:
+            return
+
+        for el_fileGrp in el_fileGrp_list:
+            fileGrp_use = el_fileGrp.get('USE')
+
+            # NOTE: For some reason the el_fileGrp_list contains None values
+            # when testing with the SBB0000F29300010000/data/mets.xml
+            if fileGrp_use is None:
+                continue
```
kba, line 106: That happens because that METS contains comments which lxml considers to be elements.

```suggestion
            if el_fileGrp.tag is ET.Comment:
                continue
            fileGrp_use = el_fileGrp.get('USE')
```

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -50,20 +51,91 @@ def empty_mets(now=None):
...
+
+            self._fileGrp_cache[fileGrp_use] = el_fileGrp
+            print("_fill_caches> file group added to the cache: %s" % fileGrp_use)
```
kba, line 109:
```suggestion
            log = getLogger('ocrd_models.ocrd_mets.cache')
            log.info("_fill_caches> file group added to the cache: %s", fileGrp_use)
```

kba: same for the other print statements. We mustn't interfere with `STDOUT`.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -50,20 +51,91 @@ def empty_mets(now=None):
...
+        for key in list(self._fileGrp_cache):
+            del self._fileGrp_cache[key]
+            fileGrp_counter += 1
+
+        # print("_clear_caches> total cleared fileGrp cache elements: %d" % fileGrp_counter)
+
+        for key in list(self._file_cache):
+            for inner_key in list(self._file_cache[key]):
+                del self._file_cache[key][inner_key]
+            del self._file_cache[key]
```
kba, line 137: Why not simply `self._fileGrp_cache = None` or `del self._file_cache`? Python will garbage-collect dangling references. But the references will remain anyway, because they are tied to `self._tree`.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -249,6 +418,11 @@ def rename_file_group(self, old, new):
             raise FileNotFoundError("No such fileGrp '%s'" % old)
         el_fileGrp.set('USE', new)
 
+        # Rename the fileGrp in both caches
+        if self._cache_flag:
+            self._fileGrp_cache[new] = self._fileGrp_cache.pop(old)
+            self._file_cache[new] = deepcopy(self._file_cache.pop(old))
```
kba, line 424: Why `deepcopy`?

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -310,10 +495,25 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
+        if el_fileGrp is None:
+            el_fileGrp = self._tree.getroot().find(".//mets:fileGrp[@USE='%s']" % (fileGrp), NS)
+
+        # the fileGrp is not in the XML tree as well
         if el_fileGrp is None:
             el_fileGrp = self.add_file_group(fileGrp)
-        mets_file = next(self.find_files(ID=ID), None)
+
+        # Since we are sure that fileGrp parameter is set,
+        # we could send that parameter to find_files for direct search
```
kba, line 515: Agreed

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -127,6 +199,95 @@ def find_all_files(self, *args, **kwargs):
 
         Equivalent to ``list(self.find_files(...))``
         """
+
+        # NOTE: This code gets complex with the REGEX.
```
kba, line 203: I gather you extended `find_all_files` instead of the `find_files` it delegates to because of issues with `return`/`yield`? Because implemented like this, you'd only benefit from the cache when doing the former, while we generally use the latter.

kba (commented):
Just a quick read-thru for now, I'll dig into the code further.

ocrd_models/ocrd_models/ocrd_mets.py
MehmedGIT: Okay.

ocrd_models/ocrd_models/ocrd_mets.py
MehmedGIT: Right. I will use the logger for that and replace the print statements.

ocrd_models/ocrd_models/ocrd_mets.py
MehmedGIT: For a similar reason as above. To take care of the nested dictionaries properly. If `deep` is not required and simply passing the `self._file_cache[new] = self._file_cache.pop(old)` is smart enough to deal with nested data structures, it could be removed.

ocrd_models/ocrd_models/ocrd_mets.py
kba: They should be, GC in Python is fairly robust.

ocrd_models/ocrd_models/ocrd_mets.py
kba: Yes, then `deepcopy` is not needed here, python will reuse the existing reference just fine and do the cleanup of `pop`.

ocrd_models/ocrd_models/ocrd_mets.py
kba: OK, so the cache as it is can only speed up lookup by `ID` and `fileGrp`. So:

* find_files(ID='...') -> cache
* find_files(fileGrp='...') -> cache
* find_files(ID='...', fileGrp='...') -> cache

In this case, we can yield those results.

If there are more kwargs to `find_files`, we can still use the cache for a list of candidates and then use the existing logic to drill down those candidates accordingly

> Then the `yield` returns all matches inside the `self._cache_flag` condition block and enters the `for cand in self._tree ...` loop to return 5 more files. 10 in total.

But why can we not deduplicate by `ID`? Or rather, why go through the XML if we have the cached information?

On a general note, have you considered [memoizing](https://docs.python.org/3/library/functools.html#functools.cache) the `find_files` method, so that calling `find_files` with the same arguments instantly returns the cached results?

kba: On another general note: In what circumstances would one want to disable the caching? I.e. do we need `_cache_flag`?

kba:
Also, why have separate `_fileGrp_cache` and `_file_cache` - doesn't the latter also contain the fileGrp info in a fast-to-access way? What would be more beneficial would be caching the `pageId` of a file because that requires a bit of a convoluted XPath to look up on-the-fly.

ocrd_models/ocrd_models/ocrd_mets.py
kba: > If there are more kwargs to `find_files`, we can still use the cache for a list of candidates and then use the existing logic to drill down those candidates accordingly

See d60c8d06e for what I mean with that.

ocrd_models/ocrd_models/ocrd_mets.py
MehmedGIT: Okay.

kba:
> The `_fileGrp_cache` holds a reference to the fileGrp element itself aka the memory location. The `_file_cache` holds just a string of the fileGrp.

I get that, but the main/only use for querying about `fileGrp` is: Does this `fileGrp` exist (a) and what files are in it (b)? These can be answered with the `_file_cache`, by checking `needle in _file_cache` (a) and `for el_file in self._file_cache[needle]:` (b). Since the `fileGrp` element itself contains no other information in the METS that I'm aware of, it seems redundant to keep it around.

MehmedGIT:
Screenshot from 2022-05-13 11-08-32

Constructing the METS file with 50 files inside got improved from 8,7s down to 508 ms! Here we go, less electricity consumption for everyone I guess :)

bertsky:
Related issue: https://github.com/OCR-D/core/issues/723

Related discussion: https://github.com/OCR-D/zenhub/issues/39

Still missing IIUC: 
> I would suggest diversifying test scenarios BTW:
>     * thousands of pages
>     * dozens of fileGrps
>     * fileGrps with hundreds of files per page (think `ocrd-cis-ocropy-dewarp` for line level image files or `ocrd-cis-ocropy-deskew` / `ocrd-tesserocr-deskew` for region level files)

And a switch to enable/disable caching – ideally somewhere central, but we don't have a general-purpose configuration mechanism yet (except for logging), cf. [options](https://github.com/OCR-D/core/issues/376#issuecomment-1163028758).

bertsky:
No, sorry, but I have no advice on specific numbers for such combinations. Your choices seem to be some "small to medium average" scenario. What I called for was adding more extreme scenarios – not necessarily under assets and not necessarily all possible combinations.

bertsky:
Thanks @MehmedGIT, very good!

Perhaps you could include your local results as a Markdown table into `ocrd_models/README.md` or a new file under `tests/model/README.md`?

> Still missing IIUC:
> 
> > * fileGrps with hundreds of files per page (think `ocrd-cis-ocropy-dewarp` for line level image files or `ocrd-cis-ocropy-deskew` / `ocrd-tesserocr-deskew` for region level files)

(in particular, hundreds of files per pages _combined_ with hundreds or even thousands of pages – this is realistic with dewarping in the workflow)

> 
> And a switch to enable/disable caching – ideally somewhere central, but we don't have a general-purpose configuration mechanism yet (except for logging), cf. [options](https://github.com/OCR-D/core/issues/376#issuecomment-1163028758).

@kba your thoughts on this aspect?

MehmedGIT:
> Perhaps you could include your local results as a Markdown table into `ocrd_models/README.md` or a new file under `tests/model/README.md`?

@bertsky, yes, I will do that once I have the additional extreme results (below).

> (in particular, hundreds of files per pages _combined_ with hundreds or even thousands of pages – this is realistic with dewarping in the workflow)

I have started now benchmarking with 750 files per page and 5000 pages - both for cached and non-cached.
Please be patient about the results since the non-cached version would probably take a lot of processing time. 
Just for comparison - the results above for 5000 pages (31 files per page) took 7.5 hours.

MehmedGIT:
@bertsky,

I have canceled the test execution because even the building of the mets file for the non-cached version has not finished in almost 5 days. Since I could not use my work laptop for anything else efficiently due to the 100% CPU usage of the test, I had to cancel the test. I have pushed the test case for 750 files per page and 5000 pages if someone wants to play and further test with different numbers for files and pages.

Screenshot from 2022-10-04 12-16-52

MehmedGIT:
> @MehmedGIT understood. (Perhaps a server machine or SSH build on CircleCI would help with your resource limitation.)
> 
> It would help at least knowing how much that test took with caching enabled, though...

@bertsky, I will execute the tests again on a cloud VM provided by my company. I was also thinking about whether 750 files per page is more than what we need. Do you think 300-400 files per page is still good enough? What is the theoretical maximum for files per page, if there are some? 

Another thing I could do is to increase the files per page but reduce the number of pages. Say 1500 files per page but "only" 2000 pages? 2000 pages are still roughly 5 times above the average page numbers from the statistics I have.

Seiten

bertsky:
> Do you think 300-400 files per page is still good enough? 

yes

> What is the theoretical maximum for files per page, if there are some?

there's nothing I can think of. Practically, we might see newspaper pages with 1000 lines or so. Also, for the typegroups classification case, there's even a use-case for word-level image files. But these are rather rare cases. And we already get the picture by running a more conservative high-file count case. Especially if your statistics allow breaking down page count scaling behaviour.

> Another thing I could do is to increase the files per page but reduce the number of pages. Say 1500 files per page but "only" 2000 pages? 2000 pages are still roughly 5 times above the average page numbers from the statistics I have.

Yes. By the same logic, I would recommend reducing the page number to under 1000.

MehmedGIT:
Since the caching_functionality branch was behind by 150 commits from the master, I merged the master. The test example with 1500 files per page and 500 pages is running inside a cloud VM. Once I have the results, I will publish them.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -215,6 +333,8 @@ def find_files(self, ID=None, fileGrp=None, pageId=None, mimetype=None, url=None
                 else:
                     if not url.fullmatch(cand_url): continue
 
+            # Note: why we instantiate a class only to find out that the local_only is set afterwards
+            # Checking local_only and url before instantiation should be better?
```
kba, line 337: Instantiation is fast, it just does assignment to instance vars and determines whether it's a local file in `OcrdFile.__init``. I n almost all cases, `local_only` will not be set and we will yield that file.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -311,20 +445,41 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
+
+        """
+        # Note: we do not benefit enough from having 
+        # a separate cache for fileGrp elements
+
+        if self._cache_flag: 
+            if fileGrp in self._fileGrp_cache:
+                el_fileGrp = self._fileGrp_cache[fileGrp]
+        """
+
```
kba, line 457: I don't think caching would help in this place but we could benefit minimally in `add_file_group`, if we determine that the `fileGrp` to add is already a key in `self._file_cache`, we could skip the element exist/create code there.

```suggestion
```

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -311,20 +445,41 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
+            # we could send that parameter to find_files for direct search
+            mets_file = next(self.find_files(ID=ID, fileGrp=fileGrp), None)
             if mets_file:
                 if mets_file.fileGrp == fileGrp and \
                    mets_file.pageId == pageId and \
                    mets_file.mimetype == mimetype:
                     if not force:
                         raise FileExistsError(f"A file with ID=={ID} already exists {mets_file} and neither force nor ignore are set")
-                    self.remove_file(ID=ID)
+                    self.remove_file(ID=ID, fileGrp=fileGrp)
```
kba, line 469: `ID` must be unique, I see no benefit in adding `fileGrp` to the call.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -362,20 +515,48 @@ def remove_one_file(self, ID):
...
+                if self._cache_flag:
+                    del self._page_cache[page_div.get('ID')]
+                    del self._fptr_cache[page_div.get('ID')]
+
+        # Delete the file reference from the cache
+        if self._cache_flag:
+            parent_use = ocrd_file._el.getparent().get('USE')
+            # Note: if the file is in the XML tree,
+            # it must also be in the file cache.
+            # Anyway, we perform the checks, then remove
```
kba, line 556: We should indeed be sure of that, lest we have some inconsistent state.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -422,9 +614,25 @@ def set_physical_page_for_file(self, pageId, ocrd_file, order=None, orderlabel=N
         """
         #  print(pageId, ocrd_file)
         # delete any page mapping for this file.ID
-        for el_fptr in self._tree.getroot().findall(
+        
+        # NOTE: The pageId coming from 'test_merge(sbb_sample_01)' is an Element not a string
+        if not isinstance(pageId, str):
+            pageId = pageId.get('ID')
```
kba, line 620:
```suggestion
```

This should not be the case. Are you positively certain that is the case? I cannot reproduce and if we can, that's a bug in merge we should fix.

kba:
I have now cleaned up the test setup and merged #955 into this, I think this is ready for merge.

* The benchmarks are not run as part of `make test`, but with `make benchmark` or `make benchmark-extreme`. For the CI we run `make test benchmarks`.
* The `test_ocrd_mets.py` will now test both uncached and cached OcrdMets

What is still missing for release?

- CLI flags
- anything else? @MehmedGIT

ocrd_models/ocrd_models/ocrd_mets.py
MehmedGIT: `remove_file()` calls `find_files()` inside. So, by passing the fileGrp as well we reduce the candidates size inside `find_files()`. The candidates list hold only the IDs inside a specific fileGrp and not all IDs. Without the suggested change we iterate over all IDs, with the suggested change we iterate only over some sub-list of the IDs.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -311,20 +445,41 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
-                    self.remove_file(ID=ID)
+                    self.remove_file(ID=ID, fileGrp=fileGrp)
                 else:
                     raise FileExistsError(f"A file with ID=={ID} already exists {mets_file} but unrelated - cannot mitigate")
-        kwargs = {k: v for k, v in locals().items() if k in ['url', 'ID', 'mimetype', 'pageId', 'local_filename'] and v}
-        mets_file = OcrdFile(ET.SubElement(el_fileGrp, TAG_METS_FILE), mets=self, **kwargs)
+
+        # To get rid of Python's FutureWarning - checking if v is not None
+        kwargs = {k: v for k, v in locals().items() if k in ['url', 'ID', 'mimetype', 'pageId', 'local_filename'] and v is not None}
+        # This separation is needed to reuse the same el_mets_file element in the caching if block
```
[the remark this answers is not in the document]
MehmedGIT: You can omit that change if you think it's not needed.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -362,20 +515,48 @@ def remove_one_file(self, ID):
             ocrd_file = ID
             ID = ocrd_file.ID
         else:
+            # NOTE: We should pass the fileGrp, if known, as a parameter here as well
+            # Leaving that out for now
```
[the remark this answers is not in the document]
MehmedGIT: Why do you think it will make the search slightly slower? Check my answer above.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -362,20 +515,48 @@ def remove_one_file(self, ID):
...
+
+        # Delete the file reference from the cache
+        if self._cache_flag:
+            parent_use = ocrd_file._el.getparent().get('USE')
+            # Note: if the file is in the XML tree,
+            # it must also be in the file cache.
+            # Anyway, we perform the checks, then remove
+            if parent_use in self._file_cache:
+                if ocrd_file.ID in self._file_cache[parent_use]:
+                    del self._file_cache[parent_use][ocrd_file.ID]
```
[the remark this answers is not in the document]
MehmedGIT: I agree with that. Better we get access related exceptions instead of silently ignoring potential inconsistencies.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -400,19 +584,27 @@ def get_physical_pages(self, for_fileIds=None):
...
         ret = [None] * len(for_fileIds)
-        for page in self._tree.getroot().xpath(
-            'mets:structMap[@TYPE="PHYSICAL"]/mets:div[@TYPE="physSequence"]/mets:div[@TYPE="page"]',
-                namespaces=NS):
-            for fptr in page.findall('mets:fptr', NS):
-                if fptr.get('FILEID') in for_fileIds:
-                    ret[for_fileIds.index(fptr.get('FILEID'))] = page.get('ID')
+        
+        # Note: This entire function potentially could be further simplified
+        # TODO: Simplify
```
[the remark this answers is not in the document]
MehmedGIT: I am no longer sure what I had in mind back then for simplification anymore. Sorry. Seems fine for me now.

ocrd_models/ocrd_models/ocrd_mets.py
MehmedGIT: That was the case back then. If you cannot reproduce it anymore, then probably it's solved now. I think this error was related to the return type of `keys()` which you changed to `list(keys())` in some part of the code.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -436,7 +644,14 @@ def set_physical_page_for_file(self, pageId, ocrd_file, order=None, orderlabel=N
         if el_seqdiv is None:
             el_seqdiv = ET.SubElement(el_structmap, TAG_METS_DIV)
             el_seqdiv.set('TYPE', 'physSequence')
-        el_pagediv = el_seqdiv.find('mets:div[@ID="%s"]' % pageId, NS)
+        
+        el_pagediv = None
+        if self._cache_flag:
+            if pageId in self._page_cache.keys():
+                el_pagediv = self._page_cache[pageId]
```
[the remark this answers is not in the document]
MehmedGIT: Agree.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -445,52 +660,90 @@ def set_physical_page_for_file(self, pageId, ocrd_file, order=None, orderlabel=N
...
         """
-        mets_fptrs = self._tree.getroot().xpath(
-            'mets:structMap[@TYPE="PHYSICAL"]/mets:div[@TYPE="physSequence"]/mets:div[@TYPE="page"]/mets:fptr[@FILEID="%s"]' % fileId,
-            namespaces=NS)
+
+        # Question: What is the reason to keep a list of mets_fptrs?
+        # Do we have a situation in which the fileId is same for different pageIds ?
+        # From the examples I have seen inside 'assets' that is not the case
+        # and the mets_fptrs list will always contain a single element.
+        # If that's the case then we do not need to iterate 2 loops, just one.
```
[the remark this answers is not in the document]
MehmedGIT: My idea back then as far as I remember was that probably we don't need the second loop inside the `remove_physical_page_fptr()` when the cache is used. However, I am no longer sure how I thought this could be achieved.

MehmedGIT (commented):
@kba, here are my answers. Sorry, since I implemented the code some months ago I don't fully remember what I had in mind for some of the cases.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -284,13 +410,21 @@ def remove_file_group(self, USE, recursive=False, force=False):
             if not recursive:
                 raise Exception("fileGrp %s is not empty and recursive wasn't set" % USE)
             for f in files:
+                # NOTE: Here we know the fileGrp, we should pass it as a parameter
```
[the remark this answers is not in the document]
kba: OK, so because

```python
if self._cache_flag:                                                                                                                                       
    if fileGrp:                                                                                                                                            
        if isinstance(fileGrp, str):                                                                                                                       
            candidates += self._file_cache.get(fileGrp, {}).values()                                                                                       
        else:                                                                                                                                              
            candidates = [x for fileGrp_needle, el_file_list in self._file_cache.items() if fileGrp.match(fileGrp_needle) for x in el_file_list.values()]  
    else:                                                                                                                                                  
        candidates = [el_file for id_to_file in self._file_cache.values() for el_file in id_to_file.values()]                                              
else:                                                                                                                                                      
    candidates = self._tree.getroot().xpath('//mets:file', namespaces=NS)                                                                                  
```

could use the fileGrp for accessing the cache, fine with me.


But thinking about the code here, we could at least use `[@ID=%s'] % ID` if the ID was provided for the non-caching version, that would reduce the number of candidates to, ideally, 1.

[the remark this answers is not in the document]
MehmedGIT: Without caching we still end up with a list of `mets:file` elements after 
`candidates = self._tree.getroot().xpath('//mets:file', namespaces=NS)`.
I don't see how the number of candidates can be 1.

In the best case, we match the first element and leave. 
In the worst case, we iterate the entire list to get the last element.
So the average is N/2.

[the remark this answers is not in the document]
MehmedGIT: Oh, okay I get what you meant. But still we end up with a linear time. Unless you provide the exact tree path to the element tree, the element tree still ends up iterating a list of `mets:file`. We then get only 1 candidate, true and match just that candidate in the loop afterward.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -311,20 +445,41 @@ def add_file(self, fileGrp, mimetype=None, url=None, ID=None, pageId=None, force
...
+        if self._cache_flag: 
+            if fileGrp in self._fileGrp_cache:
+                el_fileGrp = self._fileGrp_cache[fileGrp]
+        """
+
         el_fileGrp = self.add_file_group(fileGrp)
         if not ignore:
-            mets_file = next(self.find_files(ID=ID), None)
+            # Since we are sure that fileGrp parameter is set,
+            # we could send that parameter to find_files for direct search
```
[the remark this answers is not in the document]
kba: I meant that the comment is out-of-date, we do use `fileGrp` in the `find_files` call.

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
kba: I think it's cleaner to create a new element, we also got rid of the re-use of `mets:file` elements in `OcrdWorkspace.add_file`. So let's keep as-is.

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
kba: Yes, the search question has been resolved but the comment is also out-of-date, right?

ocrd_models/ocrd_models/ocrd_mets.py
kba: ffcd89f37

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -279,18 +394,34 @@ def remove_file_group(self, USE, recursive=False, force=False):
...
+        else:
+            files = el_fileGrp.findall('mets:file', NS)
+
         if files:
             if not recursive:
                 raise Exception("fileGrp %s is not empty and recursive wasn't set" % USE)
             for f in files:
-                self.remove_one_file(f.get('ID'))
+                # NOTE: Here we know the fileGrp, we should pass it as a parameter
+                self.remove_one_file(ID=f.get('ID'), fileGrp=f.get('USE'))
```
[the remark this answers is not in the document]
MehmedGIT: @kba, we could also just send the OcrdFile as a parameter. The `remove_one_file` method accepts both an OcrdFile and a string. Then we don't have to search for the file again with its `ID` and `fileGrp`.

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
kba: So I'll remove that comment, ok?

ocrd_models/ocrd_models/ocrd_mets.py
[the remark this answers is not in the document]
kba: True. And we don't need `fileGrp` in `remove_one_file` anymore, right?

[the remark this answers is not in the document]
kba: But that doesn't work here, because `f` is an `lxml.Element`, not an `OcrdFile` :/

kba:
It's finished now but instead of CLI flags, we've opted for an environment variable `OCRD_METS_CACHING` which must be set to `true` to globally enable the METS caching.

We decided against CLI flags, because that would require a lot of changes for all the places in the code that work with METS, especially passing the `cache_flag` value through function calls until the METS constructor is called.

We need to tackle a proper, unified configuration option soon and then align the caching behavior with it.

bertsky:
Great to hear this is finally ready. My understanding was that we merge https://github.com/OCR-D/core/pull/678 first and see how this affects RSS (or PSS) througout your test scenarios. Do you have any measurements like that already?

(And since you did decide in favour of environment variables here, should that affect the [discussion about profiling settings](https://github.com/OCR-D/core/pull/678#discussion_r1001891735) as well?)

kba:
> Great to hear this is finally ready. My understanding was that we merge #678 first and see how this affects RSS (or PSS) througout your test scenarios. Do you have any measurements like that already?

@MehmedGIT Do we?

> (And since you did decide in favour of environment variables here, should that affect the discussion about profiling settings as well?)

For now, I would use environment variables like `about:flags` in browsers. We document the behavior but require an extra step (setting the env var). For the long-term, I would now prefer a configuration file with env var overrides. We should discuss that in the tech calls and/or our workshop in December.

MehmedGIT:
@bertsky, @kba I would also refer again to this summary #946. From the table, we could see the comparisons and that memory usage does not increase in all cases when caching is used.

PR #880, opened 2022-06-14
small bugfix in OcrdMets.find_files

joschrew:
Minor bug: change maxsplits to 1, because `generate_range` expects exactly two params.

> If maxsplit is given, at most maxsplit splits are done (thus, the list will have at most maxsplit+1 elements

https://docs.python.org/3/library/stdtypes.html?highlight=split#str.split


kba (approved):
Thanks for spotting.

Issue #881, opened 2022-06-15
Investigate startup speedup by replacing pkg_resources


kba:
cf. https://github.com/hnesk/browse-ocrd/pull/48 and https://github.com/hnesk/browse-ocrd/commit/0008c510bbbf83ec70775309d446853a18591d30#commitcomment-75799474

Basically, replacing `pkg_resources` with `importlib*` could help us reduce the time it takes eg. to read schemas and other resources and potentially also startup when using e.g. https://github.com/ninjaaron/fast-entry_points to avoid pkg_resources in console_scripts entry points.

PR #882, opened 2022-06-23
WIP: No more pkg_resources

hnesk:
A first attempt to get rid of pkg_resources as mentioned in #881 
I did not measure anything yet, and subjectively not much changed, but we'll see.


ocrd/ocrd/processor/builtin/dummy_processor.py
```diff
@@ -17,7 +17,7 @@
 )
 from ocrd_modelfactory import page_from_file
 
-OCRD_TOOL = parse_json_string_with_comments(resource_string(__name__, 'dummy/ocrd-tool.json').decode('utf8'))
+OCRD_TOOL = parse_json_string_with_comments(resource_string(__name__, 'ocrd-tool.json').decode('utf8'))
```
kba, line 20: Was the change in path deliberate?

ocrd_utils/ocrd_utils/package_resources.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Using importlib for better performance
+
+    package : str
+        The package from where to start looking for resource (often __name__)
+    resource : str
+        The resource to look up
+    """
+    parent_package = package.rsplit('.',1)[0]
+    return _file_manager.enter_context(path(parent_package, resource))
```
kba, line 27: Why enter the context of an ExitStack here?

hnesk: This is adapted from a recipe in the[importlib_resources migration guide](https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename). The returned file name could be a temporary file (e.g. extracted from a zip). The importlib_resources-functions return a context manager to manage the lifecycle of that file. To mimic the semantics of resource_filename, I had to get rid of the context manager, so I used the ExitStack recipe to have the file available for the whole lifetime of the process and cleaned up at exit. 
The migration explains it better than me ;)

ocrd_utils/ocrd_utils/constants.py
```diff
@@ -1,10 +1,13 @@
 """
 Constants for ocrd_utils.
 """
-from pkg_resources import get_distribution
 from re import compile as regex_compile
 from os import environ
 from os.path import join, expanduser
+try:
```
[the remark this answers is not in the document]
hnesk: Yes, good idea. It didn't work at first at installation time, but that was because I listed the `importlib_*` requirements in `ocrd/requirements.txt` instead of `ocrd_utils/requirements.txt`

ocrd_utils/ocrd_utils/package_resources.py
kba: I see, so this ensures that paths are retained even when the package is accessed in a zipped format. Not 100% sure I understand when this would happen. I'll add a comment referencing the migration guide.

hnesk:
I just removed the WIP label, as the implementation is finished. The performance improvement is neglible, as far as I could interpret the profiler, the startup time is spent mostly on importing modules, and I have no clue how to improve on that.

PR #884, opened 2022-07-05
REST API wrapper for the processor

tdoan2010:
# Description

This pull request provides the REST API wrapper for the processor. Instead of "one-shot mode", the processor now can be started as a server. There are 2 ways to start the processor:

### 1. Use processor name

```shell
$ ocrd-dummy --server-ip=0.0.0.0 --server-port=80 --mongo-url=mongodb://localhost:27017
```

This is the recommended way to start the Pythonic processor as a server. There are 3 options:

1. `--server-ip`: the IP address to which the server should bind.
2. `--server-port`: the port number on which the server should listen.
3. `--mongo-url`: the URL to the Mongo database.

### 2. Use `ocrd` tool:

```shell
$ ocrd server --tool=ocrd-dummy --ip=0.0.0.0 --port=80 --mongo-url=mongodb://localhost:27017 ocrd/ocrd/cli/ocrd-tool.json
```
This approach is for non-Python processor. There are 4 options and 1 argument:

1. `--tool`: the name of the executable in the `ocrd-tool.json` file.
1. `--ip`: the IP address to which the server should bind.
2. `--port`: the port number on which the server should listen.
3. `--mongo-url`: the URL to the Mongo database.
4. Path to the `ocrd-tool.json`. This is the argument.


ocrd/ocrd/decorators/__init__.py
```diff
@@ -46,6 +52,40 @@ def ocrd_cli_wrap_processor(
...
+
+        # IP and port but without database
+        if server_ip and server_port and not mongo_url:
+            raise click.UsageError('--mongo-url is missing.')
+
+        # Proceed when both IP and port are provided
+        initLogging()
+
+        # Init a processor instance to get access to its information
+        processor = processorClass(workspace=None)
```
bertsky, line 72: This will not always work. The existing `Processor` constructor runs a `ParameterValidator`, which (depending on the ocrd-tool.json of the processor, notably if there are `required` params without `default`) can fail on `None`.

ocrd/ocrd/processor/helpers.py
```diff
@@ -216,6 +340,12 @@ def wrap(s):
                                   (with --page-id, remove only those)
   -p, --parameter JSON-PATH       Parameters, either verbatim JSON string
                                   or JSON file path
+  --server-ip IP                  Host name/IP to listen at. When this value is set,
+                                  --server-port must be set as well.
+  --server-port NUMBER            TCP port to listen at. When this value is set,
+                                  --server-ip must be set as well.
+  --mongo-url URL                 Connection string to a Mongo database. If the processor runs as a server,
+                                  this value must be set.
```
bertsky, line 348: That particular place in the help string (between parameter and param-override) is awkward. I suggest moving before the other non-processing options (show-resource, list-resources, dump-json, help, version).

ocrd/ocrd/decorators/__init__.py
```diff
@@ -46,6 +52,40 @@ def ocrd_cli_wrap_processor(
...
+        Config.title = processor.ocrd_tool['executable']
+        Config.description = processor.ocrd_tool['description']
+        Config.version = processor.version
+        Config.ocrd_tool = processor.ocrd_tool
+        Config.db_url = mongo_url
+
+        # Start the server
+        from ocrd.server.main import app
+        import uvicorn
+        uvicorn.run(app, host=server_ip, port=server_port, access_log=False)
```
bertsky, line 88: That's a strange pattern. IIUC you use these class attributes as a kind of cross-module global.

I know it's hard communicating Python objects from your existing runtime context (after parsing the OCR-D CLI) into the web server (I believe Uvicorn offers the same as Unicorn, including inheritance from BaseApplication...). But is that the proper Pythonic way?

ocrd/ocrd/server/models/job.py
```diff
@@ -0,0 +1,48 @@
...
+    path: str
+    description: str = None
+    state: StateEnum
+    input_file_grps: list[str]
+    output_file_grps: list[str]
+    page_id: str = None
+    parameters: dict
+
+    class Settings:
+        name = Config.collection_name
```
bertsky, line 47: What is _collection name_ for?

bertsky (commented):
Just a first glance from my side.

ocrd/ocrd/server/models/job.py
tdoan2010: It's the name of the collection in Mongo DB. The idea is, although all processors use the same database, each of them has its own collection.

So, if you start the `ocrd-dummy` server, it will work only in the collection called `ocrd-dummy`.

ocrd/ocrd/processor/helpers.py
tdoan2010: Agree.

ocrd/ocrd/decorators/ocrd_cli_options.py
```diff
@@ -26,6 +27,9 @@ def cli(mets_url):
         option('-O', '--output-file-grp', help='File group(s) used as output.', default='OUTPUT'),
         option('-g', '--page-id', help="ID(s) of the pages to process"),
         option('--overwrite', help="Overwrite the output file group or a page range (--page-id)", is_flag=True, default=False),
+        option('--server-ip', help='Host name/IP to listen at.'),
+        option('--server-port', help='TCP port to listen at', type=click.INT),
+        option('--mongo-url', help='Connection string to a Mongo database.', type=click.STRING),
```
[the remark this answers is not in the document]
tdoan2010: I have nothing against it, just curious, why is it better than 3 options? Because it's shorter?

ocrd/ocrd/processor/helpers.py
```diff
@@ -163,6 +175,142 @@ def run_cli(
...
+    # check output file groups are in METS
+    for output_file_grp in output_file_grps:
+        if output_file_grp not in workspace.mets.file_groups:
+            log.error(
+                f'Invalid state: expected output file group "{output_file_grp}" not in METS (despite processor success)')
+
+    log.debug('Finish processing')
+
+    # Save the job status to the database
+    from ocrd.server.models.job import Job, StateEnum
```
kba, line 202: late import again, why?

ocrd/ocrd/processor/helpers.py
```diff
@@ -163,6 +175,142 @@ def run_cli(
...
+    await job.save()
+
+
+@lru_cache(maxsize=32)
+def get_processor(parameter_str: str, processor_class=None):
+    """
+    Call this function to get back an instance of a processor. The results are cached based on the parameters.
+    The parameters must be passed as a string because
+    `dict <https://docs.python.org/3/tutorial/datastructures.html#dictionaries>`_ is unhashable,
+    therefore cannot be cached.
```
kba, line 298: Perhaps using `frozendict` like [this](https://stackoverflow.com/a/53394430/201318) might be a solution. The drawback of stringifying dict is that there is no guarantee two equivalent dicts are serialized in the same way (order of keys).

ocrd/ocrd/server/models/job.py
```diff
@@ -0,0 +1,46 @@
...
+    running = 'RUNNING'
+    success = 'SUCCESS'
+    failed = 'FAILED'
+
+
+class JobInput(BaseModel):
+    path: str
+    description: Optional[str] = None
+    input_file_grps: List[str]
+    output_file_grps: Optional[List[str]]
```
kba, line 19: Why is `output_file_grp` optional but `input_file_grps` is not?

ocrd/ocrd/server/models/job.py
```diff
@@ -0,0 +1,46 @@
...
+                "description": "The description of this execution",
+                "input_file_grps": ["INPUT_FILE_GROUP"],
+                "output_file_grps": ["OUTPUT_FILE_GROUP"],
+                "page_id": "PAGE_ID",
+                "parameters": {}
+            }
+        }
+
+
+class Job(Document):
```
kba, line 36: `started` and `finished` timestamps might be useful too.

ocrd/ocrd/decorators/__init__.py
```diff
@@ -46,6 +53,39 @@ def ocrd_cli_wrap_processor(
...
+
+        # Proceed when both IP and port are provided
+        initLogging()
+
+        # Init a processor instance to get access to its information (also warm up the cache with default parameters)
+        params = {}
+        processor = get_processor(json.dumps(params), processorClass)
+
+        # Start the server
+        from ocrd.server.main import ProcessorAPI
```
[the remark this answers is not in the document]
tdoan2010: The first implementation was kind of stupid from me. That's why this late import was there. I removed it already.

ocrd/ocrd/server/models/job.py
tdoan2010: Is there a case when a processor can run without input file groups?

ocrd/ocrd/processor/helpers.py
```diff
@@ -163,6 +181,99 @@ def run_cli(
...
+    input_file_grps_str = ','.join(input_file_grps)
+    output_file_grps_str = ','.join(output_file_grps)
+
+    job = await Job.get(job_id)
+    job.state = StateEnum.running
+    job.start_time = datetime.now()
+
+    # Execute the processor
+    return_code = run_cli(executable, workspace=workspace, page_id=page_id, input_file_grp=input_file_grps_str,
+                          output_file_grp=output_file_grps_str, parameter=json.dumps(parameter), mets_url='')
```
joschrew, line 197: Hi, I think I maybe found a bug running a processing-server with ocrd-olena-binarize (this is a bashlib-processor). I think `mets_url` should be changed from `''` to `workspace.mets_target`, at least that is what resolves the problem in my case.
I think to reproduce it is enough to "simply" run the processing server with ocrd-olena-binarize and call it.

joschrew:
Hi, I am/was not able to call a processing server running for ocrd-olena-binarize (a bashlib processor). I think it is because of a bug in the processing server in helpers.py in function `run_cli_from_api`. In line 196 and 197, `run_cli` is called with `mets_url=''`` which should be changed to `mets_url=workspace.mets_target` or at least that resolved my problem.
To reproduce I did nothing special, just run the server as described and call it as described.

tdoan2010:
@joschrew thank you for your comment. I pushed the fix as you suggested.

Issue #886, opened 2022-07-13
ocrd workspace merge: add fileId-mapping and pageId-mapping


bertsky:
In `ocrd workspace merge` we currently have `--fileGrp-mapping`, but nothing to prevent conflicts among file IDs or page IDs. (The latter is only relevant when merging documents page-wise, but the former can always happen.)

kba:
Closing since you  have helpfully added the functionality in #888 and it was merged and released yesterday.

PR #888, opened 2022-07-13
Improve workspace merge

### LINKED ISSUE #887: ocrd workspace merge: file href paths must be re-based
Currently, when using `ocrd workspace merge --no-copy-files`, then the path names are kept unchanged. But the other METS' location is (almost always) different, so they must be rebased to the current METS.

bertsky:
fixes #887 and #886

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -573,15 +573,23 @@ def set_id(ctx, id):   # pylint: disable=redefined-builtin
 # ocrd workspace merge
 # ----------------------------------------------------------------------
 
+def _handle_json_option(ctx, param, value):
+    return parse_json_string_or_file(value) if value else None
+
 @workspace_cli.command('merge')
 @click.argument('METS_PATH')
 @click.option('--copy-files/--no-copy-files', is_flag=True, help="Copy files as well", default=True, show_default=True)
-@click.option('--fileGrp-mapping', help="JSON object mapping src to dest fileGrp")
+@click.option('--fileGrp-mapping', help="JSON object mapping src to dest fileGrp", callback=_handle_json_option)
+@click.option('--fileId-mapping', help="JSON object mapping src to dest file ID", callback=_handle_json_option)
+@click.option('--pageId-mapping', help="JSON object mapping src to dest page ID", callback=_handle_json_option)
 @mets_find_options
 @pass_workspace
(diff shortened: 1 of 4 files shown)


bertsky:
BTW, IMO ocrd workspace merge needs to be test-covered more. At least the `--no-copy-files` behaviour, perhaps also the new mappings.

Also, we could document useful recipes for the mappings like
```sh
ocrd workspace merge --no-copy-files --fileId-mapping <(
  echo '{'; 
  # prefix the file IDs with the name of the WS
  ocrd workspace -d otherws find -k ID | sed "s/.*/\"&\": \"otherws_&\"/;\$!s/\$/,/"; 
echo '}')  --pageId-mapping <(
  echo '{'; 
  # prefix the page IDs with the name of the WS
  ocrd workspace -d otherws list-page | sed "s/.*/\"&\": \"otherws_&\"/;\$!s/\$/,/"; 
  echo '}') otherws/mets.xml
```

kba:
> BTW, IMO ocrd workspace merge needs to be test-covered more. At least the `--no-copy-files` behaviour, perhaps also the new mappings.

I've added a basic test for `no_copy_files` and `fileId_mapping`.

BTW is `fileId_mapping` the most consistent naming? I don't really care either way but finishing #874 reminded me.

> 
> Also, we could document useful recipes for the mappings like
> 
> ```shell
> ocrd workspace merge --no-copy-files --fileId-mapping <(
>   echo '{'; 
>   # prefix the file IDs with the name of the WS
>   ocrd workspace -d otherws find -k ID | sed "s/.*/\"&\": \"otherws_&\"/;\$!s/\$/,/"; 
> echo '}')  --pageId-mapping <(
>   echo '{'; 
>   # prefix the page IDs with the name of the WS
>   ocrd workspace -d otherws list-page | sed "s/.*/\"&\": \"otherws_&\"/;\$!s/\$/,/"; 
>   echo '}') otherws/mets.xml
> ```

That is a neat pattern. Perhaps we should offer this as a general option `--prefix-file-id` or similar?

bertsky:
> BTW is `fileId_mapping` the most consistent naming? I don't really care either way but finishing #874 reminded me.

Probably not, I just extrapolated from the existing `fileGrp_mapping` / `--fileGrp-mapping`.

> That is a neat pattern. Perhaps we should offer this as a general option `--prefix-file-id` or similar?

If you think this is general enough, then yes, it _would_ save the user a lot of typing. (But then again, I like a single but very general mechanism better than multiple redundant mechanisms. In bulk-add, we just added some practical examples to the docstring, and documented even more cases in the Wiki.)

Issue #363, opened 2019-12-03
workspace bagger: in-place behaves odd


bertsky:
I am trying to fix bagit checksums in OCR-D/assets#64 with `ocrd zip bag -Z -I`:

- When I use `-d data`, then the bagger will move everything to `data/data`.
- When I omit `-d data` (i.e. effectively use `-d .` in the bagit directory), then the bagger won't find the mets.xml.
- When I use `-m data/mets.xml` this does not work either, because apparently the `--mets` option _in this command_ (in stark contrast to all the other commands and the workspace processor CLI) does not denote the (absolute or CWD-relative) METS path on the input side, but the workspace-relative METS path on the output side.

So how do I use this tool at all?

kba:
I will look into this, but in the meantime, if you want to update checksums, have a look at https://github.com/kba/ocrd-docs/blob/master/update-bagit It's what I use for assets

bertsky:
Also not behaving in a useful way: without `-I`, whatever is specified as `DEST` does **not** get used directly as output directory. Instead, a **new** subdirectory with a randomized name gets used (silently).

kba:
I will revisit this as soon as possible in the context of revamping our GT setup.

joschrew:
Hi, I recently came across recreating checksums for bagits and wanted to know if ocrd can do it. This way I came across this issue. I hope you don't mind me dug this issue up again and that I understand it correctly.

My first question is what your goal is? Is it to recreate/fix the checksums with ocrd or is there a problem with the `-I` switch (if so i don't understand that)? Because I think `-I` does exactly what is stated in the description (even if that is pretty useless): The description says: "Replace workspace with bag (like bagit.py does)". The switch replaces/deletes a workspace (mets and filegroups, just a workspace not a bag) and puts the bagit-stuff: data-dir (containing the workspace), bag-info.txt, manifest-sha512.txt etc to where the workspace has been.

What I would want to do/change is this: use `ocrd zip bag -I -d some/bag/dir foo/` and then the bag at `some/bag/dir` or rather the stuff in data/ and bag-info.txt is used to create a new bag which is put into `foo/`.

bertsky:
> My first question is what your goal is? 

For `bag -Z -I` it is simply to update the bag checksums (say after some changes to the METS and files).

> Is it to recreate/fix the checksums with ocrd or is there a problem with the `-I` switch (if so i don't understand that)? 

[...]

PR #899, opened 2022-08-09
workspace_bagger: fixes for bag and spill with bagits having mets-name other than mets.xml

joschrew:
currently bagging and spilling with bags not containing mets.xml but a different name for the mets don't work. This pr should fix this and make it possible to create bagits and to convert (spill) workspaces from bagits where the metsfile-name is changed and Ocrd-Mets is set in bag-info.txt

### DIFF
--- a/ocrd/ocrd/workspace_bagger.py
+++ b/ocrd/ocrd/workspace_bagger.py
@@ -6,7 +6,7 @@ from tempfile import mkdtemp
 import re
 import tempfile
 import sys
-from bagit import Bag, make_manifests  # pylint: disable=no-name-in-module
+from bagit import Bag, make_manifests, _load_tag_file  # pylint: disable=no-name-in-module
 
 from ocrd_utils import (
     pushd_popd,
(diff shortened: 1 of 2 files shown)


bertsky:
Which of the bagger issues should this fix exactly?

joschrew:
I didn't fill issues for that if you mean that with your question. Do I have to do that before I make a pull request?

There are ocrd-zipfiles which do not have a data/mets.xml but for example data/foo.xml which is the metsfile. In this cases in bag-info.txt `Ocrd-Mets` has to be set to `data/foo.xml`. This is valid according to the specs. But if you have such an ocrd-zipfile and try to spill it with workspace_bagger.py spill() with the current version of the code, it throws an error.
If you try to create such a ocrd-zip file with workspace_bagger.py bag() it does not work because `Ocrd-Mets` is not set in bag-info.txt to `data/foo.xml` as expected.

kba:
> I didn't fill issues for that if you mean that with your question. Do I have to do that before I make a pull request? Or maybe my this PullRequest's text/comment has to be changed?

No, we had a private discussion on this and I agreed that the behavior @joschrew observed is a bug and asked him to fix it.

> 
> There are ocrd-zipfiles which do not have a data/mets.xml but for example data/foo.xml which is the metsfile. In this cases in bag-info.txt `Ocrd-Mets` has to be set to `data/foo.xml`. This is valid according to the specs. But if you have such an ocrd-zipfile and try to spill it with workspace_bagger.py spill() with the current version of the code, it throws an error. If you try to create such a ocrd-zip file with workspace_bagger.py bag() it does not work because `Ocrd-Mets` is not set in bag-info.txt to `data/foo.xml` as expected.

:+1:

kba (approved):

bertsky:
Thanks for explaining. Sounds a lot like https://github.com/OCR-D/core/issues/363 to me.

PR #904, opened 2022-08-18
Resource manager respecting processor's moduledir

kba:
This implements proper support for processor's moduledir:

* Processors have a new CLI option `--dump-module-dir` that prints the `moduledir` of the processor
* When downloading a model
  * unless `--location` is explicitly overridden, resmgr will inspect the `resource_location` in a processor's `ocrd-tool.json` and select the first one
  * if the location is `module`, will get the directory via `--dump-module-dir`
* For `list-installed`, looks up the directory via `--dump-module-dir` and will look in it as well.

This also removes the `find_resources` method which is mostly redundant with `list-available`

It is a draft because I introduced a bug where downloading a model duplicates all the entries for that executable. Will have to debug this further tomorrow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Versioned according to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [2.38.0] - 2022-08-14
+
 Fixed:
 
   * `ocrd zip`: Properly respect `Ocrd-Mets`, #899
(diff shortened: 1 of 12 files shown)


kba:
> It is a draft because I introduced a bug where downloading a model duplicates all the entries for that executable. Will have to debug this further tomorrow.

Debugged and ready to go.

PR #912, opened 2022-09-15
ocrd_cli_wrap_processor: no sys.argv or sys.exit

### LINKED ISSUE #911: ocrd_cli_wrap_processor: reference to sys.argv is wrong
https://github.com/OCR-D/core/blob/71d295ac1fccbeb4164e230bd584e1920b9ab3c8/ocrd/ocrd/decorators/__init__.py#L39-L41

This makes the CLI wrapper behave non-functionally. It _usually_ still works (because click itself of course uses `sys.argv`), but fails with things like `click.testing.CliRunner`.

IMO the only correct way of testing for empty arguments is by checking all named kwargs for non-emptiness and then `kwargs` itself.

bertsky:
Fixes #911

### DIFF
--- a/ocrd/ocrd/decorators/__init__.py
+++ b/ocrd/ocrd/decorators/__init__.py
@@ -1,4 +1,5 @@
 from os.path import isfile
+import sys
 
 import click
 
@@ -38,7 +39,7 @@ def ocrd_cli_wrap_processor(
     if not (mets or working_dir or 
             dump_json or dump_module_dir or help or version or show_resource or list_resources):
         processorClass(workspace=None, show_help=True)
-        return 1
+        sys.exit(1)
     if dump_json or dump_module_dir or help or version or show_resource or list_resources:
         processorClass(
             workspace=None,
@@ -49,7 +50,7 @@ def ocrd_cli_wrap_processor(
             show_resource=show_resource,
             list_resources=list_resources
         )
-        return
+        sys.exit(0)
     else:
         initLogging()
         LOG = getLogger('ocrd_cli_wrap_processor')
--- a/ocrd/ocrd/decorators/__init__.py
+++ b/ocrd/ocrd/decorators/__init__.py
@@ -1,5 +1,4 @@
 from os.path import isfile
-import sys
 
 import click
 
@@ -36,9 +35,10 @@ def ocrd_cli_wrap_processor(
     list_resources=False,
     **kwargs
 ):
-    if not sys.argv[1:]:
+    if not (mets or working_dir or 
+            dump_json or dump_module_dir or help or version or show_resource or list_resources):
         processorClass(workspace=None, show_help=True)
-        sys.exit(1)
+        return 1
     if dump_json or dump_module_dir or help or version or show_resource or list_resources:
         processorClass(
             workspace=None,
@@ -49,7 +49,7 @@ def ocrd_cli_wrap_processor(
             show_resource=show_resource,
             list_resources=list_resources
         )
-        sys.exit()
+        return
     else:
         initLogging()
         LOG = getLogger('ocrd_cli_wrap_processor')


ocrd/ocrd/decorators/__init__.py
```diff
@@ -36,9 +35,10 @@ def ocrd_cli_wrap_processor(
     list_resources=False,
     **kwargs
 ):
-    if not sys.argv[1:]:
+    if not (mets or working_dir or
```
kba, line 38: Agreed, that is more robust.

ocrd/ocrd/decorators/__init__.py
```diff
@@ -36,9 +35,10 @@ def ocrd_cli_wrap_processor(
     list_resources=False,
     **kwargs
 ):
-    if not sys.argv[1:]:
+    if not (mets or working_dir or 
+            dump_json or dump_module_dir or help or version or show_resource or list_resources):
         processorClass(workspace=None, show_help=True)
-        sys.exit(1)
```
kba, line 41: Why remove the sys.exit? They can still be caught (`except SystemExit`) in tests but with a return value, processors will need to adapt their CLI wrapping code.

kba (changes requested):

ocrd/ocrd/decorators/__init__.py
bertsky: Yes, I was probably mistaken about that. CliRunner surely can handle (i.e. catches) SystemExit, and merely returning may not be the right thing in all contexts.

bertsky: I have reverted that bit.

bertsky:
I'm puzzled why `test_processor_no_mets` no longer works. It seems to fall through in `ocrd_cli_wrap_processor` to the processing case when calling `tests.base.TestCase.invoke_cli` with `[]` as args. But that should just delegate to the decorated's [main](https://click.palletsprojects.com/en/7.x/api/?highlight=clirunner#click.BaseCommand.main), which should parse all kwargs with their defaults and thus make the first case condition true...

kba:
> I'm puzzled why test_processor_no_mets no longer works.

Yeah, can reproduce and share the confusion. Debugging it now.

kba:
OK, just realized that I introduced the sys.argv construct in 77fd2b487 for a reason, now trying to remember what reason that was.

kba:
Now I remember. The problem is that we have a default value `mets.xml` for the `-m/--mets` argument. That means that `mets` kwarg is always set to something and we cannot use it to make sure that there were actually no explicit arguments to the CLI call - except to use `sys.argv[1:]`.

So, I think the version before this PR is actually more robust.

Which test cases fail for you with `sys.argv`?

bertsky:
> Now I remember. The problem is that we have a default value `mets.xml` for the `-m/--mets` argument. 

ah, yes, of course (in `ocrd_cli_options`)! 

> That means that `mets` kwarg is always set to something and we cannot use it to make sure that there were actually no explicit arguments to the CLI call - except to use `sys.argv[1:]`.

Well, we could still change the behaviour (at least in Python, better not in CLI) to use `None` as default and give that a semantic which falls back to `mets.xml`. But I would not recommend it.

> So, I think the version before this PR is actually more robust.

Probably, yes. 

But it's relying on a side effect. I wonder whether Click does not offer any better alternative. (Should be some hook between its parsing and default filling...)

> Which test cases fail for you with `sys.argv`?

See CI and discussion and commits preceding https://github.com/OCR-D/ocrd_tesserocr/pull/187#issuecomment-1269085693

ocrd/ocrd/decorators/__init__.py
kba: No, it [is not](https://github.com/OCR-D/core/pull/912#issuecomment-1286988322)

kba:
> But it's relying on a side effect. I wonder whether Click does not offer any better alternative. (Should be some hook between its parsing and default filling...)

Yes, this is not ideal and can be improved. As the only remaining change in this PR is `sys.exit(0)` vs `sys.exit(1)`. I propose to close the PR but keep the issue in case we find the time to find a more robust solution.

PR #914, opened 2022-09-29
Makefile: make default args actually changeable

joschrew:
While trying to build images with ocrd-core I noticed that default args are not working / can not be overridden. For example it was not possible to set the docker-tag when building a core-image. I think the question mark was just forgotten.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,16 +46,16 @@ help:
 # END-EVAL
 
 # Docker tag. Default: '$(DOCKER_TAG)'.
-DOCKER_TAG = ocrd/core
+DOCKER_TAG ?= ocrd/core
 
 # Docker base image. Default: '$(DOCKER_BASE_IMAGE)'.
-DOCKER_BASE_IMAGE = ubuntu:18.04
+DOCKER_BASE_IMAGE ?= ubuntu:18.04
 
 # Additional arguments to docker build. Default: '$(DOCKER_ARGS)'
-DOCKER_ARGS = 
+DOCKER_ARGS ?=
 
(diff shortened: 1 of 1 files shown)


bertsky (approved):
Note: This is not strictly necessary. 

Even without your change, you can override variable assignments by passing them to the `make` cmdline, e.g.

    make DOCKER_BASE_IMAGE=ubuntu:20.04 docker

Crucially, this is different from the variable assignments of the shell, which can be used in a makefile, but normally do not override.

With your change (called conditional assignment operator), even the shell variables take precedence, so you can _also_ do:

    DOCKER_BASE_IMAGE=ubuntu:20.04 make docker

See `info make "Setting Variables"`

joschrew:
Thanks, didn't know about `make DOCKER_BASE_IMAGE=ubuntu:20.04 docker` I thought `DOCKER_BASE_IMAGE=ubuntu:20.04 make docker` is the (only) way to do it.

kba:
Thanks @joschrew for reporting this. AFAICS the change in assignment operator seems unnecessary, so leave them be?

BTW, unless there is a strong reason, such as a particular shell or use in advanced scripts, I would always use the `make NAME=value` over the `NAME=value make` mechanism, the latter makes it hard to distinguish between `make` and shell variables and behaves differently depending on whether `export` is set. To avoid all of that, I normally use the simple `=` assignment and override where necessary with `make NAME=value`.

joschrew:
No problem for me to close this pr, because I just didn't know that it works and is intended the way it is currently.

PR #917, opened 2022-10-05
Processor.module: handle namespace pkg case

### LINKED ISSUE #916: regression in processor.resolve_resource: module may be a namespace package only
Since #904 we look up … https://github.com/OCR-D/core/blob/71d295ac1fccbeb4164e230bd584e1920b9ab3c8/ocrd/ocrd/processor/base.py#L246-L250 … to determine the moduledir location. 

But that fails if the processor module itself is distributed via PEP-420 namespace packages (so the top level does not contain any resources).

For example, in case of ocrd-eynollah-segment, we have `qurator.eynollah` as the actual processor module, but `qurator` itself is empty:

```
  File "/data/ocr-d/ocrd_all/venv/lib/python3.7/site-packages/qurator/eynollah/processor.py", line 47, in process
    'dir_models': self.resolve_resource(self.parameter['models']),
  File "/data/ocr-d/ocrd_all/venv/lib/python3.7/site-packages/ocrd/processor/base.py", line 219, in resolve_resource
    cwd=cwd, moduled=self.moduledir)
  File "/data/ocr-d/ocrd_all/venv/lib/python3.7/site-packages/ocrd/processor/base.py", line 257, in moduledir
    return resource_filename(self.module, '')
  File "/data/ocr-d/ocrd_all/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1154, in resource_filename
    return get_provider(package_or_requirement).get_resource_filename(
  File "/data/ocr-d/ocrd_all/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 361, in get_provider
    return _find_adapter(_provider_factories, loader)(module)
  File "/data/ocr-d/ocrd_all/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1403, in __init__
    self.module_path = os.path.dirname(getattr(module, '__file__', ''))
  File "/usr/lib/python3.7/posixpath.py", line 156, in dirname
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I don't know what's the correct way to deal with this, but it seems to suffice checking that a `.__file__` exists at the module level – for namespaces this is `None`.

bertsky:
fixes #916

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -247,7 +247,16 @@ class Processor():
         """
         The top-level module this processor belongs to.
         """
-        return self.__module__.split('.')[0]
+        # find shortest prefix path that is not just a namespace package
+        fqname = ''
+        for name in self.__module__.split('.'):
+            if fqname:
+                fqname += '.'
+            fqname += name
+            if sys.modules[fqname].__file__:
+                return fqname
+        # fall-back
+        return self.__module__
 
     @property
     def moduledir(self):


bertsky:
CI failure is only on macos, and [very strange](https://app.circleci.com/pipelines/github/OCR-D/core/1389/workflows/9d0c400e-630c-4761-b4a6-bc8e0702245b/jobs/4616/steps): all tests seem to pass, but the overall status is still failure.

I don't have permissions to rerun on CircleCI.

kba (commented):
Had not considered namespace packages as used by @qurator-spk, also for the binarization and the old textline detector. Elegant solution LGTM

kba (approved):

ocrd/ocrd/processor/base.py
```diff
@@ -247,7 +247,16 @@ def module(self):
...
+        # find shortest prefix path that is not just a namespace package
+        fqname = ''
+        for name in self.__module__.split('.'):
+            if fqname:
+                fqname += '.'
+            fqname += name
+            if sys.modules[fqname].__file__:
+                return fqname
+        # fall-back
+        return self.__module__
```
bertsky, line 259: @kba unfortunately, this still does not seem to work:

```
ocrd resmgr list-installed -e ocrd-eynollah-segment
  File "/bin/ocrd-eynollah-segment", line 8, in <module>
    sys.exit(main())
  File "/lib/python3.6/site-packages/click/core.py", line 1134, in __call__
  File "/lib/python3.6/site-packages/click/core.py", line 1059, in main
    rv = self.invoke(ctx)
  File "/lib/python3.6/site-packages/click/core.py", line 1401, in invoke
  File "/lib/python3.6/site-packages/click/core.py", line 767, in invoke
  File "/lib/python3.6/site-packages/qurator/eynollah/ocrd_cli.py", line 8, in main
    return ocrd_cli_wrap_processor(EynollahProcessor, *args, **kwargs)
  File "/lib/python3.6/site-packages/ocrd/decorators/__init__.py", line 51, in ocrd_cli_wrap_processor
    list_resources=list_resources
  File "/lib/python3.6/site-packages/qurator/eynollah/processor.py", line 30, in __init__
    super().__init__(*args, **kwargs)
  File "/lib/python3.6/site-packages/ocrd/processor/base.py", line 109, in __init__
    print(self.moduledir)
  File "/lib/python3.6/site-packages/ocrd/processor/base.py", line 266, in moduledir
    return resource_filename(self.module, '')
  File "/lib/python3.6/site-packages/ocrd/processor/base.py", line 256, in module
    if sys.modules[fqname].__file__:
AttributeError: module 'qurator' has no attribute '__file__'
```

bertsky: see #985 for a fix

kba: Strange, I cannot reproduce this, I've tested it with 3.6 to 3.9 with fresh venvs, installed core and eynollah and

```python
import qurator
qurator.__file__
```

never raises `AttributeError` for `__file__`. It is not set because it's a namespace module but a module not having `__file__ at all seems very wrong to me. Are you sure this is not a broken installation or do you have an idea how to reproduce this?

bertsky: Are you sure you tried 3.6, too? I can see `__file__` from 3.7 onwards. (In other cases of namespace packages, I do get the attribute even in 3.6. Perhaps it depends on the concrete [method](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) of how it was declared?)

kba: I did not try it in a fresh 3.6 venv apparently. I did now and can reproduce (which took ages because of opencv compilation btw). Apparently namespace modules in Python <= 3.6 have only `['__doc__', '__loader__', '__name__', '__package__', '__path__', '__spec__']` as attributes.

So I can confirm it is a problem in 3.6 and while we do not support it anymore, there's no reason to actively break it. So thanks, I'll merge the PR!

PR #925, opened 2022-10-21
ocrd validate: accept --mets-basename, bashlib: use that

kba:
When calling a bashlib processor like `ocrd-olena-binarize` with a workspace backed by a METS not called `mets.xml`, this leads to an error because we call `ocrd validate tasks` at the end of the bashlb argument parsing and it only specifies the `--workspace` (directory) so fails to instantiate the workspace, or even uses the wrong file.

I would like to consolidate the options to a single `--mets` option like we did for the processors and `ocrd workspace` option, but that will probably best happen in conjunction with the METS server change.

### DIFF
--- a/ocrd/ocrd/cli/validate.py
+++ b/ocrd/ocrd/cli/validate.py
@@ -101,15 +101,17 @@ def validate_page(page, **kwargs):
 
 @validate_cli.command('tasks')
 @click.option('--workspace', nargs=1, required=False, help='Workspace directory these tasks are to be run. If omitted, only validate syntax')
+@click.option('-M', '--mets-basename', nargs=1, default='mets.xml', help='Baasename of the METS file, used in conjunction with --workspace')
 @click.option('--overwrite', is_flag=True, default=False, help='When checking against a concrete workspace, simulate overwriting output or page range.')
 @click.option('-g', '--page-id', help="ID(s) of the pages to process")
 @click.argument('tasks', nargs=-1, required=True)
-def validate_process(tasks, workspace, overwrite, page_id):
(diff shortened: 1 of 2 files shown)


ocrd/ocrd/cli/validate.py
```diff
@@ -101,15 +101,17 @@ def validate_page(page, **kwargs):
 
 @validate_cli.command('tasks')
 @click.option('--workspace', nargs=1, required=False, help='Workspace directory these tasks are to be run. If omitted, only validate syntax')
+@click.option('-M', '--mets-basename', nargs=1, default='mets.xml', help='Baasename of the METS file, used in conjunction with --workspace')
```
bertsky, line 104:
```suggestion
@click.option('-M', '--mets-basename', nargs=1, default='mets.xml', help='Basename of the METS file, used in conjunction with --workspace')
```

bertsky (approved):
Good catch!

We should probably also add some basename tests with the dummy processor (like `test_mets_basename` and `test_ocrd_dummy.TestDummyProcessor` combined).

Also, we have next to no tests for bashlib processing yet.

BTW, before ocrd-olena-binarize (and the rest of them) can actually handle the non-default basename case correctly, we would still have to rewrite all the `ocrd workspace` calls to something like `ocrd workspace -m $(basename ${ocrd__argv[mets_file]})`.

And so to facilitate that, in turn, we should probably add something like `ocrd__workspace` function to lib.bash (which would work both before and after the usual `cd` call)...

kba:
Merged with the basename typo fix, but amended, so GH thinks there's unmerged  changes here.

PR #932, opened 2022-10-24
add RESOURCE_LOCATIONS==module (for resmgr CLI)

bertsky:
Another glitch that went under the radar. It must be possible to choose `ocrd resmgr download -l module` (even if `module` may already be the default, but it should at least appear as a CLI option).


kba (approved):
Agreed that it makes sense for consistency's sake, but is it realistic? If a processor has `module` as the default, it might make sense to override that to install to one of the standard location. But when would you want it the other way around?

bertsky:
> but is it realistic? If a processor has `module` as the default, it might make sense to override that to install to one of the standard location. But when would you want it the other way around?

Cannot think of a case. But there is a third scenario: modules that do not have `module` as default, but you would still want to install new models into the `module` directory if possible (because you want all files to be in one place, or because there is a standalone use outside of OCR-D).

Issue #915, opened 2022-10-04
ocrd workspace remove: logging too noisy


bertsky:
A typical output of `ocrd workspace remove` looks like this:
```
today at 14:28:02Oct  4 12:28:01 ocrd-manager 12:28:01.985 WARNING ocrd.workspace.remove_file - File not locally available <OcrdFile fileGrp=FULLTEXT ID=FILE_0021_FULLTEXT, mimetype=text/xml, url=https://digital.slub-dresden.de/data/kitodo/LankDres_1760234508/LankDres_1760234508_ocr/00000021.xml, local_filename=---]/> 
today at 14:28:02Oct  4 12:28:01 ocrd-manager 12:28:01.989 INFO ocrd_models.ocrd_mets.remove_one_file - remove_one_file(<OcrdFile fileGrp=FULLTEXT ID=FILE_0021_FULLTEXT, mimetype=text/xml, url=https://digital.slub-dresden.de/data/kitodo/LankDres_1760234508/LankDres_1760234508_ocr/00000021.xml, local_filename=---]/> )
today at 14:28:02Oct  4 12:28:01 ocrd-manager 12:28:01.991 INFO ocrd_models.ocrd_mets.remove_one_file - Delete fptr element <Element {http://www.loc.gov/METS/}fptr at 0x7f3e566159c8> for page 'FILE_0021_FULLTEXT'
today at 14:28:02Oct  4 12:28:01 ocrd-manager 12:28:01.991 INFO ocrd.workspace.save_mets - Saving mets '/data/testdata-presentation/mets.xml'
```

(The same repeats for _every_ file.) IMO that's way too much.

Suggestions:
- suppress the non-download warning when using `--force` anyway
- downgrade the two ocrd_models.ocrd_mets.remove_one_file messages to loglevel DEBUG
- downgrade the ocrd.workspace.save_mets message to loglevel DEBUG

kba:
> suppress the non-download warning when using --force anyway

+1

> downgrade the two ocrd_models.ocrd_mets.remove_one_file messages to loglevel DEBUG

+1

> downgrade the ocrd.workspace.save_mets message to loglevel DEBUG

Also +1

PR #933, opened 2022-10-25
resmgr list-installed: downgrade log msg to debug

bertsky:
another one


kba (approved):

bertsky:
Your commit msg references the wrong issue IINM, it should be #915

kba:
> Your commit msg references the wrong issue IINM, it should be #915

Thanks for spotting, force-pushing.

PR #939, opened 2022-10-27
cli.resmgr: allow missing content-length

### LINKED ISSUE #924: resmgr download fails if server does not provide content-length
https://github.com/OCR-D/core/blob/71d295ac1fccbeb4164e230bd584e1920b9ab3c8/ocrd/ocrd/cli/resmgr.py#L132

This fails with [main.tar.gz](https://github.com/OCR-D/ocrd_tesserocr/blob/0427f6f31070bdc782a27e71cd6ac88142da9924/ocrd_tesserocr/ocrd-tool.json#L282-L289) in ocrd-tesserocr-recognize.

(In this case, the `size` is already known by the registered resource, but it should even work without that. So we have a bug and another one behind it.)

bertsky:
fixes #924


kba (approved):

kba:
Thanks, the `Content-Length` request is unnecessary in almost all cases.

Issue #922, opened 2022-10-12
ocrd workspace bulk-add: auto fileid should include filegrp


bertsky:
The normal convention for new file IDs in OCR-D is (due to `make_file_id` implementation) the pattern `grp + '_' + page`. But the current bulk-add behaviour automatically assigns `file_id` based on the input pattern directly:

https://github.com/OCR-D/core/blob/71d295ac1fccbeb4164e230bd584e1920b9ab3c8/ocrd/ocrd/cli/workspace.py#L304-L305

Not only does this usually omit the fileGrp, it also needlessly restricts IDs to only use _ASCII_ letters (where in fact all _Unicode_ letters should be allowed).

Also, in case the `file_path` is distinct from `src_path` (i.e. if the input expression is not a filename glob but a complex pattern), I believe the ID should be taken from the latter.

(The same is true for auto MIME type BTW.)

bertsky:
Luckily, (and surprisingly,) this seems to be the only usage of `safe_filename` to date. But IMO that function should be fixed first (by allowing alle Unicode codepoints from the letter class).

bertsky:
Fixed by #943

PR #943, opened 2022-11-02
bulk-add: Improve deriving file_id from other fields

kba:
As raised in #922, if `--file-id` is not provided explicitly, `bulk-add` derived it from the `safe_filename(file_path)` and this had three problems:

* `safe_filename` allowed only a subset of ASCII, now all Unicode letters and digits
* `file_id` was derived from `file_path` even when the "file_path" was actually CSV data. Now it either derives it from `file_path` if it is an actual filepath or from the `srcpath` if not. 
* Derived `file_id` did not follow the convention of starting with the `file_grp`. It does now and also only uses the basename of either `file_path` or `srcpath`

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -332,7 +332,7 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
 
         # derive --file-id from filename if not --file-id not explicitly set
         if not file_id:
-            id_field = srcpath.name if file_path != srcpath else file_path.name
+            id_field = srcpath.stem if file_path != srcpath else file_path.stem
             file_dict['file_id'] = safe_filename('%s_%s' % (file_dict['file_grp'], id_field))
         if not mimetype:
             try:
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -304,13 +304,6 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
         # set up file info
         file_dict = {'url': url, 'mimetype': mimetype, 'file_id': file_id, 'page_id': page_id, 'file_grp': file_grp}
 
-        # guess mime type
-        if not file_dict['mimetype']:
-            try:
-                file_dict['mimetype'] = EXT_TO_MIME[file_path.suffix]
-            except KeyError:
-                log.error("Cannot guess mimetype from extension '%s' for '%s'. Set --mimetype explicitly" % (file_path.suffix, file_path))
-
         # Flag to track whether 'url' should be 'src'
         url_is_src = False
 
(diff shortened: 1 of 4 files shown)


ocrd/ocrd/cli/workspace.py
```diff
@@ -323,6 +320,10 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
                 if param_name == 'url':
                     url_is_src = True
                     continue
+                elif param_name == 'file_id':
+                    # XXX file_id might be unset so we derive it once the other
+                    # replacements have happened
```
bertsky, line 325: We forgot to do the same for `--mimetype`. It is currently auto-filled above based on `file_path.suffix`, which is wrong. It should be `srcpath.suffix` below.

```suggestion
                elif param_name in ['mimetype', 'file_id']:
                    # auto-filled below once the other
                    # replacements have happened
```

ocrd/ocrd/cli/workspace.py
```diff
@@ -301,11 +301,8 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
...
 
-        # derive --file-id from filename if not --file-id not explicitly set
-        file_id_ = file_id or safe_filename(str(file_path))
-
         # set up file info
-        file_dict = {'url': url, 'mimetype': mimetype, 'file_id': file_id_, 'page_id': page_id, 'file_grp': file_grp}
+        file_dict = {'url': url, 'mimetype': mimetype, 'file_id': file_id, 'page_id': page_id, 'file_grp': file_grp}
 
         # guess mime type
         if not file_dict['mimetype']:
```
bertsky, line 308: Should be deleted here (see below).

ocrd/ocrd/cli/workspace.py
```diff
@@ -336,6 +337,11 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
         else:
             srcpath = file_path
 
+        # derive --file-id from filename if not --file-id not explicitly set
+        if not file_id:
+            id_field = srcpath.name if file_path != srcpath else file_path.name
+            file_dict['file_id'] = safe_filename('%s_%s' % (file_dict['file_grp'], id_field))
+
```
bertsky, line 344:
```suggestion
        if not mimetype:
            try:
                file_dict['mimetype'] = EXT_TO_MIME[srcpath.suffix]
            except KeyError:
                log.error("Cannot guess MIME type from extension '%s' for '%s'. Set --mimetype explicitly" % (srcpath.suffix, srcpath))

```

bertsky (changes requested):
We're nearly there, see suggestions.

Also: the help text for `--file-id` should get something like `If not provided, guess from filename`, too.

ocrd/ocrd/cli/workspace.py
```diff
@@ -336,6 +337,11 @@ def workspace_cli_bulk_add(ctx, regex, mimetype, page_id, file_id, url, file_grp
         else:
             srcpath = file_path
 
+        # derive --file-id from filename if not --file-id not explicitly set
+        if not file_id:
+            id_field = srcpath.name if file_path != srcpath else file_path.name
```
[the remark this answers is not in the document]
kba: Good point. So `.stem` rather than `.name`.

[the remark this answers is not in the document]
kba: a4a7ab3fb

ocrd/ocrd/cli/workspace.py
kba: 958a7f9a0

bertsky (approved):
Thanks!

bertsky:
> Also: the help text for `--file-id` should get something like `If not provided, guess from filename`, too.

That's on L221 BTW

kba:
> > Also: the help text for `--file-id` should get something like `If not provided, guess from filename`, too.
> 
> That's on L221 BTW

Added in 4de1cdc0b, thanks for being thorough!

PR #944, opened 2022-11-03
Cache functionality kba

kba:
@MehmedGIT These are for #875.

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_mets.py
+++ b/ocrd_models/ocrd_models/ocrd_mets.py
@@ -105,18 +105,14 @@ class OcrdMets(OcrdXmlDocument):
         tree_root = self._tree.getroot()
 
         # Fill with files
-        el_fileGrp_list = tree_root.find(".//mets:fileSec", NS)
-        if el_fileGrp_list is None or len(el_fileGrp_list) == 0:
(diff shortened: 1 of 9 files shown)


ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -210,8 +210,8 @@ def file_groups(self):
         """
 
         # WARNING: Actually we cannot return strings in place of elements!
```
MehmedGIT, line 212: As the warning warns, we cannot just return a list of strings and use it as a list of fileGrp memory objects. However, I am also a bit confused because all tests seem to pass successfully which was not the case when I first implemented the caching mechanism. It should rather throw an exception saying that the string object does not support a call of a fileGrp method.

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -210,8 +210,8 @@ def file_groups(self):
...
-        #if self._cache_flag:
-        #    return self._file_cache.keys()
+        if self._cache_flag:
+           return list(self._file_cache.keys())
```
MehmedGIT, line 214: This returns a list of the keys() of the `_file_cache` and the keys are of type `String` not of type `el_fileGrp`.

PR #947, opened 2022-11-08
make install: workaround for broken shapely wheels

bertsky:


kba (approved):
Until https://github.com/shapely/shapely/issues/1598 is fixed, this is the best way to prevent the segfault, so LGTM.

kba:
Ah, for the CI we need to install `libgeos-dev`, right?

bertsky:
> Ah, for the CI we need to install `libgeos-dev`, right?

Yes – sorry, should have checked right away.

PR #949, opened 2022-11-09
improve/complete bashlib description, add APIdoc refs

bertsky:


ocrd_models/README.md
```diff
@@ -2,48 +2,50 @@
...
 2. Edit `ocrd_models/ocrd_page_user_methods.py` and append to the `METHOD_SPECS` list:
 
-```python
-METHOD_SPECS = (
-  # ...
-  _add_method(r'^PageType$', 'get_FirstTextRegion')
-  # ...
-)
-```
+    ```python
```
kba, line 18: Why the additional indentation?

kba (approved):
Thank you, improving the documentation is always appreciated.

Formerly (with the `<!-- BEGIN-RENDER ...` macros), the bashlib API was generated from the comments in `ocrd/ocrd/lib.bash`. I see no need to keep that but maybe your more detailed comments would be worth copying to bashlib?

bertsky:
> Formerly (with the `<!-- BEGIN-RENDER ...` macros), the bashlib API was generated from the comments in `ocrd/ocrd/lib.bash`. 

Yes. It needed to be completed and updated. (And the parser apparently did not catch everything, anyway.)

> I see no need to keep that but maybe your more detailed comments would be worth copying to bashlib?

Perhaps so. But then are the hyperlinks into other parts of the documentation still gonna work?

Also, if the sh parser is not robust enough, then I'd rather keep everything in one place only.

On the other hand, if we had something like sphinx-apidoc / sphinx_click for the bashlib, that would be much better of course. (We could everything generated in one place.)

ocrd_models/README.md
kba: Ah, of course, I was only watching the changed README instead of the original state :)

kba:
> Yes. It needed to be completed and updated. (And the parser apparently did not catch everything, anyway.)

> Also, if the sh parser is not robust enough, then I'd rather keep everything in one place only.

The problem is that by default, shrender expects comments to be extracted from bash to start with `## ` (to distinguish these docstring-like comments from normal comments).

> But then are the hyperlinks into other parts of the documentation still gonna work?

Yes, the docstring-like comments have `## ` removed but otherwise are unchanged.

But really, it isn't worth the effort, `lib.bash` is fairly readable as it is and it's more important to have good documentation in the README.

PR #951, opened 2022-11-13
Update bagit-checksums / fix for #363

joschrew:
Hi,
this PR contains the functionality to recreate bagit-checksums (tagmanifest-sha512.txt, manifest-sha512.txt, Payload-Oxum in bag-info.txt ). This is done with a new command: `ocrd zip update`.
The second thing this PR does is remove the `-I` flag from `ocrd zip bag`, because, as seen in the issue #363 the `-I`-flag is hard to understand and does not work as expected.

### DIFF
--- a/ocrd/ocrd/cli/zip.py
+++ b/ocrd/ocrd/cli/zip.py
@@ -118,10 +118,14 @@ def validate(src, **kwargs):
 @zip_cli.command('update')
 @click.argument('src', type=click.Path(dir_okay=True, readable=True, resolve_path=True), required=True)
 @click.argument('dest', type=click.Path(dir_okay=True, readable=True, writable=True, resolve_path=True), required=False)
-@click.option('-', '--overwrite', help="overwrite bag in SRC", type=bool, default=False)
+@click.option('-o', '--overwrite', help="overwrite bag in SRC", is_flag=True)
 def update(src, dest=None, overwrite=False):
     """
-    Recretae checksums of an OCRD-ZIP
+    Recreate files containing checksums (manifest-sha512.txt, tagmanifest-sha512.txt and
+    'Payload-Oxum' contained in bag-info.txt) of an OCRD-ZIP.
 
+    Open the bag (zip file or directory) ``src``, create or update its manifests/checksums and
+    output to (zip file or directory) ``dest``. It is also possible to output to ``src`` / overwrite
+    ``src`` in place when ``--overwrite``-flag is given.
     """
     WorkspaceBagger(Resolver()).recreate_checksums(src, dest=dest, overwrite=overwrite)
--- a/ocrd/ocrd/cli/zip.py
+++ b/ocrd/ocrd/cli/zip.py
@@ -110,3 +110,18 @@ def validate(src, **kwargs):
     print(report)
     if not report.is_valid:
         sys.exit(1)
+
+# ----------------------------------------------------------------------
+# ocrd zip update
+# ----------------------------------------------------------------------
+
+@zip_cli.command('update')
+@click.argument('src', type=click.Path(dir_okay=True, readable=True, resolve_path=True), required=True)
+@click.argument('dest', type=click.Path(dir_okay=True, readable=True, writable=True, resolve_path=True), required=False)
+@click.option('-', '--overwrite', help="overwrite bag in SRC", type=bool, default=False)
+def update(src, dest=None, overwrite=False):
+    """
+    Recretae checksums of an OCRD-ZIP
+
+    """
+    WorkspaceBagger(Resolver()).recreate_checksums(src, dest=dest, overwrite=overwrite)
--- a/ocrd/ocrd/workspace_bagger.py
+++ b/ocrd/ocrd/workspace_bagger.py
@@ -1,13 +1,15 @@
 from datetime import datetime
 from os import makedirs, chdir, walk
-from os.path import join, isdir, basename, exists, relpath, isfile
+from os.path import join, isdir, basename, exists, relpath
+from pathlib import Path
 from shutil import make_archive, rmtree, copyfile, move
-from tempfile import mkdtemp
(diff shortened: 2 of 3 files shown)


bertsky (changes requested):

ocrd/ocrd/cli/zip.py
```diff
@@ -39,13 +39,12 @@ def zip_cli():
               help='Basename of the METS file.',
               show_default=True)
 @click.option('-i', '--identifier', '--id', help="Ocrd-Identifier", required=True)
-@click.option('-I', '--in-place', help="Replace workspace with bag (like bagit.py does)", is_flag=True)
```
kba, line 42: This is a backwards-incompatible change. AFAICT the two aspects of the PR - dropping `--in-place` vs `ocrd zip update` - seem independent, so maybe we can have a separate PR for dropping `--in-place`, so we can warn potential users like @tboenig.

kba (changes requested):

joschrew:
Thanks @bertsky for your suggestions. With my commit 819c6bfc0 I tried to apply them (but I have unfortunately forgotten to answer here, sorry).

Thanks for your suggestions @kba. I tried to use all of them except the one for copytree with my last commit (2606ad689). I changed all file-function calls from os to pathlib, removed the removement of the `-I`-Flag, with-statement is now used for temporary-directory-creation and I changed the two exception-classes mentioned to be more specific.

shutil.copytree: shutil.copytree does not work for me because I wanted to support creating the bag in an existing folder. shutil.copytree() prior to python 3.8 does not allow copying to an existing directory that's why I choose copy_tree from distutils.dir_utils. This is why I'd rather not change it.

ocrd/ocrd/cli/zip.py
bertsky: Dropping `--in-place` is backwards incompatible indeed, but it does **not break** anything that has not **already**  been broken – is what Jonas is arguing IMHO correctly here, see https://github.com/OCR-D/core/issues/363.

bertsky:
> With my commit [819c6bf](https://github.com/OCR-D/core/commit/819c6bfc084f71ca6e8e69bb1aa3e077fa9b20bb) I tried to apply them (but I have unfortunately forgotten to answer here, sorry).

@joschrew FYI suggestions can be applied **directly** on GH. You can choose between single commits and batched commits. The advantage is two-fold: Contributors get credited as commit author, and the change request gets directly resolved in the review (and so reviewers don't have to come back looking what did change since their review and compare it to their suggestion, and everyone can focus on the outstanding requests/comments).

joschrew:
Thanks for pointing that out. I knew it was possible, but wasn't aware of the advantages. Will try to follow it from now on.

ocrd/ocrd/workspace_bagger.py
```diff
@@ -1,12 +1,15 @@
...
 from shutil import make_archive, rmtree, copyfile, move
-from tempfile import mkdtemp
+from tempfile import mkdtemp, TemporaryDirectory
 import re
 import tempfile
 import sys
-from bagit import Bag, make_manifests, _load_tag_file  # pylint: disable=no-name-in-module
+from bagit import Bag, make_manifests, _load_tag_file, _make_tag_file, _make_tagmanifest_file  # pylint: disable=no-name-in-module
+from distutils.dir_util import copy_tree
+from contextlib import nullcontext
```
bertsky, line 12: Even if we do not _support_ 3.6 anymore, we should not try to actively _break_ things unnecessarily, forcing users to update at that particular time.
```suggestion
try:
    from contextlib import nullcontext
except ImportError:
    from contextlib import contextmanager
    @contextmanager
    def nullcontext(enter_result=None):
        yield enter_result
```

bertsky (changes requested):

ocrd/ocrd/cli/zip.py
```diff
@@ -110,3 +108,18 @@ def validate(src, **kwargs):
...
+# ocrd zip update
+# ----------------------------------------------------------------------
+
+@zip_cli.command('update')
+@click.argument('src', type=click.Path(dir_okay=True, readable=True, resolve_path=True), required=True)
+@click.argument('dest', type=click.Path(dir_okay=True, readable=True, writable=True, resolve_path=True), required=False)
+@click.option('-', '--overwrite', help="overwrite bag in SRC", type=bool, default=False)
+def update(src, dest=None, overwrite=False):
+    """
+    Recretae checksums of an OCRD-ZIP
```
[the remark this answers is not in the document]
bertsky: > What is this refering to?

I have already forgotten, sry.

ocrd/ocrd/workspace_bagger.py
joschrew: I agree to try to not break compatibility if it can be avoided. But tbh I find that import error handling "suboptimal" and hard to read. What do you think about just removing the nullcontext? Then there will be cases when a temporary directory is created but not used but it gets deleted afterwards anyway:
```
with TemporaryDirectory() as tempdir:
```
(I tried, but I think it is not possible to response with a change suggestion on another location here)

bertsky: oh, absolutely! Did not see it myself. Should definitely work, much better that way.

joschrew: I removed the nullcontext as discussed with 6d13bbe4d. (btw: I don't see a possibility to commit parts of your suggestion or to somehow use github-tools to do this commit with better reference to this discussion. Please correct me if I am wrong and there is a better way to make this code-change)

ocrd/ocrd/cli/zip.py
kba: Since it isn't working, nobody will be inconvenienced by removing it, cannot argue with that logic, so LGTM.

kba: PR for dropping `--in-place` follows.

Issue #802, opened 2022-02-15
ocrd workspace validate: strange behaviour with symbolic links


stefanCCS:
If you have an workspace, which has symbolic links in Image Source Folder (OCR-D-IMG), then it looks like that the command
`ocrd workspace validate`
creates an copy for this image additional to the symbolic link.
This is only the case, if the symbolic link has an absolute path (not a relative one).

Before validate:
![grafik](https://user-images.githubusercontent.com/61045115/154033514-3fa6aa15-7b0b-4fc5-ae6f-5060070b4d26.png)

After validate:
![grafik](https://user-images.githubusercontent.com/61045115/154033411-a8a5e05d-ba61-4826-bf32-a0a263d84701.png)

Example workspace:
[reproduceStrangeValidate.zip](https://github.com/OCR-D/core/files/8067700/reproduceStrangeValidate.zip)

stefanCCS:
Hint: Maybe this is related to "`workspace add`" and not "`validate`" --> I am not that sure anymore ...

kba:
I suspect that during validation, data is persisted at the wrong place. We had that problem before with [spurious `TEMP` folders inside workspaces after validation](https://github.com/OCR-D/core/issues/324). But I haven't yet investigated, will update when I do.

mexthecat:
I think the problem is in workspace.py on line 146

`url_path = Path(f.url).resolve()`

If the file url is with symbolic links, resolve will return the real path and this is usually not in the workspace directory. 

any chance to replace resolve with absolute?

`url_path = Path(f.url).absolute()`

stefanCCS:
It looks like, that `workspace add` does NOT create a copy of the (image) file in addition to the link anymore (using version 2.32).

But now I have the same problem using `ocrd-cis-ocropy-binarize`.

--> please, clarify

bertsky:
Thanks @stefanCCS for the report and @mexthecat for the analysis!

IIUC `resolve()` must indeed be replaced by `absolute`, and that this causes not only the validator but also all processors to behave the same way.

@kba maybe we should also cover this in the tests by using symlinks somewhere in assets.

mexthecat:
I've created a pull request

https://github.com/OCR-D/core/pull/954

Maybe I made it to easy - I use this patch for quit some time - couldn't see any strange behaviour.

kba:
Fixed in the v2.42.0. Thanks @mexthecat!

stefanCCS:
[...]

PR #954, opened 2022-11-16
https://github.com/OCR-D/core/issues/802

mexthecat:
I use this patch for quit some time - couldn't see any strange behaviour.


kba (approved):
Thanks for this long-needed fix. AFAICT this works fine. I've added a test with the data from #802 and if that passes, we can merge.

PR #956, opened 2022-11-22
:fire: require python >= 3.7

kba:
In the Dockerfile, install python3.7 from the deadsnakes PPA instead of the 18.04 standard python3.6 and stop testing for python3.6.

With the updated base image, we can then tackle the update in ocrd_all.


bertsky:
GH already tells me that on their side (GH Actions), the ubuntu:18.04 base image is deprecated now. I don't know about Circle CI. 

But we do have the problem that most of our `deps-ubuntu` rules (in various places) do not differentiate between e.g. `python3-dev` and (say) `python3.7-dev`. Why should they? Because Ubuntu 18's default is actually python-3.6-dev, which is wrong.

So I guess I am saying we should **also** upgrade the base image now. Surprising myself there, but my guess is the other way is more painful overall.

bertsky (approved):

kba:
> So I guess I am saying we should also upgrade the base image now. Surprising myself there, but my guess is the other way is more painful overall.

Agreed, let's make the jump now. I'll adapt.

Dockerfile
```diff
@@ -19,21 +19,22 @@ COPY README.md .
...
 RUN apt-get update && apt-get -y install software-properties-common \
-    && add-apt-repository ppa:deadsnakes/ppa \
     && apt-get update && apt-get -y install \
         ca-certificates \
-        python3.7-dev \
-        python3.7-venv \
+        python3-dev \
+        python3-pip \
+        python3-venv \
+        gcc \
```
bertsky, line 27: Oh? Why is this necessary in core now? (AFAICS we have it ocrd_all/Makefile's `deps-ubuntu` → `deps-ubuntu-modules` → `CUSTOM_DEPS`.)

bertsky (approved):

Dockerfile
kba: `gcc` and `python3-dev` because we need to install shapely from source in core itself, otherwise CI would fail at `make install`, wouldn't it? `python3-venv` and `python3-pip` are probably not necessary, I'll check.

bertsky: right, of course! I forgot that core needs to be self-serving. (Should perhaps get a comment though so we can remove again when not needed.)

.circleci/config.yml
```diff
@@ -80,8 +80,8 @@ workflows:
               only: master
   test-pull-request:
     jobs:
-      - test-python36
       - test-python37
       - test-python38
       - test-python39
+      - test-python310
       - test-macos
```
stweil, line 87: Should Python 3.11 be added, too?

.circleci/config.yml
```diff
@@ -16,39 +16,39 @@ jobs:
       - run: make deps-test install PIP=pip3
       - run: make test benchmark PYTHON=python3
 
-  test-python36:
+  test-python37:
     docker:
-      - image: python:3.6.12
+      - image: python:3.7.9
```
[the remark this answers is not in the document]
kba: Sure, c35a753b8

.circleci/config.yml
kba: Yes 8698fc2ff

PR #958, opened 2022-11-29
Fix regressions from caching refactoring

kba:
In the process of merging #875 we introduced subtle bugs. These were not picked up because not all the fixtures in `test_ocrd_mets` had the caching on/off mechanism and also because we did not do `isinstance` tests.


bertsky (approved):
Many thanks!

MehmedGIT (approved):
Seems fine.

PR #971, opened 2023-01-15
resmgr: insert from tool instead of append

bertsky:
(otherwise dedup will stick to the existing entries)


bertsky:
CI failure: only macos (unrelated)

bertsky:
> (otherwise dedup will stick to the existing entries)

to the relevance of this: `ocrd resmgr download -o ...` does not work without it

kba (approved):
Makes sense, thanks!

PR #972, opened 2023-01-16
Processor builder outside the run_processor + cached version

MehmedGIT:
I have created a separate method to get the desired processor class. Both cached and non-cached types are supported. By default, the non-cached processor is returned to preserve the current state. It's possible to try the cached version by setting the additional flag parameter to True.

### DIFF
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -13,7 +13,7 @@ from memory_profiler import memory_usage
 from sparklines import sparklines
 
 from click import wrap_text
-from ocrd import Workspace
+from ocrd.workspace import Workspace
 from ocrd.processor import Processor
 from ocrd_utils import freeze_args, getLogger
 
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -7,11 +7,15 @@ from functools import lru_cache
 import json
 import inspect
 from subprocess import run, PIPE
+from typing import List, Type
+
 from memory_profiler import memory_usage
 from sparklines import sparklines
 
 from click import wrap_text
-from ocrd_utils import getLogger, freeze_args
+from ocrd import Workspace
+from ocrd.processor import Processor
+from ocrd_utils import freeze_args, getLogger
 
 
 __all__ = [
@@ -19,6 +23,7 @@ __all__ = [
     'run_cli',
     'run_processor'
 ]
+
 
 def _get_workspace(workspace=None, resolver=None, mets_url=None, working_dir=None):
     if workspace is None:
@@ -78,6 +83,7 @@ def run_processor(
     log.debug("Running processor %s", processorClass)
 
     processor = get_processor(
+        # TODO: Warning: processorClass of type Object gets auto casted to Type[Processor]
         processor_class=processorClass,
         parameter=parameter,
         workspace=workspace,
@@ -276,7 +282,7 @@ Default Wiring:
 # TODO: Decide how much maxsize is optimal as a default
 @freeze_args
 @lru_cache(maxsize=32)
-def get_cached_processor(parameter: dict, processor_class=None):
+def get_cached_processor(parameter: dict, processor_class: Type[Processor]):
     """
     Call this function to get back an instance of a processor.
     The results are cached based on the parameters.
@@ -294,13 +300,13 @@ def get_cached_processor(parameter: dict, processor_class=None):
 
 
 def get_processor(
-        processor_class,
+        processor_class: Type[Processor],
         parameter: dict,
-        workspace=None,
-        ocrd_tool=None,
-        page_id=None,
-        input_file_grp=None,
-        output_file_grp=None,
+        workspace: Workspace = None,
+        ocrd_tool: dict = None,
+        page_id: str = None,
+        input_file_grp: List[str] = None,
+        output_file_grp: List[str] = None,
         cached_processor: bool = False,
 ):
     if processor_class:
@@ -314,13 +320,12 @@ def get_processor(
             cached_processor.input_file_grp = input_file_grp
             cached_processor.output_file_grp = output_file_grp
             return cached_processor
-        else:
-            return processor_class(
-                workspace,
-                ocrd_tool=ocrd_tool,
-                page_id=page_id,
-                input_file_grp=input_file_grp,
-                output_file_grp=output_file_grp,
-                parameter=parameter
-            )
+        return processor_class(
+            workspace=workspace,
+            ocrd_tool=ocrd_tool,
+            page_id=page_id,
+            input_file_grp=input_file_grp,
+            output_file_grp=output_file_grp,
+            parameter=parameter
+        )
     raise ValueError("Processor class is not known")
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -65,7 +65,7 @@ def run_processor(
     - :py:attr:`output_file_grp`
     - :py:attr:`parameter` (after applying any :py:attr:`parameter_override` settings)
 
-    Warning: Avoid setting the `cached_processor` flag to True. It may have unexpected side effects.
+    Warning: Avoid setting the `instance_caching` flag to True. It may have unexpected side effects.
     This flag is used for an experimental feature we would like to adopt in future.
 
     Run the processor on the workspace (creating output files in the filesystem).
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -81,6 +81,7 @@ def run_processor(
         processor_class=processorClass,
         parameter=parameter,
         workspace=workspace,
+        ocrd_tool=ocrd_tool,
         page_id=page_id,
         input_file_grp=input_file_grp,
         output_file_grp=output_file_grp,
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -3,6 +3,8 @@ Helper methods for running and documenting processors
 """
 from os import environ
 from time import perf_counter, process_time
+from functools import lru_cache, wraps
+from frozendict import frozendict
 import json
 import inspect
 from subprocess import run, PIPE
@@ -42,6 +44,7 @@ def run_processor(
         parameter=None,
         parameter_override=None,
         working_dir=None,
+        cached_processor=False
 ): # pylint: disable=too-many-locals
     """
     Instantiate a Pythonic processor, open a workspace, run the processor and save the workspace.
@@ -73,14 +76,17 @@ def run_processor(
     )
     log = getLogger('ocrd.processor.helpers.run_processor')
     log.debug("Running processor %s", processorClass)
-    processor = processorClass(
-        workspace,
-        ocrd_tool=ocrd_tool,
+
+    processor = get_processor(
+        processor_class=processorClass,
+        parameter=parameter,
+        workspace=workspace,
         page_id=page_id,
         input_file_grp=input_file_grp,
         output_file_grp=output_file_grp,
-        parameter=parameter
+        cached_processor=cached_processor
     )
+
     ocrd_tool = processor.ocrd_tool
     name = '%s v%s' % (ocrd_tool['executable'], processor.version)
     otherrole = ocrd_tool['steps'][0]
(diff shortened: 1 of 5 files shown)


ocrd/ocrd/processor/helpers.py
```diff
@@ -263,3 +269,71 @@ def wrap(s):
     ocrd_tool.get('input_file_grp', 'NONE'),
     ocrd_tool.get('output_file_grp', 'NONE')
 )
+
+
+# Taken from https://github.com/OCR-D/core/pull/884
+def freeze_args(func):
```
kba, line 276: This seems generic enough to move it to `ocrd_utils.introspect`.

bertsky (commented):
A few comments from my side: 

So you are caching multiple processor instances, giving them time to parse parameters and load models into memory. I suppose this is all meant for the single-worker model carved out by Triet? I am asking because in multi-threading/processing it will be problematic to "steal" vital runtime objects like `Process.workspace` and `Process.input_file_grp` from the existing/cached instances. But in any case IMO you still need some function for cache cleaning. (And for GPU-enabled processors, disallowing caching or setting the cache size at runtime in order to prevent GPURAM OOM might be necessary, too.)

Furthermore, comparing to my own [refactoring and networking](https://github.com/OCR-D/core/pull/652) from 2yrs ago, a couple of differences stick out. (Omitting the networking side:)
- your PR offers multiple instances each by their own parameters instead of just one
- my PR also exposes `run_api` for what in your case would be `run_processor(...cached_processor=True)`; I recommend offering the same interface (i.e. `get_processor` plus `run_api`
- my PR catches runtime exceptions and allows to address them differently via `run_api`
- my PR encapsulates the CWD changes necessary during processing (storing the old CWD prior to the run, and restoring it afterwards, even in the case of failure); I believe your current draft will fail resolving input files when caching is allowed
- my PR also addresses the `task_sequence.ProcessorTask` API: it tries to find the class and instance from the executable name (this part is naturally messy), then uses the actual Processor instance for parameter validation (and even avoiding `--dump-json` callouts); not sure you want this, but then clearly the Processor class itself should expose parameter valdiation (currently only in the constructor); also, IMO it is worthwhile on your side already thinking about how to backport the `task_sequence.run_tasks` functionality (i.e. `ocrd process`) for processing servers (as a minimalistic/fallback/prototype use-case)

MehmedGIT:
Hi @bertsky. The idea of this PR is to refactor the `run_processor` method and hide the instantiation of an OCR-D processor behind a separate method. I thought it would be bad to just use the cached version from Triet's PR (https://github.com/OCR-D/core/pull/884) and decided to make it flexible - keep the default as it is now but still provide a way to use the cached version. 

Currently, it's not possible to get a processor instance outside of `run_processor` and we need such a method in the network package we are implementing. The [ocr-d network](https://github.com/OCR-D/core/tree/dev-processing-broker/ocrd/ocrd/network) to realize the architecture proposed by Triet. Inside the network package we have our `ProcessingWorker` agent which indeed will have it's own wrappers for `run_cli` and `run_processor`, e.g. `run_cli_from_worker` and `run_processor_from_worker`, to address the things you mentioned in bullet points. 
We are trying to implement this entire networking thing without altering the existing and working code base as much as possible.

ocrd/ocrd/processor/helpers.py
```diff
@@ -73,14 +80,19 @@ def run_processor(
     )
     log = getLogger('ocrd.processor.helpers.run_processor')
     log.debug("Running processor %s", processorClass)
-    processor = processorClass(
-        workspace,
+
+    processor = get_processor(
+        # TODO: Warning: processorClass of type Object gets auto casted to Type[Processor]
```
kba, line 85: How do you mean? Is the class "downgraded" to a `ocrd.Processor`?

kba (approved):
Except for the caching, which we will need to revisit once the [`ocrd_network` package](https://github.com/OCR-D/core/tree/dev-processing-broker) is refactored, this looks like a relatively small change.

I would propose to create a draft PR for the `ocrd_network` stuff soon and transfer the points raised by @bertsky to the discussion of that, so we can continue discussing with the context of the full changeset.

ocrd/ocrd/processor/helpers.py
MehmedGIT: The comment is not relevant anymore if no typing is provided for the `processorClass`. I should have removed it. I was not sure if the auto-casting may potentially be problematic here since my IDE highlighted it as a potential problem.

MehmedGIT:
Regarding the draft PR for the ocrd network package - that will come sometime soon. We are first trying to implement a working prototype before opening it for discussion.

kba:
> Regarding the draft PR for the ocrd network package - that will come sometime soon. We are first trying to implement a working prototype before opening it for discussion.

Sure, no problem, I just want to make sure we don't lose the discussion thread in a closed PR.

bertsky:
> The idea of this PR is to refactor the `run_processor` method and hide the instantiation of an OCR-D processor behind a separate method. I thought it would be bad to just use the cached version from Triet's PR (#884) and decided to make it flexible - keep the default as it is now but still provide a way to use the cached version.

ah, sry, should have commented in #884 first. (I thought I could home in from the "outside" ;-)

ok, so here you merely wanted to provide an uncached version out of that, too.

> Currently, it's not possible to get a processor instance outside of `run_processor` and we need such a method in the network package we are implementing. 

But you do have `get_processor` already!

> The [ocr-d network](https://github.com/OCR-D/core/tree/dev-processing-broker/ocrd/ocrd/network) to realize the architecture proposed by Triet. Inside the network package we have our `ProcessingWorker` agent which indeed will have it's own wrappers for `run_cli` and `run_processor`, e.g. `run_cli_from_worker` and `run_processor_from_worker`, to address the things you mentioned in bullet points.

Ok, I think I understand now. I like `run_*_from_worker` better than `run_*_from_api` BTW. (To me, API would by default stand for Python API, not Web API.)

So, I can already see most of my above points solved in #884:
- CWD switching 
- job validation prior to actual runtime
- exception handling (logging + conditionally re-exposing)

But I am still missing:
- resetting CWD in case of failure (in my implementation, resetting CWD was part of the inner `finally` clause; but in your implementation, the CWD stuff is in `run_processor` and the exception handling happens in the caller)
- some function for (processor instance) cache cleaning. (And for GPU-enabled `Processor` heirs, disallowing such caching or setting the cache size at runtime in order to prevent GPURAM OOM)

MehmedGIT:
> But you do have get_processor already!

If I directly reuse the `get_processor` from #884 then it may end up (potentially) causing problems (which you already covered some) because of the caching. 

> But in any case IMO you still need some function for cache cleaning. (And for GPU-enabled processors, disallowing caching or setting the cache size at runtime in order to prevent GPURAM OOM might be necessary, too.)

You said this in your previous comment, didn't you? Exactly, to avoid currently unknown for us scenarios I did not just reuse the already available `get_processor` directly! Instead, I renamed that method to `get_cached_processor` and wrapped it in a method `get_processor` which based on a flag returns the cached/non-cached processor. Then reused that method inside `run_processor`.

> Ok, I think I understand now. I like run_*_from_worker better than run_*_from_api BTW. (To me, API would by default stand for Python API, not Web API.)

Yeah, I am usually trying to select names that potentially won't confuse others. For the same reason I like `processing worker` more than `processing server` when we are referring to an agent that is in fact just an ocr-d processor which is not exposed to the user directly, but indirectly through a `processing broker` aka `processing server` with the new term.

> But I am still missing:

I am still not sure whether you completely get the idea behind the PR. It's not supposed to provide a full working extension for `run_processor` that covers all requirements you have mentioned. The main question is: Does this PR break the core in any way? Of course, assuming that the `cached_processor` flag of `run_processor` is never set to True. If yes, let me know what is problematic and missing? 

One thing now I see I have missed is to place a documentation comment that says to not set this flag. This will prevent others getting confused and thinking it's safe to be set. And, probably, that's why you are confused a bit. Another potential reason for your confusion is probably the fact that #884 was dropped and won't be merged to core. It's there only for reference to me and Jonas while the `ocr-d network` package is being implemented.

bertsky:
> If I directly reuse the `get_processor` from #884 then it may end up (potentially) causing problems (which you already covered some) because of the caching.
> ... to avoid currently unknown for us scenarios I did not just reuse the already available get_processor directly! Instead, I renamed that method to `get_cached_processor` and wrapped it in a method `get_processor` which based on a flag returns the cached/non-cached processor. Then reused that method inside run_processor.

I can see that. What I asked for is some additional mechanism which can then be used via Web API for resetting the cache (using `get_cached_processor.cache_clear()`) and changing the size (using `get_cached_processor.cache_info().maxsize = ...`).

(Also, perhaps `instance_caching` would be a better kwarg than `cached_processor`?)

> > But I am still missing: ...
> 
> I am still not sure whether you completely get the idea behind the PR. It's not supposed to provide a full working extension for `run_processor` that covers all requirements you have mentioned. The main question is: Does this PR break the core in any way? Of course, assuming that the `cached_processor` flag of `run_processor` is never set to True. If yes, let me know what is problematic and missing?
> 
> One thing now I see I have missed is to place a documentation comment that says to not set this flag. This will prevent others getting confused and thinking it's safe to be set. And, probably, that's why you are confused a bit. Another potential reason for your confusion is probably the fact that #884 was dropped and won't be merged to core. It's there only for reference to me and Jonas while the `ocr-d network` package is being implemented.

I am not confused (any more). I have already stated that these latter points are about the direction #884 is going. And since we are already discussing them here, and AFAICS there is no other PR to discuss `dev-processing-broker` and `ocrd-webapi-implementation` yet, why not?

MehmedGIT:
> What I asked for is some additional mechanism which can then be used via Web API for resetting the cache (using get_cached_processor.cache_clear()) and changing the size (using get_cached_processor.cache_info().maxsize = ...).

Triet implemented and tested the caching. I just rely on that currently. If you know (from experience) how the suggested functionalities are to be implemented in the easiest way, feel free to share. I have not checked that yet. And currently, I am busy with higher-priority tasks (still related to the ocr-d network package).

> AFAICS there is no other PR to discuss dev-processing-broker 

Because it's not ready to be discussed yet.

> and ocrd-webapi-implementation yet, why not?

That was an initial proof-of-concept implementation for the WebAPI started by Jonas. Then, in order to extend and test its usability, we integrated that into Operandi and continued the development there. Feel free to open issues in that repo if you see something odd related to the WebAPI.

In the end, everything (#884, ocrd-webapi-implementation, rabbitmq-utils, and server/worker topology) will go under the same hood - ocr-d network package.

MehmedGIT:
> (Also, perhaps instance_caching would be a better kwarg than cached_processor?)

Agree, I have changed it.

> And since we are already discussing them here, and AFAICS there is no other PR to discuss dev-processing-broker and ocrd-webapi-implementation yet, why not?

We are almost there with the PR for the Processing Server. The current draft PR: https://github.com/OCR-D/core/pull/974

> I can see that. What I asked for is some additional mechanism which can then be used via Web API for resetting the cache (using get_cached_processor.cache_clear()) and changing the size (using get_cached_processor.cache_info().maxsize = ...).

Providing endpoints for that is not a problem, but isn't this something to be handled internally and not through the Web API? I am still not sure how should the `.cache_clear()` or changing the `.cache_info().maxsize = ...` should be handled by an already running processing worker. There may be other complications arising from modifying the cache from the outside. I think we should further discuss this in #974 once the PR is ready. 

The #972 should now be ready to be merged.

ocrd/ocrd/processor/helpers.py
```diff
@@ -58,6 +65,9 @@ def run_processor(
     - :py:attr:`output_file_grp`
     - :py:attr:`parameter` (after applying any :py:attr:`parameter_override` settings)
 
+    Warning: Avoid setting the `cached_processor` flag to True. It may have unexpected side effects.
```
bertsky, line 68:
```suggestion
    Warning: Avoid setting the `instance_caching` flag to True. It may have unexpected side effects.
```

bertsky (approved):

bertsky:
> > What I asked for is some additional mechanism which can then be used via Web API for resetting the cache (using get_cached_processor.cache_clear()) and changing the size (using get_cached_processor.cache_info().maxsize = ...).
> 
> Providing endpoints for that is not a problem, but isn't this something to be handled internally and not through the Web API? 

I stand corrected. For management of processor instances, no Web API or web anything is necessary. If it's the Processing Worker (in the queue model) or the Processing Server (in the integrated model), that component can just control its internal cache references via Python API.

> I am still not sure how should the `.cache_clear()` or changing the `.cache_info().maxsize = ...` should be handled by an already running processing worker.

Ideally, it enforces some (preconfigured) max-workers size, and updates the cache whenever instances crash. For GPU workers, the memory consumption must also be watched – but by adding the envvar, you already gave the admin some control.

> The #972 should now be ready to be merged.

Agreed.

kba:
CI failure is only for macosx, merging.

PR #977, opened 2023-02-03
OcrdMets.add_agent: insert before metsDocumentID

### LINKED ISSUE #976: add_agent creates invalid METS if metsHdr exists
If we import/clone a METS which already contains a metsHdr with some content (agent or metsDocumentID), then the following …

https://github.com/OCR-D/core/blob/ee92cfc084b8e4f2dd5be2f7016f957c87ae1573/ocrd_models/ocrd_models/ocrd_mets.py#L216

… may create a subelement ordering that is invalid (agent after metsDocumentID).

bertsky:
fixes #976 (but you may want to add a test case, featuring some document with existing `mets:metsDocumentID` and optionally also `mets:agent`s)


kba (approved):

kba:
added a test, works as expected, thank you @bertsky

PR #978, opened 2023-02-06
bashlib: fix ocrd__files if no input-files

bertsky:


kba:
> Just found that older branch, seems I forgot to create a PR for it.

Thanks, good catch

> IIRC, we dropped wrap.bash in the meantime, right?

Yes, no more `bashlib/src`, just `lib.bash`.

PR #979, opened 2023-02-08
WorkspaceValidator mets_files: allow files without pageId

### LINKED ISSUE #485: workspace validator: allow files not representing a single page
Currently, the workspace validator will complain about any kind of file (including derived images) that is not contained in the structMap as physical page.

IMO this is an error on the side of the validator: Only files **representative** of a complete page (like derived images processed _on the page level_, but not on the region, line or word level) should have a pageId.

bertsky:
Fixes #485


kba (approved):
Make sense, thanks.

kba:
CI fails for macos with missing libmagic. Triggered a re-run but fail is unrelated I think, will merge it now.

PR #980, opened 2023-02-09
OcrdMets.find_files: don't force @LOCTYPE=OTHER

bertsky:
This fixes a hitherto undetected bug where any use of `OcrdMets.find_files` (by instantiating an OcrdFile with the default `loctype=OTHER`) would invariably change the `@LOCTYPE` of the respective files to `OTHER` and their `@OTHERLOCTYPE` to `FILE` (even if they are still URLs).


kba (approved):
Good catch. I'll need to revisit this once the two-urls-per-file PR is ready but until then, this fixes a bug, I'll merge it.

Issue #982, opened 2023-02-14
ocrd from pypi: failed to find magic


bertsky:
I get…

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/magic/__init__.py", line 209, in <module>
    libmagic = loader.load_lib()
  File "/usr/local/lib/python3.6/dist-packages/magic/loader.py", line 49, in load_lib
    raise ImportError('failed to find libmagic.  Check your installation')
ImportError: failed to find libmagic.  Check your installation
```

I already got `imagemagick` from apt and `magick` from pip. Seems we now need some additional library. I thought we made this optional (for casual users who don't enter via our Makefile's `deps-ubuntu`)?

bertsky:
Ah, that's not related to IM, but `libmagic-dev`, which we now need for MIME type guessing:

https://github.com/OCR-D/core/blob/ee92cfc084b8e4f2dd5be2f7016f957c87ae1573/ocrd/ocrd/resource_manager.py#L320

bertsky:
Why don't we just trust our EXT_TO_MIME rules here?

bertsky:
How about the (dep-free) Python pkg [filetype](https://pypi.org/project/filetype/)?

(For suffix based detection, how about also conferring builtin [mimetypes](https://docs.python.org/3/library/mimetypes.html)?)

kba:
> https://storage.googleapis.com/drive-bulk-export-anonymous/20230310T105258.886+0000/4133399871716478688/ffca26bf-0e9c-4e1a-8757-395da4fb2a80/1/d8dffe73-127e-4ce7-8ebf-7c95f9220783?authuser

I'm getting a 403 for that, what HTTP headers does it return for you, esp. `Content-Type`? Then we could use that info.

But yes, https://github.com/h2non/filetype.py looks like a good  dependency-free alternative, I'll investigate.

bertsky:
> I'm getting a 403 for that,

[...]

PR #984, opened 2023-02-16
Drop dependency on libmagic

kba:
As raised in #982, I inadvertently introduced a system requirement (libmagic) when implementing support for ZIP files in the resource manager. This PR replaces that logic with a simple lookup of file extensions of the URL in the `EXT_TO_MIME` dict.


ocrd_utils/ocrd_utils/constants.py
```diff
@@ -42,8 +42,11 @@
     '.ppm': 'image/x-portable-pixmap',
     '.pnm': 'image/x-portable-anymap',
     '.pbm': 'image/x-portable-bitmap',
+    '.tar.gz': 'application/gzip',
+    '.tar.xz': 'application/x-xz',
```
bertsky, line 46:
```suggestion
    '.tar.gz': 'application/gzip',
    '.tar.xz': 'application/x-xz',
    '.tgz': 'application/gzip',
    '.txz': 'application/x-xz',
```

bertsky (approved):

Issue #983, opened 2023-02-14
Docker build: opencv-python-headless fails


bertsky:
Our last Docker release is now 9 months old. It seems there is a [persistent error](https://app.circleci.com/pipelines/github/OCR-D/core/1581/workflows/58b2759d-e91c-4c57-a720-b1686f64e261/jobs/5574) around the OpenCV Build:
```
ERROR: Failed building wheel for opencv-python-headless
```

Perhaps Python 3.6 is not supported anymore (no wheels on PyPI)? Or do we need more [system dependencies](https://docs.opencv.org/4.x/d2/de6/tutorial_py_setup_in_ubuntu.html) now?

bertsky:
@kba your [CI config change](https://github.com/OCR-D/core/commit/27427e22116eb0a866c79dc87c5435c08146a0dd) still does not work on the Scrutinizer chain – it for some reason needs `sudo` for `make deps-ubuntu`. Also, on `master` we [fail to deploy](https://app.circleci.com/pipelines/github/OCR-D/core/1605/workflows/6eeae1ef-a008-4163-9b58-9d8570852a3f/jobs/5705) awkwardly:
```
/bin/sh: 1: pip: not found
The command '/bin/sh -c apt-get update && apt-get -y install software-properties-common     && apt-get update && apt-get -y install         ca-certificates         python3-dev         python3-pip         gcc         make         wget         time         curl         sudo         git     && make deps-ubuntu     && pip install --upgrade pip setuptools     && make install     && apt-get remove -y gcc     && apt-get autoremove -y     && $FIXUP     && rm -rf /build-ocrd' returned a non-zero code: 127
Makefile:217: recipe for target 'docker-cuda' failed
```

So `make docker` runs through, but `docker-cuda` does not.

Regarding the original error reported here – that seems to have been specific to 3.6 (and I believe it would suffice to do a simple `pip install -U pip` first, to get `pip install cmake` working; so we could even reactivate our 3.6 CI...)

mikegerber:
I've been using this sometimes:

```
pip3 install -U pip
pip3 install --only-binary :all: opencv-python-headless
```

This installs an older opencv-python-headless, but from a wheel and without the compile time. I'd recommend always testing anything involving opencv-python-headless, we had some [tricky issues with it](https://github.com/qurator-spk/sbb_textline_detection/issues/60).

bertsky:
[...]

PR #986, opened 2023-02-17
upgrade pip3

bertsky:
attempts to fix the [docker-cuda](https://app.circleci.com/pipelines/github/OCR-D/core/1605/workflows/6eeae1ef-a008-4163-9b58-9d8570852a3f/jobs/5705) deploy failure.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ ENV DEBIAN_FRONTEND noninteractive
 ENV PYTHONIOENCODING utf8
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
+ENV PIP=pip3
 
 WORKDIR /build-ocrd
 COPY ocrd ./ocrd
@@ -31,7 +32,7 @@ RUN apt-get update && apt-get -y install software-properties-common \
         sudo \
         git \
     && make deps-ubuntu \
-    && pip install --upgrade pip setuptools \
+    && pip3 install --upgrade pip setuptools \
     && make install \
     && apt-get remove -y gcc \
     && apt-get autoremove -y \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,17 @@ jobs:
       - run: make deps-test install PIP=pip3
       - run: make test benchmark PYTHON=python3
 
+  test-python36:
+    docker:
+      - image: python:3.6.15
+    working_directory: ~/ocrd-core
+    steps:
+      - checkout
+      - run: apt-get -y update
+      - run: pip install -U pip
+      - run: make deps-ubuntu deps-test install
+      - run: make test benchmark
+
   test-python37:
     docker:
       - image: python:3.7.16
@@ -90,6 +101,7 @@ workflows:
               only: master
   test-pull-request:
     jobs:
+      - test-python36
       - test-python37
       - test-python38
       - test-python39
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,11 +10,14 @@ jobs:
       xcode: "12.5.1"
     # resource_class: macos.x86.medium.gen2
     working_directory: ~/ocrd-core
+    environment:
+      PIP: pip3
+      PYTHON: python3
     steps:
       - checkout
       - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install imagemagick geos
-      - run: make deps-test install PIP=pip3
-      - run: make test benchmark PYTHON=python3
+      - run: make install 
+      - run: make deps-test test benchmark
 
   test-python36:
     docker:
@@ -24,8 +27,8 @@ jobs:
       - checkout
       - run: apt-get -y update
       - run: pip install -U pip
-      - run: make deps-ubuntu deps-test install
-      - run: make test benchmark
+      - run: make deps-ubuntu install
+      - run: make deps-test test benchmark
 
   test-python37:
     docker:
@@ -34,8 +37,8 @@ jobs:
     steps:
       - checkout
       - run: apt-get -y update
-      - run: make deps-ubuntu deps-test install
-      - run: make test benchmark
+      - run: make deps-ubuntu install
+      - run: make deps-test test benchmark
 
   test-python38:
     docker:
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -31,7 +32,7 @@ RUN apt-get update && apt-get -y install software-properties-common \
         sudo \
         git \
     && make deps-ubuntu \
-    && pip install --upgrade pip setuptools \
+    && pip3 install --upgrade pip setuptools \
```
kba, line 35: I did try that already, that fixes this problem. But I stil fail locally to get opencv built.

bertsky: See [this comment](https://github.com/OCR-D/core/issues/983#issuecomment-1434607460)

bertsky:
Ok, it seems we got everything working here, except Scrutinizer, now failing for an outdated libgeos. Looks like the base image there is quite old.

bertsky:
> Looks like the base image there is quite old.

Wow! Just found out Scrutinizer still uses **Trusty** (Ubuntu 14)! I'll try [upgrading to Bionic](https://scrutinizer-ci.com/docs/guides/upgrading-trusty-image)...

bertsky:
@kba this resolves everything, at last!

bertsky:
Not so sure about _everything_ though: At least the Python 3.6 run looks [suspiciously short](https://app.circleci.com/pipelines/github/OCR-D/core/1611/workflows/6fcb30ce-b3c8-4ffd-8d70-505807ce948f/jobs/5747).

bertsky:
Note:
> 46a53ab

… is necessary because when deps-test comes first, opencv-python-headless and numpy are already pulled, and so 385c595 would be too late

> 276ab39

… is a first shot at a GHCR deployment. I'm not sure whether it works. (One condition is that in the project Settings > Actions > General > Workflow Permissions, you have allowed read+write for GITHUB_TOKEN of this repo.)

kba:
> Note:
> 
> > [46a53ab](https://github.com/OCR-D/core/commit/46a53ab4ef6d821b45e293eb3b5c80ba3d6b709a)
> 
> … is necessary because when deps-test comes first, opencv-python-headless and numpy are already pulled, and so [385c595](https://github.com/OCR-D/core/commit/385c5955d22d9a0971ef53ff65154b23439e2118) would be too late
> 
> > [276ab39](https://github.com/OCR-D/core/commit/276ab3906a9dc5a9c9d8e38687c6b6c215f06861)
> 
> … is a first shot at a GHCR deployment. I'm not sure whether it works. (One condition is that in the project Settings > Actions > General > Workflow Permissions, you have allowed read+write for GITHUB_TOKEN of this repo.)

Excellent. Yes `GITHUB_TOKEN` is allowed.

I'm testing the docker deployment - we also need to update the CUDA Ubuntu base image to 20.04, right?

kba:
> > I'm testing the docker deployment - we also need to update the CUDA Ubuntu base image to 20.04, right?
> 
> No we should already have that since [f1527a2](https://github.com/OCR-D/core/commit/f1527a2f605f3ea5ec89180b934291418ef8693e).
> 
> The CI VM itself uses ubuntu-latest.

https://github.com/OCR-D/core/pull/986/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R222 For `make docker-cuda` we currently use `11.3.1-cudnn8-runtime-ubuntu18.04` as the base image, should be using `11.3.1-cudnn8-runtime-ubuntu20.04` is what I meant.

bertsky:
> hould be using `11.3.1-cudnn8-runtime-ubuntu20.04` is what I meant.

Oh, right!

kba:
Unfortunately, I cannot seem to get `docker-cuda` build running.

The old 18.04/3.6 image (`11.3.1-cudnn8-runtime-ubuntu18.04`) fails because opencv-python-headless won't build.

With `11.3.1-cudnn8-runtime-ubuntu20.04`:

```
ldconfig
apt-get -y install --no-install-recommends cuda-runtime-10-0 cuda-runtime-10-1 cuda-runtime-10-2 cuda-runtime-11-0 cuda-runtime-11-1 cuda-runtime-11-3 libcudnn7    
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package cuda-runtime-10-0
E: Unable to locate package cuda-runtime-10-1
E: Unable to locate package cuda-runtime-10-2
E: Unable to locate package libcudnn7
make: *** [Makefile:234: cuda-ubuntu] Error 100
```

From running `nvidia/cuda:11.3.1-cudnn8-runtime-ubuntu20.04` directly, I can see that we need another `apt-get update` before installing. But:

```
root@b4480b7d3280:/# apt-cache search cuda-runtime
cuda-runtime-11-0 - CUDA Runtime 11.0 meta-package
cuda-runtime-11-1 - CUDA Runtime 11.1 meta-package
cuda-runtime-11-2 - CUDA Runtime 11.2 meta-package
cuda-runtime-11-3 - CUDA Runtime 11.3 meta-package
cuda-runtime-11-4 - CUDA Runtime 11.4 meta-package
cuda-runtime-11-5 - CUDA Runtime 11.5 meta-package
cuda-runtime-11-6 - CUDA Runtime 11.6 meta-package
cuda-runtime-11-7 - CUDA Runtime 11.7 meta-package
cuda-runtime-11-8 - CUDA Runtime 11.8 meta-package
cuda-runtime-12-0 - CUDA Runtime 12.0 meta-package
cuda-runtime-12-1 - CUDA Runtime 12.1 meta-package
```

and 

```
root@b4480b7d3280:/# apt-cache search libcudnn
libcudnn8 - cuDNN runtime libraries
libcudnn8-dev - cuDNN development libraries and headers
```

I am not sure about the ramifications of updating libcudnn and not supporting CUDA 10 runtime here - @bertsky?

bertsky:
> I am not sure about the ramifications of updating libcudnn and not supporting CUDA 10 runtime here - @bertsky?

Indeed, this will have grave implications for our TF 1.15 processors, probably also Detectron2. I recommend keeping core-cuda on the old (now defunct) image, tracking this in a new issue and merging here – we need the new release!

bertsky:
> we need the new release!

ok, this sort of defies the original intent of this PR. But the fixes here are mostly for non-cuda now....

Dockerfile
stweil: Upgrading to a container which is based on Ubuntu > 18.04 would also have solved the problem because then `python3-pip` installs `/usr/bin/pip`, too.

PR #974, opened 2023-01-24
Processing-Server

joschrew:
This pull request implements the processing server (https://github.com/OCR-D/spec/pull/222). The processing server starts a rabbitmq, mongodb and processing workers via ssh and docker. It receives calls to run ocr-d processors. The jobs are enqueued and the workers read the jobs from the queue and execute them. In the mongdb the jobs are stored and status changes of the jobs are reflected in the database and can be queried through the processing workers API.

To start a worker it must be defined in the configuration file. There the host of the worker has to be set. It is expected that the worker executables (e.g. ocrd-dummy) are available in the PATH. PATH can be modified via ~/.bash_profile or ~/.profile.

Example configuration file:
```
process_queue:
  address: localhost
  port: 5672
  credentials:
    username: testuser
    path_to_privkey: /home/testuser/.ssh/testuser.key
  ssh:
    username: testuser
    path_to_privkey: /home/testuser/.ssh/testuser.key
database:
  address: localhost
  port: 27018
  credentials:
    username: admin
    password: admin
  ssh:
    username: testuser
    path_to_privkey: /home/testuser/.ssh/testuser.key
hosts:
  - address: localhost
    username: testuser
    path_to_privkey: /home/testuser/.ssh/testuser.key
    workers:
      - name: ocrd-dummy
        number_of_instance: 1
        deploy_type: native
```

Example calls for the processing-server endpoints:

Run a processor:
`curl 'http://localhost:8080/processor/ocrd-dummy' -H 'accept: application/json' -H 'Content-Type: application/json' -d '{ "path": "/home/testuser/.local/share/ocrd-workspaces/data/mets.xml", "input_file_grps": ["OCR-D-IMG"], "output_file_grps": ["IMG-COPY-1"], "parameters": { "copy_files": true } }'`

Request processor status:
`curl 'http://localhost:8080/processor/ocrd-dummy/<insert-job-id-here>'`

List available processors:
`curl 'http://localhost:8080/processor'`

Get information about a single processor
`curl 'http://localhost:8080/processor/ocrd-dummy'`

PR #987, opened 2023-02-17
Fix the processor instance caching

MehmedGIT:
In #972, I missed using the context switch required for the cached processor instances to work correctly. Now I am able to reuse cached instances. According to my observations and testing, for both cached and non-cached, that additional line has no side effects.

### DIFF
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -1,7 +1,7 @@
 """
 Helper methods for running and documenting processors
 """
-from os import environ
+from os import chdir, environ, getcwd
 from time import perf_counter, process_time
 from functools import lru_cache
 import json
@@ -95,6 +95,9 @@ def run_processor(
         instance_caching=instance_caching
     )
 
+    old_cwd = getcwd()
+    chdir(processor.workspace.directory)
+
     ocrd_tool = processor.ocrd_tool
     name = '%s v%s' % (ocrd_tool['executable'], processor.version)
     otherrole = ocrd_tool['steps'][0]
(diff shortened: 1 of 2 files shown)


bertsky (changes requested):
This is problematic:

1. obviously, you forgot the other branch of the conditional (with profiling enabled, there's also a `processor.process()`...)
2. your cached processors were all constructed with a non-null `workspace` argument, which causes them to `chdir` into that:
https://github.com/OCR-D/core/blob/9ae6ddfc8798a07bced98e1d1c110e514c058bfd/ocrd/ocrd/processor/base.py#L140-L142
3. your current `workspace.directory` might not be an absolute path; so it would become relative to the previous workspace, not to CWD of the caller

So either you rebase to the `processor.old_pwd`, or pass `workspace=None` when caching, or resolve to absolute here before calling.

(BTW, in my [PR](https://github.com/OCR-D/core/pull/652) I also moved the chdir to `run_api` alone, which would be the equivalent of the second option I mentioned.)

MehmedGIT:
@bertsky, thanks for the feedback, and insight. Indeed, I missed the first `processor.process()` when I skimmed through the code. It was a bit of rushed PR through the browser on Friday. 

> so it would become relative to the previous workspace, not to CWD of the caller

Oh. Why is this so? This feels a bit odd to what I would have expected - the CWD of the caller once the process() method exits. I will check in more detail how to get the desired behavior without changing the current logic much.

bertsky:
> > so it would become relative to the previous workspace, not to CWD of the caller
> 
> Oh. Why is this so? This feels a bit odd to what I would have expected - the CWD of the caller once the process() method exits. I will check in more detail how to get the desired behavior without changing the current logic much.

Because your `run_processor` does change back to the old CWD, but the initial (constructor's) chdir does **not** get reverted ever after. (So "old CWD" could well become initial workspace at runtime, as opposed to caller's CWD.)

ocrd/ocrd/processor/helpers.py
```diff
@@ -96,7 +94,8 @@ def run_processor(
         output_file_grp=output_file_grp,
         instance_caching=instance_caching
     )
-
+    
+    old_cwd = getcwd()
```
bertsky, line 98: Could you dump a stacktrace of the regression here?

bertsky: See my [analysis here](https://github.com/OCR-D/core/pull/974#discussion_r1134717424).

IMO this should be fixed here first (undoing the revert but compensating in the tests). Subsequently, it could be cherry-picked (or repeated) in #974.

bertsky: @MehmedGIT you still need to compensate in the regression tests, as described.

bertsky (commented):
Looks better now, but I still feel we should not pass `workspace` to `get_processor` at all.

(The processor constructor will chdir into that, so if it is a _relative_ path, then our own subsequent `chdir` on L98 will not work.)

Better to omit the workspace kwarg (so `None` gets passed to the constructor), and then set `processor.workspace = workspace` afterwards (next to the chdir on L98).

bertsky (approved):
LGTM – thanks!

MehmedGIT:
@kba, this should be good to be merged now. I will apply the same changes to #974 as well.

PR #991, opened 2023-03-10
resmgr download: guess type from content

bertsky:
fixes #982

PR #993, opened 2023-03-10
Add resmgr Google Drive downloads

### LINKED ISSUE #992: resmgr download: support Google Drive
Unfortunately, increasingly model providers resort to Google for webspace, which does not allow direct download links. Interactive downloads are nonce-backed and shortlived.

See [example](https://drive.google.com/file/d/1olLQOXvt6nYbvW4pES3T_ABqSlWRq9mY).

According to [this Stackoverflow entry](https://stackoverflow.com/questions/25010369/wget-curl-large-file-from-google-drive), they require users to make a second request based on some previous cookie.

This is very annoying, but luckily there is a Python package [gdown](https://github.com/wkentaro/gdown) which can do the dirtywork.

The problem now becomes: how do we integrate this, so other URLs are not affected, and our `_download_impl`'s progress meter and file output still work?

bertsky:
fixes #992 

But the supported URLs are somewhat limited IIUC. For our [example](https://drive.google.com/file/d/1olLQOXvt6nYbvW4pES3T_ABqSlWRq9mY), we have to append `/view`. Besides this `/file/d/.../view` pattern, gdown knows about `/presentation/d/.../edit` and `/uc?id=...` URLs.

Anyway – this seems least painful. Might have to extend when more cases arise.

### DIFF
--- a/ocrd/ocrd/resource_manager.py
+++ b/ocrd/ocrd/resource_manager.py
@@ -8,6 +8,8 @@ from tarfile import open as open_tarfile
 from urllib.parse import urlparse, unquote
 from zipfile import ZipFile
 
+from mimetypes import guess_type
+from filetype import guess
 import requests
 from yaml import safe_load, safe_dump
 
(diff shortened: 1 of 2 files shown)


bertsky:
Sry, PR is based on #991 already.

bertsky:
> But the supported URLs are somewhat limited IIUC. [...]

Note: in the case of folders, you first need to use `gdown --folder URL` on the CLI to determine the single-file URLs (which the resmgr downloader then supports). [Example](https://drive.google.com/drive/folders/1COTV5f7dEAA4Txmxy3LVfcNHiPSc4Bmp)

kba (approved):

PR #999, opened 2023-03-15
ocrd_cli_wrap_processor: remove fake ocrd_tool param

### LINKED ISSUE #998: ocrd_cli_wrap_processor: drop passing ocrd_tool=None
We seem to have a bug from very early on that never surfaced:

Obviously, nowhere in the CLI params there is a setting for the ocrd_tool.

However, our Processor decorator `ocrd_cli_wrap_processor` does take such a kwarg:

https://github.com/OCR-D/core/blob/5512569316f659f13c3a2d12727e9a7d830089fc/ocrd/ocrd/decorators/__init__.py#L26

It does nothing but pass it on to `run_processor`:

https://github.com/OCR-D/core/blob/5512569316f659f13c3a2d12727e9a7d830089fc/ocrd/ocrd/decorators/__init__.py#L117

In that context, `ocrd_tool` could be meaningful (e.g. when running a processor from API):

https://github.com/OCR-D/core/blob/5512569316f659f13c3a2d12727e9a7d830089fc/ocrd/ocrd/processor/helpers.py#L38

So it gets passed onto the processor constructor:

https://github.com/OCR-D/core/blob/5512569316f659f13c3a2d12727e9a7d830089fc/ocrd/ocrd/processor/helpers.py#L87-L91

Still, this is not usually a problem, because all of our Processor subclasses simply override that kwarg.

But what if they just default to their own value?

IMO the top-level ocrd_tool param is wrong must be removed.

bertsky:
Fixes #998

### DIFF
--- a/ocrd/ocrd/decorators/__init__.py
+++ b/ocrd/ocrd/decorators/__init__.py
@@ -23,7 +23,6 @@ from .mets_find_options import mets_find_options
 
 def ocrd_cli_wrap_processor(
     processorClass,
-    ocrd_tool=None,
     mets=None,
     working_dir=None,
     dump_json=False,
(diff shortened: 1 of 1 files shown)


kba (approved):
Í should really read all the issue first, I just fixed it the same way before I saw this PR :)

PR #1003, opened 2023-03-15
partial revert of b4c98ce266

bertsky:


stweil (approved):
This is a required fix.

kba (approved):

PR #1004, opened 2023-03-15
Dockerfile: Remove unneeded installation of python3-pip

stweil:
Fixes: f1527a2f605f3ea5 ("upgrade base image to 20.04")

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,6 @@ RUN apt-get update && apt-get -y install software-properties-common \
     && apt-get update && apt-get -y install \
         ca-certificates \
         python3-dev \
-        python3-pip \
         python3-venv \
         gcc \
         make \


stweil:
Because it is unneeded as the commit message says?

stweil:
PR (and commit message) updated to fix conflict with preceding PR.

bertsky:
> Because it is unneeded as the commit message says?

It says:
> [Remove unneeded installation of python3-pip](https://github.com/OCR-D/core/pull/1004/commits/c03059ffe5f838e36df0fca97b3fec59ef4f8e33)

That's not an explanation. Why are we investing so much time to remove a tiny package that is usually needed (except with the `python -m pip` workaround)?

> It was (re-)added in commit [f1527a2](https://github.com/OCR-D/core/commit/f1527a2f605f3ea5ec89180b934291418ef8693e) without giving a reason.

IMO the burden for an explanation is still on your side (even if it had been removed earlier).

bertsky:
This **did** create a regression – see [new CI failure](https://app.circleci.com/pipelines/github/OCR-D/core/1686/workflows/1617bf7b-2f32-4a86-b2d2-30acb0b25cde/jobs/6218).

So even if I can see your argument that in Docker we want /usr/local as venv and therefore must avoid conflict with /usr – that now necessitates doing your [python3 -m pip trick](https://github.com/OCR-D/ocrd_all/pull/352#discussion_r1137579831) in **in core as well**.

kba:
😆 OK, that was a long way to to go to get back to where we were before we started messing with `python3-{pip,venv}`. I'll reintroduce it now, so I can end the work day with a working docker image

bertsky:
> So even if I can see your argument that in Docker we want /usr/local as venv and therefore must avoid conflict with /usr – that now necessitates doing your [python3 -m pip trick](https://github.com/OCR-D/ocrd_all/pull/352#discussion_r1137579831) in **in core as well**.

On top of that: in ocrd/core, we don't even manage venvs – everything is native/system-wide. So at least here we **do need** python3-pip.

stweil:
Why does the docker build try `pip3 install --upgrade pip setuptools` outside of a virtual environment?
And why did we switch from pip to pip3 again?

I am afraid both goes in the wrong direction.

kba:
I will investigate it further tomorrow, but no more hotfixes today :)

bertsky:
> Why does the docker build try `pip3 install --upgrade pip setuptools` outside of a virtual environment?

Like I said: in ocrd/core, we have no venv, only a system-wide installation. (Which is arguably better than the pseudo-venv in ocrd_all, but there we also need sub-venvs, so...)

> And why did we switch from `pip` to `pip3` again?

Because that's what `python3-pip` delivers. `pip` is not available (unless someone installs Python 2 or sets up their local venv).

stweil:
> that's what python3-pip delivers. pip is not available

I just tried it in the Docker container, and `pip` is available:
```
root@a3351d49b876:/# ls -l /usr/bin/pip*
-rwxr-xr-x. 1 root root 365 Feb 28 10:41 /usr/bin/pip
-rwxr-xr-x. 1 root root 367 Feb 28 10:41 /usr/bin/pip3
```

bertsky:
> I just tried it in the Docker container, and `pip` is available:

In **which** container?

stweil:
I tried `ocrd/core` and its base container (that one after installing `python3-pip`).

bertsky:
> I tried `ocrd/core` and its base container (that one after installing `python3-pip`).

Still don't know an which state of the Dockerfile. This is becoming tedious. See [original bug](https://github.com/OCR-D/core/pull/1004#issuecomment-1470581076) if you want to help.

stweil:
`pip3` was needed for Ubuntu 18.04 without a virtual environment because its `python3-pip` did not install `/usr/bin/pip`. The reason is simple: old Linux versions reserved `pip` for Python 2.

Ubuntu 20.04 and other current Linux distributions no longer provide Python 2. Its `python3-pip` installs `/usr/bin/pip`.

So with current supported software or when using virtual environments (my preferred way) `pip3` is not needed.

PR #1008, opened 2023-03-15
Dockerfile: Use virtual environment

stweil:
python3-pip can now be removed, and running pip instead of pip3 is sufficient.

This also avoids the installation of python3-setuptools and python3-wheel (requirements of python3-pip).


kba (approved):
Agree in principle but will release it separately tomorrow.

PR #1009, opened 2023-03-15
run_processor / get_processor: do not pass (empty) ocrd_tool

bertsky:
Follow-up to premature #999: there are more places where we pass an empty `ocrd_tool` around that will never be used.

Reasoning (from [ocrd_tesserocr discussion](https://github.com/OCR-D/ocrd_tesserocr/pull/191#issuecomment-1470706375)):
> > That's despite ocrd==2.47.1. Where does the empty ocrd_tool come from???
> 
> Aha! It turns out that my https://github.com/OCR-D/core/pull/999 was premature: we also pass an unnecessary ocrd_tool in other places:
> 
> - run_processor kwarg: https://github.com/OCR-D/core/blob/cff8ddaefec72451795a5e0b4bdb7efc18223d99/ocrd/ocrd/processor/helpers.py#L38
> - run_processor calling get_processor: https://github.com/OCR-D/core/blob/cff8ddaefec72451795a5e0b4bdb7efc18223d99/ocrd/ocrd/processor/helpers.py#L91
> - get_processor kwarg: https://github.com/OCR-D/core/blob/cff8ddaefec72451795a5e0b4bdb7efc18223d99/ocrd/ocrd/processor/helpers.py#L306
> - get_processor calling constructor: https://github.com/OCR-D/core/blob/cff8ddaefec72451795a5e0b4bdb7efc18223d99/ocrd/ocrd/processor/helpers.py#L325
> 
> Now, the big story here is: none of these places is needed (or used) – neither currently, nor for https://github.com/OCR-D/core/pull/974 nor https://github.com/OCR-D/core/pull/884!
 
It's also clear now, why we never used this: you need **runtime information** (from the CLI) for this. Either we are already _on_ the CLI (Processor CLI decorator in all processor modules) or we have to make foreign CLI calls and parse the JSON response from the concrete Processor subclasses (which don't live in our venv, or we don't know the class name of, or are bashlib anyway).

### DIFF
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -35,7 +35,6 @@ def _get_workspace(workspace=None, resolver=None, mets_url=None, working_dir=Non
 
 def run_processor(
         processorClass,
-        ocrd_tool=None,
         mets_url=None,
         resolver=None,
         workspace=None,
@@ -59,7 +58,6 @@ def run_processor(
 
     Instantiate a Python object for :py:attr:`processorClass`, passing:
     - the workspace,
-    - :py:attr:`ocrd_tool`
     - :py:attr:`page_id`
     - :py:attr:`input_file_grp`
     - :py:attr:`output_file_grp`
@@ -88,7 +86,6 @@ def run_processor(
         processor_class=processorClass,
         parameter=parameter,
         workspace=workspace,
-        ocrd_tool=ocrd_tool,
         page_id=page_id,
         input_file_grp=input_file_grp,
         output_file_grp=output_file_grp,
@@ -303,7 +300,6 @@ def get_processor(
         processor_class,
         parameter: dict,
         workspace: Workspace = None,
-        ocrd_tool: dict = None,
         page_id: str = None,
         input_file_grp: List[str] = None,
         output_file_grp: List[str] = None,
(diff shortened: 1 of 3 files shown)


bertsky:
Now also updated the tests to not pass on `ocrd_tool` for no reason.

BTW, we should try to cover cached processor instances in the processor tests, too.

tdoan2010 (approved):

MehmedGIT (approved):

bertsky:
@kba as this modifies existing API args, will necessitate a new minor – 2.48. I have scanned all ocrd_all code and have found no other use cases throwing `ocrd_tool` kwarg at `run_processor` other than in core's own regression tests. It may be that outside repos do this, though.

bertsky:
@kba please prioritise. If you find any arguments against this, I can still try to find a different solution [in ocrd_tesserocr](https://github.com/OCR-D/ocrd_tesserocr/pull/191), but as it stands, I don't see a reason. And since after https://github.com/OCR-D/core/pull/972, ocrd_tesserocr [stopped working](https://github.com/OCR-D/ocrd_tesserocr/issues/190), this is blocking the next ocrd_all release.

joschrew (approved):

joschrew:
My thoughts behind my approval: I see this PR as kind of reminder to not provide `ocrd_tool=None` in
a processor call (or in a run_processor invocation). This must be kept in mind / ensured for the
PR 972 and for other possible changes.

kba (approved):

bertsky:
@kba for the v2.48 changelog:

Changed

 * :fire: `ocrd.run_processor` / `ocrd.processor.get_processor`: rm unnecessary `ocrd_tool` kwarg #998

PR #1010, opened 2023-03-15
Remove all trailing whitespace

stweil:
Use `git log --check` to see the commits which introduced trailing whitespace.


kba:
Not a big fan to be honest, because it updates

* some files that haven't been updated in a long while, making it harder to see what has actually changed recently
* changes generated files and XML schemas we import from elsewhere
* requires changes to the tests because of changed checksums

Besides the aesthetics, is there a functional reason for this?

bertsky:
I concur with @kba. Adding:
* conflicts with outstanding PRs, making it harder to finalize and catch up

stweil:
I rebased and updated this pull request. The bad news is that now 29 files show whitespace issues. Some months ago 26 files were affected.

PR #1014, opened 2023-03-18
avoid forcing opencv and numpy updates

bertsky:
385c595 (Mike's trick to speed up OpenCV and Numpy builds on older Python) should have been formulated without the `--update` here, as it turns out this unfortunately interferes with [numpy version conflicts between TF and H5py](https://github.com/OCR-D/ocrd_all/pull/356#issuecomment-1473642630).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ deps-test:
 install:
 	$(PIP) install -U pip wheel setuptools fastentrypoints
 	@# speedup for end-of-life builds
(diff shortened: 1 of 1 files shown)


bertsky:
> 119d459

Note: I have try to fix the CUDA build here, too, because the CUDA version is used for CI anyway. For the I have not seen problems with missing downwards compatibility yet, but perhaps my testing was not thorough enough (and it's difficult to keep track of all the [open ends at the same time](https://github.com/OCR-D/ocrd_all/pull/356).

kba (approved):
Agreed, I will focus on getting the docker setup fixed and ocrd_all release finished tomorrow.

PR #1028, opened 2023-03-28
Extension to the Processing Server (#974)

MehmedGIT:
This is the required temporal extension for the processing server to download the ocrd all tool json when the server is started. Implemented separately from #974, to not block #974 anymore and continue the required changes here.

### DIFF
--- a/ocrd_network/ocrd_network/processing_server.py
+++ b/ocrd_network/ocrd_network/processing_server.py
@@ -223,8 +223,7 @@ class ProcessingServer(FastAPI):
     async def push_processor_job(self, processor_name: str, data: PYJobInput) -> PYJobOutput:
         """ Queue a processor job
         """
-        if not self.rmq_publisher or not self.rmq_publisher._connection or not self.rmq_publisher._channel:
-            self.log.error('RMQPublisher is not connected')
+        if not self.rmq_publisher:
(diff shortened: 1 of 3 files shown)


bertsky:
@MehmedGIT here it would make sense rebasing the PR on the `dev-processing-broker` branch, too.

tdoan2010:
@MehmedGIT I think the code from this PR was already merged in another PR. So, we can close this, or?

bertsky:
BTW, we still need to add `ocrd_network` to the Makefile's `BUILD_ORDER`, otherwise `make install` will not install it.

kba:
> BTW, we still need to add `ocrd_network` to the Makefile's `BUILD_ORDER`, otherwise `make install` will not install it.

`ocrd_network` is in `BUILD_ORDER` or did I miss something?

https://github.com/OCR-D/core/blob/b7130307ce68c5f074c1ceea24b66ae9ee9ef289/Makefile#L12

bertsky:
> `ocrd_network` is in `BUILD_ORDER` or did I miss something?

Oops, I was using the wrong branch locally and got confused. Sorry for the noise!

MehmedGIT:
This PR is no longer needed since #1030 already implemented the required extension.

Issue #413, opened 2020-01-17
smoke test bashlib-based processor


kba:
@bertsky in #410

> Maybe we should have some pseudo-processor test in ocrd/bashlib/test (beside the arg-parsing test).

We should.

kba:
@bertsky kindly added a bashlib test with an actual processor in https://github.com/OCR-D/core/pull/1031 which has already helped us spot issues like the shebang.

PR #1031, opened 2023-03-29
bashlib: make input-files n-ary

### LINKED ISSUE #1027: bashlib: implement input-files for multiple input fileGrps
Currently, we have
https://github.com/OCR-D/core/blob/de084535e18a2f2fa9293085cda4431abfe2e191/ocrd/ocrd/processor/base.py#L269
as an alias for the mono-valued variant of the generic
https://github.com/OCR-D/core/blob/de084535e18a2f2fa9293085cda4431abfe2e191/ocrd/ocrd/processor/base.py#L294
and we delegate to the former from bashlib via
https://github.com/OCR-D/core/blob/de084535e18a2f2fa9293085cda4431abfe2e191/ocrd/ocrd/cli/bashlib.py#L80-L119
which can then be used in bash processors after
https://github.com/OCR-D/core/blob/de084535e18a2f2fa9293085cda4431abfe2e191/ocrd/ocrd/lib.bash#L241-L256
in the form of `ocrd__input_file num feat` (where `num` is the physical page number and `feat` is `url|ID|mimetype|pageId`).

So for processors that (can) consume multiple input fileGrps at once, like [ocrd-pagetopdf](https://github.com/UB-Mannheim/ocrd_pagetopdf/issues/21), we need to generalize. 

And because we would not want to break existing usage, and bash's data structures are poor (associative arrays are not recursive), I suggest exposing an additional variable+function `ocrd__input_files num feat` which yields a list of results (one per fileGrp).

bertsky:
fixes #1027

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
       PYTHON: python3
     steps:
       - checkout
-      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install imagemagick geos
+      - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install imagemagick geos bash
       - run: make install 
       - run: make deps-test test benchmark
 
--- a/README.md
+++ b/README.md
@@ -140,6 +140,8 @@ For example:
     declare -A NAMESPACES MIMETYPES
     eval NAMESPACES=( `ocrd bashlib constants NAMESPACES` )
     echo ${NAMESPACES[page]}
+    eval MIMETYPE_PAGE=( `ocrd bashlib constants MIMETYPE_PAGE` )
+    echo $MIMETYPE_PAGE
     eval MIMETYPES=( `ocrd bashlib constants EXT_TO_MIME` )
     echo ${MIMETYPES[.jpg]}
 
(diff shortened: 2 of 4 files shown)


bertsky:
Note: usage is as follows:
- the builtin function `ocrd__input_file` is still available (and I kept the name – no plural), but **if there are 2 input fileGrps**, then all its result will be a (single-quote delimited) **list**, e.g. `[url]='file1.xml file2.xml' [ID]='id_file1 id_file2' ...`
- whenever it needs to be called, one should therefore encapsulate its results in a (non-associative) **array variable** and without extra quotes, e.g. `in_file=($(ocrd__input_file 3 url))`
- this will yield the first fileGrp's results on index 0, which in bash will always be the same as if you referenced the array without index (so code does not need to be changed much), e.g. `test -f $in_file`
- if there is a second fileGrp, then its result will have to be fetched from index 1, e.g. `test -f ${in_file[1]}`

For a full example, see https://github.com/UB-Mannheim/ocrd_pagetopdf/pull/22

bertsky:
> For a full example, see [UB-Mannheim/ocrd_pagetopdf#22](https://github.com/UB-Mannheim/ocrd_pagetopdf/pull/22)

Since I will probably [rewrite ocrd-pagetopdf](https://github.com/UB-Mannheim/ocrd_pagetopdf/pull/22#issuecomment-1488376858) to not use multiple input fileGrps anymore, we could use another example for this capability.

How about some generic bashlib processor in tests which merely does a `cp` from input to output? (Or for multiple input fileGrps `cp` the file from the last grp, respectively.)

kba (approved):
That's a good way to ensure backwards-compatibility and support multiple input file groups, even [if they're not necessary for ocrd_pagetopdf anymore](https://github.com/UB-Mannheim/ocrd_pagetopdf/pull/22#issuecomment-1488376858).

kba:
> Note: usage is as follows:

Can you add this as a comment to `ocrd__input_file` pls?

bertsky:
> Can you add this as a comment to `ocrd__input_file` pls?

https://github.com/bertsky/core/commit/2b51a5e0db22636eccf8ad8ba727657436aece3f

bertsky:
sorry for the noise – was too lazy to do it locally

bertsky:
fixes #413

bertsky:
> fixes #413

it's really strange: I cannot link to this issue (in text or directly in `development`). Neither in the opposite direction (from issue to this PR). Another GH quirk?

kba:
> > fixes #413
> 
> it's really strange: I cannot link to this issue (in text or directly in `development`). Neither in the opposite direction (from issue to this PR). Another GH quirk?

Apparently so, I cannot link them either...

bertsky:
Finally got it working – ready to be merged :sweat:

bertsky:
CI: now only macos failure remains.

kba:
I struggle to fix the macos test, now it's complaining that `bash` is too old, I installed bash `5.1.8` via homebrew but that's apparenly not the first `bash` in `$PATH`. When I try to debug it via SSH, it restarts the job but does not offer an SSH session. So I'll have to debug that further in the context of the other failing macos tests.

bertsky:
@kba can we please make a quick release 2.50 out of the current master? Remember that in https://github.com/UB-Mannheim/ocrd_pagetopdf/pull/22 I had to require a new 2.50 for the n-ary bashlib fix.

PR #1041, opened 2023-04-01
Reduce ocrd/core-cuda

bertsky:
see [here](https://github.com/OCR-D/ocrd_all/pull/362#issuecomment-1492885975)


Dockerfile
```diff
@@ -34,12 +34,15 @@ RUN apt-get update && apt-get -y install software-properties-common \
         git \
     && make deps-ubuntu \
     && pip3 install --upgrade pip setuptools \
+    && hash -r \
```
kba, line 37: This rebuilds `$PATH`, right - why is that necessary?

Dockerfile
```diff
@@ -34,12 +34,15 @@ RUN apt-get update && apt-get -y install software-properties-common \
...
     && make install \
-    && apt-get remove -y gcc \
-    && apt-get autoremove -y \
     && $FIXUP \
     && rm -rf /build-ocrd
 
 WORKDIR /data
 
+# remove any entry points from base image
+RUN rm -fr /opt/nvidia
```
kba, line 45: Why not add that command to the previous `RUN` to reduce number of layers?

Makefile
```diff
@@ -216,38 +214,15 @@ pyclean:
 
 # Build docker image
 docker docker-cuda:
-	docker build -t $(DOCKER_TAG) --build-arg BASE_IMAGE=$(DOCKER_BASE_IMAGE) $(DOCKER_ARGS) .
+	docker build --progress=plain -t $(DOCKER_TAG) --build-arg BASE_IMAGE=$(DOCKER_BASE_IMAGE) $(DOCKER_ARGS) .
```
kba, line 217: :+1: will make the build log more readable.

Makefile
```diff
@@ -216,38 +214,15 @@ pyclean:
...
-	@echo /usr/local/cuda-11.0/lib64 >> $@
-	@echo /usr/local/cuda-11.0/targets/x86_64-linux/lib >> $@
-	@echo /usr/local/cuda-11.1/lib64 >> $@
-	@echo /usr/local/cuda-11.1/targets/x86_64-linux/lib >> $@
-	@echo /usr/local/cuda-11.3/lib64 >> $@
-	@echo /usr/local/cuda-11.3/targets/x86_64-linux/lib >> $@
-	@echo /usr/local/cuda-11.7/lib64 >> $@
-	@echo /usr/local/cuda-11.7/targets/x86_64-linux/lib >> $@
-	@echo /usr/local/cuda-12.1/lib64 >> $@
-	@echo /usr/local/cuda-12.1/targets/x86_64-linux/lib >> $@
```
kba, line 250: If it really can be simplified like this: :tada: great.

kba (approved):

Dockerfile
bertsky: Because after the above command we actually have two competing versions installed:
- /usr/bin/pip → /usr/lib/python3/dist-packages/pip (v20)
- /usr/local/bin/pip → /usr/local/lib/python3.8/dist-packages/pip (v23)

And to prefer the newer one, you need to rehash or restart the shell – see [here](https://github.com/NVIDIA/tensorflow/issues/86#issuecomment-1490923653)

Makefile
```diff
@@ -72,8 +72,6 @@ install:
 	@# speedup for end-of-life builds
 	if $(PYTHON) -V | fgrep -e 3.5 -e 3.6; then $(PIP) install --prefer-binary opencv-python-headless numpy; fi
 	for mod in $(BUILD_ORDER);do (cd $$mod ; $(PIP_INSTALL) .);done
-	@# workaround for shapely#1598
-	$(PIP) install --no-binary shapely --force-reinstall shapely
```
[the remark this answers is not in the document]
bertsky: I'm not so sure actually. Remember, this was an [issue](https://github.com/shapely/shapely/issues/1598) of interference between Shapely and Pytorch. We now use a different version on either side. I did not see the problem surface, but should actually check directly. (Unfortunately, this can only be done once I am finished with configuring a GPU-enabled ocrd/all of managable size.)

Makefile
bertsky: with the newer progress meter, there simply is no build log (the output simply gets overwritten by new output along the way)

Makefile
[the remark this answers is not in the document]
bertsky: Unfortunately, the problem does persist with Shapely 2.x. I see no other option but to import torch early / shapely late everywhere...

But because ocrd_validators already imports shapely, and ocrd_cli_wrap_processor imports ocrd_validators, this is quite difficult: we would have to import torch all over the place (even in modules which don't need it)...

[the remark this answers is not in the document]
bertsky: > I see no other option but to import torch early / shapely late everywhere...

Or – of course – to keep compiling shapely via `pip install --no-binary shapely --force-reinstall shapely`.

BTW, currently Kraken also interferes:
```
kraken 4.3.9 requires numpy~=1.23.0, but you have numpy 1.24.2 which is incompatible.
kraken 4.3.9 requires shapely==1.8.5, but you have shapely 2.0.1 which is incompatible.
```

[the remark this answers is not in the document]
bertsky: > And it looks like [pip configuration](https://pip.pypa.io/en/stable/topics/configuration/) does **not** allow setting `no-binary` statically.

Ah, it [is supported now](https://github.com/pypa/pip/commit/530464b15762907d011f6d0fa006995f91c5961b), just not documented. It does work with a simple `echo no-binary = shapely >> /etc/pip.conf`. We just have to be careful if there already is a config / global section / no-binary line.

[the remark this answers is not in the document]
bertsky: I am wondering what to do about this problem in native installations with CUDA available though. We cannot just write to /etc/pip.conf or even ~/.config/pip/pip.conf. But we also cannot just sit around and wait for the next `pip install` to bring us the shapely of death. If only there was a way to confine this constraint to the current venv...

bertsky:
@kba, 12e781c fixes #1050, which I had to include here, because Torch and Tensorflow [compete over cuDNN](https://github.com/OCR-D/ocrd_all/pull/362/commits/fc35815243c3820722600fd83ed8ada73ad2ffc3). So we currently have to break Kraken's explicit dependencies in order to satisfy all TF processors' implicit dependencies. (Kraken/Torch does work with the newer version, but TF would crash on the older one.)

Image size of ocrd/core-cuda is only 3.7 GB again now. And ocrd/all:maximum-cuda-git then [comes out](https://github.com/OCR-D/ocrd_all/pull/362) with 22 GB, which is acceptable.

This review should have priority – next would be https://github.com/OCR-D/ocrd_all/pull/362.

bertsky:
oh, and 47eff22 is needed because kraken just renamed master → main, so our URL for blla.mlmodel resource does not work anymore (and the correct model URL is already part of https://github.com/OCR-D/ocrd_kraken/pull/38).

Dockerfile.cuda
```diff
@@ -0,0 +1,63 @@
...
+        for lib in /usr/local/lib/python3.8/site-packages/nvidia/$pkg/lib/lib*.so.*; do \
+            base=$(basename $lib); \
+            ln -s $lib $CONDA_PREFIX/lib/$base.so; \
+            ln -s $lib $CONDA_PREFIX/lib/${base%.so.*}.so; \
+        done \
+     && ln -s /usr/local/lib/python3.8/site-packages/nvidia/$pkg/include/* $CONDA_PREFIX/include/; \
+    done \
+ && ldconfig
+# gputil/nvidia-smi would be nice, too – but that drags in Python as a conda dependency...
+RUN ldconfig
```
kba, line 57: Redundant, because `ldconfig` happens in the call before?

Dockerfile.cuda
```diff
@@ -0,0 +1,63 @@
...
+                  cuda-cccl \
+ && conda clean -a \
+ && find $CONDA_PREFIX -name "*_static.a" -delete
+                  # cuda-cudart-dev \
+                  # cuda-libraries-dev \
+#RUN conda install -c conda-forge \
+#          cudatoolkit=11.8.0 \
+#          cudnn=8.8.* && \
+#    conda clean -a && \
+#    find $CONDA_PREFIX -name "*_static.a" -delete
```
kba, line 33:
```suggestion
```

Dockerfile.cuda
bertsky: oops, indeed

ocrd/ocrd/constants.py
```diff
@@ -1,7 +1,7 @@
 """
 Constants for ocrd.
 """
-from ocrd_utils.package_resources import resource_filename
+from pkg_resources import resource_filename
```
kba, line 4: What is wrong with `ocrd_utils.package_resources.resource_filename`?

kba: Or rather, should we adopt `ocrd_utils.package_resources` to provide the basic `pkg_resources` functionality, so it is consistent?

kba: OK, reading commit-by-commit would be wise, `ocrd_utils.package_resources` is gone now, which is also a solution.

kba (approved):
LGTM, testing now.

Dockerfile.cuda
bertsky: I thought it would be valuable keeping this recipe (which itself was hard to arrive at) visible in this context. Compare with the pip-based next block, which does something equivalent (but not trivially equal).

ocrd/ocrd/constants.py
bertsky: Like I said above, I have simply reverted the merge of #882 here.

Makefile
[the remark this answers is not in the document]
bertsky: Well, I was surprised to learn that, unfortunately, no: `pip config set prefer-binary` does not actually seem to work. See the CI failures in the changes leading up to b3618a9.

AFAICT the problem is that pip does not have a correct (workable) syntax for a list of kw options (`shapely=1 opencv-python-headless=1`).

PR #1045, opened 2023-04-29
allow plain paths in page_from_file

bertsky:
I repeatedly find myself in need of this.

BTW, there still is a problem with our tests. I keep getting [`I/O operation on closed file`](https://github.com/pytest-dev/pytest/issues/14) in test_logging.py when running in local environments. More explanations [here](https://github.com/pytest-dev/pytest/issues/5743). Also explains our previous problems with [click.CliRunner](https://github.com/pytest-dev/pytest/issues/3344) I believe. I don't know how to actually fix this, but running with `-s` helps as a workaround.

Let's see how the CI reacts.

### DIFF
--- a/ocrd_modelfactory/ocrd_modelfactory/__init__.py
+++ b/ocrd_modelfactory/ocrd_modelfactory/__init__.py
@@ -8,11 +8,9 @@ from pathlib import Path
 from yaml import safe_load, safe_dump
 
 from PIL import Image
-from mimetypes import guess_type
-from filetype import guess
 from lxml import etree as ET
 
-from ocrd_utils import VERSION, MIMETYPE_PAGE, EXT_TO_MIME
+from ocrd_utils import VERSION, MIMETYPE_PAGE, guess_media_type
 from ocrd_models import OcrdExif, OcrdFile
 from ocrd_models.ocrd_page import (
     PcGtsType, PageType, MetadataType,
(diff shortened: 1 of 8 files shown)


bertsky:
CI test is only macOS (still broken).

BTW, is there a reason for not exporting `OcrdPage` in `ocrd_models` directly?

kba:
> CI test is only macOS (still broken).

Good  :+1:

> BTW, is there a reason for not exporting `OcrdPage` in `ocrd_models` directly?

No, just an oversight AFAICT.

ocrd_modelfactory/ocrd_modelfactory/__init__.py
```diff
@@ -79,15 +82,32 @@ def page_from_image(input_file, with_tree=False):
...
             and reverse mapping, too (cf. :py:func:`ocrd_models.ocrd_page.parseEtree`)
     """
+    if not isinstance(input_file, OcrdFile):
+        mimetype = guess(input_file)
+        if mimetype is None:
+            mimetype = guess_type(input_file)[0]
+        else:
+            mimetype = mimetype.mime
+        if mimetype is None:
+            mimetype = EXT_TO_MIME.get(Path(input_file).suffix, None)
```
kba, line 103: Perhaps we should move this to a utility function, also to document why we need the three mechanisms (`mimetype.guess`, `filetype.guess_type` and `EXT_TO_MIME`).

kba (approved):
Makes sense to allow passing string paths.

ocrd_modelfactory/ocrd_modelfactory/__init__.py
bertsky: Sure. But note that there's already a difference to what we need (regarding these detectors) in resmgr.download.

kba: Yes, but these differences (falling back to `application/octet-stream` vs. raising an error) can be handled with function args.

kba: https://github.com/OCR-D/core/pull/1045/commits/ed19b654d70eb9437ea1d46ce065c7471f08787e

PR #1051, opened 2023-06-02
Improve packaging

bertsky:
Since [python setup.py sdist bdist_wheel is deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html), this uses `build` instead. Also, it turns out, we forgot to add our `requirements.txt` to the manifest (and `ocrd.processor.builtin.dummy` as a package) for correct packaging. Further improves the makefile.

Note: in the long run we should
- switch to pyproject.toml, with setuptools as build backend, and setup.cfg
- move requirements.txt to the latter's `install_requires` key, and either abandon `requirements.txt` or keep it for dev purposes (with a workable state of `pip freeze`)
- tie versioning to git tags (so there will be no need for manual synchronization anymore, and not interdependencies at build time) via [setuptools_scm recipe](https://www.moritzkoerber.com/posts/versioning-with-setuptools_scm/) – see [this example](https://github.com/slub/textract2page/blob/master/pyproject.toml)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,8 @@ env2/
 
 # vim swp
 *.swp
+# emacs bkup
+*~
 
 ocrd.egg-info
 .pytest_cache
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,8 @@ SPHINX_APIDOC =
 
 BUILD_ORDER = ocrd_utils ocrd_models ocrd_modelfactory ocrd_validators ocrd_network ocrd
 
-FIND_VERSION = grep version= ocrd_utils/setup.py|grep -Po "([0-9ab]+\.?)+"
+PEP_440_PATTERN := '([1-9][0-9]*!)?(0|[1-9][0-9]*)(\.(0|[1-9][0-9]*))*((a|b|rc)(0|[1-9][0-9]*))?(\.post(0|[1-9][0-9]*))?(\.dev(0|[1-9][0-9]*))?'
+OCRD_VERSION != fgrep version= ocrd_utils/setup.py | grep -Po $(PEP_440_PATTERN)
 
 # BEGIN-EVAL makefile-parser --make-help Makefile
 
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -104,6 +104,8 @@ env2/
 
 # vim swp
 *.swp
+# emacs bkup
+*~
```
stweil, line 108: Personally I'd add such ignore patterns to my global settings and restrict the .gitignore of the project to patterns which are specific for the project. But I know that there are different opinions.

Makefile
```diff
@@ -10,8 +10,10 @@ TESTDIR = tests
 SPHINX_APIDOC =
```
stweil, line 10: trailing whitespace

stweil, line 10:
```suggestion
SPHINX_APIDOC =
```

.gitignore
kba: Sure, that would be better practice, but if the project-specific `.gitignore` prevents any new contributor from accidentally comitting unwanted files, I think it's fine to add at least vim/emacs in addition to the other python-specific patterns.

Makefile
```diff
@@ -7,11 +7,13 @@ LOG_LEVEL = INFO
 PYTHONIOENCODING=utf8
 TESTDIR = tests
 
-SPHINX_APIDOC = 
+SPHINX_APIDOC =
 
 BUILD_ORDER = ocrd_utils ocrd_models ocrd_modelfactory ocrd_validators ocrd_network ocrd
+reverse = $(if $(wordlist 2,2,$(1)),$(call reverse,$(wordlist 2,$(words $(1)),$(1))) $(firstword $(1)),$(1))
```
kba, line 13: I'm reminded of "intro to functional programming" 20 years ago 👍

kba (approved):
LGTM!

Makefile
bertsky: just plain old recursion – not even a difflist!

I later realised this is not even necessary, since `pip uninstall` seems to be "all at once" (in contrast to `pip install`).

stweil (approved):
Thanks.

PR #1060, opened 2023-06-22
Use grep with extended instead of Perl regexp

stweil:
The command works with both variants.
-E is also supported on macOS while -P is not.


stweil:
Maybe getting the version could also use `git describe --tags` to get intermediate versions which are still not tagged, too.

bertsky (approved):
agreed, no need for PCRE here.

regarding tag based versioning: see https://github.com/OCR-D/core/pull/1051

kba (approved):

PR #1065, opened 2023-06-28
Git versioning and modern packaging

bertsky:
This brings our packaging up to speed, including our surprisingly challenging case of mutually pinned "sub-distributions". I had to comment out the actual requirement ocrd → ocrd_network to break a dependency circle, which is nececssary because otherwise pip goes completely crazy and downloads some arbitrary other versions and then complains of broken requirements. I also improved the makefile.


bertsky:
[CI failure](https://app.circleci.com/pipelines/github/OCR-D/core/1882/workflows/7a95929b-5226-4944-98ef-1b9ebc836d2e/jobs/7425) is macos only – no idea what's going on there.

bertsky:
@kba I resolved the conflict by keeping an empty `setup()`  script. This conflict will re-emerge whenever we release a new version on master. (One of the benefits after merging this PR is that there will only be a single place where the version is specified – in the repo.)

Makefile
```diff
@@ -115,16 +115,24 @@ deps-test:
...
+	$(foreach MODULE,$(BUILD_ORDER),$(PYTHON) -m build ./$(MODULE) &&) echo done
+# or use -n ?
+
 # (Re)install the tool
-install:
-#	$(PIP_INSTALL) $(BUILD_ORDER:%=./%/dist/ocrd$*(OCRD_VERSION)*.whl)
+install: #build
+#	$(PIP_INSTALL) $(BUILD_ORDER:%=./%/dist/ocrd*`$(PYTHON) -m setuptools_scm 2>/dev/null`*.whl)
 	$(foreach MODULE,$(BUILD_ORDER),$(PIP_INSTALL) ./$(MODULE) &&) echo done
 	@# workaround for shapely#1598
```
kba, line 129: Is there anything I can do to help resolve https://github.com/shapely/shapely/issues/1598?

ocrd_utils/.keras/keras.json
```diff
@@ -1,6 +0,0 @@
-{
```
kba, line 1: Oops, never realized I comitted that :o

kba (changes requested):
Thanks for preparing this, I've now finally tested it, esp. love the version-from-git-tag mechanic.

Everything works fine except `make install-dev`. If I try `pip install -e` in ocrd_models, I get an error like 

```
setuptools.extern.packaging._tokenizer.ParserSyntaxError: Expected end or semicolon (after version specifier)
    ocrd_utils == 2.59.2.dev7+g9fbddff75.d20231214 == 2.59.2.dev7+g9fbddff75.d20231214                       
```

I think the issue is that finalize_options is called twice, I don't know why but this simple workaround fixes it for me:

```diff
-self.distribution.install_requires = [                                
-    req + vers if req.startswith('ocrd') else req 
-    for req in self.distribution.install_requires                     
-]                                                                     
+self.distribution.install_requires = [                                
+    req + vers if req.startswith('ocrd') and '==' not in req else req 
+    for req in self.distribution.install_requires                     
+]                                                                     

If that's okay with you, I'll commit that and we can merge this before the next release.

BTW: Do you think there's a way to have a single place to define the `build` class and import it in the `setup.py`s? Not a big issue since I don't expect us to have to change `setup.py` anymore in the future, but it would be neat.

Makefile
bertsky: AFAIK the issue persists. (You can try `make -C ocrd_all test-workflow` in a GPU environment to see the crash, but make sure you comment out the workaround when installing/building everything.)

bertsky:
> If I try `pip install -e` in ocrd_models, I get an error like [...]

Very strange. I don't recall seeing that – perhaps it's one of the recent setuptools changes...

> I think the issue is that finalize_options is called twice, I don't know why but this simple workaround fixes it for me: [...]
> If that's okay with you, I'll commit that and we can merge this before the next release.

Sure, sounds reasonable – please go ahead.

> BTW: Do you think there's a way to have a single place to define the `build` class and import it in the `setup.py`s? Not a big issue since I don't expect us to have to change `setup.py` anymore in the future, but it would be neat.

I did think about inheriting this, but gave up quickly IIRC.

Makefile
kba: OK, nothing much we can do and as long as the workaround is as simple as that, let's focus on more pressing issues.

kba:
OK, added the fix, now everything seems in order. I'll merge and release tomorrow.

kba:
macos still fails but I'll debug that separately.

Issue #450, opened 2020-02-28
workspace validator: suppress notices about download


bertsky:
When I run workspace validation these days, I get a lot of these in the final report:

```
<notice>_validate_dimension: Not executed because --download wasn't set and PAGE might reference remote (Alternatve)Images <...></notice>
```

IMHO these do not belong in the report at all. If anything, they should be output as warning-level messages in the log during validation. 

But to me this is not even justified in the first place: `ocrd_validators.workspace_validator.WorkspaceValidator._validate_dimension` never actually checks whether the PAGE file exists in the filesystem. 

And the problem definitely does _not_ extend to `AlternativeImage`. (So at least that part of the notice is unwarranted.)

PR #1066, opened 2023-06-29
workspace_validator: allow skipping pcgtsid check

bertsky:
This is necessary as some of our processors do not use `set_pcGtsId(file.ID)`, IINM currently:
- [X] all ocrd_segment processors
- [x] ~~ocrd-keraslm-rate~~ (was already correct)
- [ ] ocrd-repair-inconsistencies
- [X] ocrd-detectron2-segment
- [ ] ~~ocrd-fileformat-transform~~ (hard to enforce in submodule transform, only for PAGE output)
- [x] ~~ocrd-im6convert~~ (does not apply: only image output)
- [x] ocrd-page-transform

### DIFF
--- a/ocrd_validators/ocrd_validators/workspace_validator.py
+++ b/ocrd_validators/ocrd_validators/workspace_validator.py
@@ -20,7 +20,7 @@ from .xsd_mets_validator import XsdMetsValidator
 
 class WorkspaceValidator():
     """
-    Validates an OCR-D/METS workspace against the specs.
+    Validator for `OcrdMets <../ocrd_models/ocrd_models.ocrd_mets.html>`.
     """
 
     @staticmethod
@@ -72,8 +72,14 @@ class WorkspaceValidator():
             src_dir (string):
             skip (list):
             download (boolean):
-            page_strictness ("strict"|"lax"|"fix"|"off"):
-            page_coordinate_consistency ("poly"|"baseline"|"both"|"off"):
+            page_strictness ("strict"|"lax"|"fix"|"off"): how strict to check
+                multi-level TextEquiv consistency of PAGE XML files
+            page_coordinate_consistency ("poly"|"baseline"|"both"|"off"): check whether each
+                segment's coords are fully contained within its parent's:
+                 * `"poly"`: *Region/TextLine/Word/Glyph in Border/*Region/TextLine/Word
+                 * `"baseline"`: Baseline in TextLine
+                 * `"both"`: both `poly` and `baseline` checks
+                 * `"off"`: no coordinate checks
         """
         self.report = ValidationReport()
         self.skip = skip if skip else []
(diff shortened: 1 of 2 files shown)


kba (approved):
Makes sense to have this check optional since we cannot always enforce it (though we should, as you have started, ensure that OCR-D processors handle this correctly).

Minor note: Perhaps just `pcgtsid` instead of `mets_fileid_page_pcgtsid` which is easier to type?

bertsky:
> Minor note: Perhaps just `pcgtsid` instead of `mets_fileid_page_pcgtsid` which is easier to type?

It's not clear what is meant by just `pcgtsid`, and it's also not obvious that we are trying to enforce  `mets:file/@ID=pc:PcGts/@pcGtsId`. So I still think the longer identifier is warranted. (We could try to improve the filtering options and docstrings though...)

bertsky:
@kba in 6991166 I also changed the behaviour of `download=True`: IMO it was not correct for the workspace validator to convert remote refs to local – instead, that option now does what all our processors do, i.e. downloading on demand. Also, I fixed #450 (which had been closed earlier by accident): not downloaded files must _not_ add to the report, these belong to the logger.

bertsky:
But actually IMHO the workspace validator should be rewritten even more: It does not make sense to iterate over the same workspace for each PAGE XML test independently. Instead, `imagefilename`, `pixel_density`, `dimension`, `multipage` and `page` should enter as a `page_checks` post-parsing step (as I have started with `mets_fileid_page_pcgtsid`) when the PAGE XML has been downloaded and opened.

bertsky:
> But actually IMHO the workspace validator should be rewritten even more: It does not make sense to iterate over the same workspace for each PAGE XML test independently. Instead, `imagefilename`, `dimension` and `page` should enter as a `page_checks` post-parsing step (as I have started with `mets_fileid_page_pcgtsid`) when the PAGE XML has been downloaded and opened.

see #1070 – to be discussed

ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -171,7 +186,7 @@ def _validate_imagefilename(self):
         self.log.debug('_validate_imagefilename')
         for f in self.mets.find_files(mimetype=MIMETYPE_PAGE):
             if not is_local_filename(f.url) and not self.download:
-                self.report.add_notice("Won't download remote PAGE XML <%s>" % f.url)
+                self.log.warning("Won't download remote PAGE XML '%s'", f.url)
```
kba, line 189: Yes, that is probably better logged than written to the validation report.

kba (approved):

kba:
OK, let's merge this now and continue the discussion on improving the validator (which has not been substantially updated in too long a time) in #1070 .

ocrd_validators/ocrd_validators/workspace_validator.py
bertsky: Like I already said in https://github.com/OCR-D/core/issues/450!

kba:
> Since AFAIK one cannot re-target PR

Oh, I did not realize, sry about that.

PR #1071, opened 2023-07-10
Workspace validator page checks loop

bertsky:
To be discussed, but the current code is inefficient and hard to reuse.

Since the existing functions are all private anyway, perhaps we could be more bold here and encapsulate each single PAGE check with a public member in PageValidator?


ocrd_validators/ocrd_validators/workspace_validator.py
```diff
@@ -146,16 +147,10 @@ def _validate(self):
...
-                    self._validate_dimension()
-                if 'imagefilename' not in self.skip:
-                    self._validate_imagefilename()
-                if 'page' not in self.skip:
-                    self._validate_page()
-                if 'page_xsd' not in self.skip:
-                    self._validate_page_xsd()
                 if 'mets_xsd' not in self.skip:
                     self._validate_mets_xsd()
+                if len(self.page_checks):
```
kba, line 152:
```suggestion
                if self.page_checks:
```

Empty array evaluates to `False`.

bertsky: It does, but since a few years, linters will complain that this is not reliable or something. Is there something to that, or can we ignore+skip such recommendations?

kba: I think at least pylint will actually warn about `len(list)` in boolean contexts.

Maybe you mean the lxml warnings that `if element` is not the same as `if len(element)` to check for no child elements?

PEP8 says


> For sequences, (strings, lists, tuples), use the fact that empty sequences are  false
```python
# Correct:
if not seq:
if seq:
# Wrong:
if len(seq):
if not len(seq):
```

bertsky: Indeed, it does warn about my construct. IIRC it complained the opposite in the past, but I'll happily use the simpler variant from now on.

kba:
I know this is a draft but it is a useful improvement even without further changes to the API - can we merge?

bertsky:
> I know this is a draft but it is a useful improvement even without further changes to the API - can we merge?

Sure. Will try to be more bold next time.

PR #1073, opened 2023-07-12
download: add retry and timeout kwargs, controlled via env vars

### LINKED ISSUE #973: workspace download: add retry option
> Currently, the `find` command of the `Workspace` has a `--download` and `--wait` options. Do you see any potential issues if we provide an extra option that could retry downloading a file if the initial try fails? Something like:
> 
> ```python
> @click.option('--retry-times', type=int, default=0, help="Try this many times if a download request fails")
> ```
> 
> Some METS files inside VD18 repository of SUB contain host links that sometimes fail randomly if many files are retrieved back to back. No matter how much waiting time I put between the requests it would still randomly fail on some file. I know this is a problem on the SUB side, but it's unknown to me how many METS files do they have in that state. So the suggested change above, should at least, give a bit more flexibility. Trying once or twice more (which seems to solve the problem) should also take less time than waiting few seconds between the requests.
> 
> Since the downloading happens in a loop if a file is missing and the download fails, e.g., with `Exception: HTTP request failed: URL (HTTP 500)` - the process is terminated with failure. I have tracked the execution path. There is a method `Workspace.download_file` which internally calls `Resolver.download_to_directory` which triggers that exception if the response status code is not `200`. But since I am confused with the argument `_recursion_count` of `Worksapce.download_file` I did not spent more time to understand the logic.

_originally [posted](https://github.com/OCR-D/core/pull/966#issuecomment-1398341131) by @MehmedGIT_

bertsky:
Fixes #973 (retry facility) and also implements a timeout facility, and provides those to any download use-case besides CLI `ocrd workspace find --download` (e.g. processor's typical `Workspace.download_file(input_file)`, or `WorkspaceValidator(... download=True ...)`.

Since there are so many places which could ultimately trigger `Resolver.download_to_directory`, and we don't want to drag along the new kwargs `retries` and `timeout` everywhere, I chose to control them via **environment variables**:
- `OCRD_DOWNLOAD_RETRIES` → number of attempts
- `OCRD_DOWNLOAD_TIMEOUT` → a single float or a comma-separated tuple of floats (connection timeout, read timeout)

The retry facility itself only triggers on certain HTTP statuses (i.e. opt-in), which is debatable (e.g. I have seen proposals opting out of everything but 200 and 404).

### DIFF
--- a/ocrd/ocrd/resolver.py
+++ b/ocrd/ocrd/resolver.py
@@ -1,6 +1,7 @@
 from tempfile import mkdtemp
 from pathlib import Path
 from warnings import warn
+from os import environ
 
 import requests
 
@@ -23,7 +24,7 @@ class Resolver():
     Handle uploads, downloads, repository access, and manage temporary directories
     """
 
-    def download_to_directory(self, directory, url, basename=None, if_exists='skip', subdir=None):
+    def download_to_directory(self, directory, url, basename=None, if_exists='skip', subdir=None, retries=None, timeout=None):
         """
         Download a file to a directory.
 
@@ -36,11 +37,15 @@ class Resolver():
 
         Args:
             directory (string): Directory to download files to
-            basename (string, None): basename part of the filename on disk.
             url (string): URL to download from
-            if_exists (string, "skip"): What to do if target file already exists. \
+
+        Keyword Args:
+            basename (string, None): basename part of the filename on disk.
+            if_exists (string, "skip"): What to do if target file already exists.
                 One of ``skip`` (default), ``overwrite`` or ``raise``
             subdir (string, None): Subdirectory to create within the directory. Think ``mets:fileGrp``.
+            retries (int, None): Number of retries to attempt on network failure.
+            timeout (tuple, None): Timeout in seconds for establishing a connection and reading next chunk of data.
 
         Returns:
             Local filename string, *relative* to directory
@@ -100,7 +105,46 @@ class Resolver():
             dst_path.write_bytes(src_path.read_bytes())
         else:
             log.debug("Downloading URL '%s' to '%s'", url, dst_path)
-            response = requests.get(url)
+            if 'OCRD_DOWNLOAD_RETRIES' in environ:
+                retries = retries or int(environ['OCRD_DOWNLOAD_RETRIES'])
+            retries = retries or 0
+            for _ in range(retries + 1):
+                try:
+                    if timeout is None and 'OCRD_DOWNLOAD_TIMEOUT' in environ:
+                        timeout = environ['OCRD_DOWNLOAD_TIMEOUT'].split(',')
+                        if len(timeout) > 1:
+                            timeout = tuple(float(x) for x in timeout)
+                        else:
+                            timeout = float(timeout[0])
+                    response = requests.get(url, timeout=timeout)
+                    if response.status_code in [
+                            # probably too wide:
+                            408, # Request Timeout
+                            409, # Conflict
+                            410, # Gone
+                            412, # Precondition Failed
+                            417, # Expectation Failed
+                            423, # Locked
+                            424, # Fail
+                            425, # Too Early
+                            426, # Upgrade Required
+                            428, # Precondition Required
+                            429, # Too Many Requests
+                            440, # Login Timeout
+                            500, # Internal Server Error
+                            503, # Service Unavailable
+                            504, # Gateway Timeout
+                            509, # Bandwidth Limit Exceeded
+                            529, # Site Overloaded
+                            598, # Proxy Read Timeout
+                            599, # Proxy Connect Timeout
+                    ]:
+                        continue
+                except (requests.Timeout, requests.ConnectionError) as error:
+                    response = error
+                    continue
+            if isinstance(response, Exception):
+                raise Exception("HTTP request failed: %s (%s)" % (url, response))
             if response.status_code != 200:
                 raise Exception("HTTP request failed: %s (HTTP %d)" % (url, response.status_code))
             contents = handle_oai_response(response)
--- a/ocrd/ocrd/resolver.py
+++ b/ocrd/ocrd/resolver.py
@@ -4,6 +4,8 @@ from warnings import warn
 from os import environ
 
 import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 
 from ocrd.constants import TMP_PREFIX
 from ocrd_utils import (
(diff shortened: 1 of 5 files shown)


MehmedGIT (approved):
Covering client-side 4XX and server-side 5XX response codes should be enough.

ocrd/ocrd/resolver.py
```diff
@@ -100,7 +105,46 @@ def download_to_directory(self, directory, url, basename=None, if_exists='skip',
...
+                            417, # Expectation Failed
+                            423, # Locked
+                            424, # Fail
+                            425, # Too Early
+                            426, # Upgrade Required
+                            428, # Precondition Required
+                            429, # Too Many Requests
+                            440, # Login Timeout
+                            500, # Internal Server Error
+                            503, # Service Unavailable
```
kba, line 135: Perhaps also `502 Bad Gateway` in case there is a connection issue between outward-facing and internal web service?

@MehmedGIT What HTTP status is returned when  retrieval of @subugoe images fails?

kba (approved):
LGTM, also I think environment variables are the right tool here. We need to document those, though. Perhaps we should add a section for that to the main `ocrd --help` output? And to the README.

ocrd/ocrd/resolver.py
MehmedGIT: I think everything from 4XX and 5XX should be included, except the 404.

@kba It's already covered there - 500.

kba:
There's an issue with mocking `request.get` in the tests not accepting `timeout`. I'm investigating.

ocrd/ocrd/resolver.py
bertsky: I don't think it's correct to include all error cases. For a retry mechanism we should focus on transient failures, not on permanent ones.

bertsky:
> There's an issue with mocking `request.get` in the tests not accepting `timeout`. I'm investigating.

I have already repaired the tests and uploaded the fixes to my feature branch, but somehow the PR does not update...

bertsky:
> I have already repaired the tests and uploaded the fixes to my feature branch, but somehow the PR does not update...

duh, sorry – turns out I had pushed to upstream instead of bertsky. I'll remove the stale branch from upstream now.

ocrd/ocrd/resolver.py
kba: 404 and 410 are likely permanent, so should not be retried. But I would include all server side errors.

ocrd/ocrd/resolver.py
```diff
@@ -100,7 +105,46 @@ def download_to_directory(self, directory, url, basename=None, if_exists='skip',
...
+                        if len(timeout) > 1:
+                            timeout = tuple(float(x) for x in timeout)
+                        else:
+                            timeout = float(timeout[0])
+                    response = requests.get(url, timeout=timeout)
+                    if response.status_code in [
+                            # probably too wide:
+                            408, # Request Timeout
+                            409, # Conflict
+                            410, # Gone
```
kba, line 124: If not 404 then also not 410.

bertsky: Indeed – agreed!
```suggestion
```

bertsky: I'll just wait for the CI before committing.

ocrd/ocrd/resolver.py
```diff
@@ -100,7 +105,46 @@ def download_to_directory(self, directory, url, basename=None, if_exists='skip',
...
+                            412, # Precondition Failed
+                            417, # Expectation Failed
+                            423, # Locked
+                            424, # Fail
+                            425, # Too Early
+                            426, # Upgrade Required
+                            428, # Precondition Required
+                            429, # Too Many Requests
```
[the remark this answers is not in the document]
bertsky: Indeed, `requests.packages.urllib3.util.retry.Retry` is [more powerful](https://scrapeops.io/python-web-scraping-playbook/python-requests-retry-failed-requests/) than a caller-side loop. It has a backoff behaviour with logarithmically increasing timeouts and a customisable list of HTTP codes.

Do you want me to switch to that paradigm?

kba:
> > There's an issue with mocking `request.get` in the tests not accepting `timeout`. I'm investigating.
> 
> I have already repaired the tests and uploaded the fixes to my feature branch, but somehow the PR does not update...

I still needed to adapt the OAI request test:

```diff
-Subproject commit 506b33936d89080a683fa8a26837f2a23b23e5e2
+Subproject commit 1194310c18d90b280c380bdc3cb04adb6a41120f-dirty
diff --git a/tests/test_resolver_oai.py b/tests/test_resolver_oai.py
index ca5c590d8..c0ecf64aa 100644
--- a/tests/test_resolver_oai.py
+++ b/tests/test_resolver_oai.py
@@ -72,7 +72,7 @@ def test_handle_common_oai_response(mock_get, response_dir, oai_response_content                                                                          
     result = resolver.download_to_directory(response_dir, url)

     # assert
-    mock_get.assert_called_once_with(url, timeout=None)
+    mock_get.assert_called_once_with(url)
     assert result == 'oai'


@@ -100,7 +100,7 @@ def test_handle_response_for_invalid_content(mock_get, response_dir):                                                                                   
     resolver.download_to_directory(response_dir, url)

     # assert
-    mock_get.assert_called_once_with(url, timeout=None)
+    mock_get.assert_called_once_with(url)
     log_output = capt.getvalue()
     assert 'WARNING ocrd_models.utils.handle_oai_response' in log_output

```

Then tests pass.

bertsky:
> I still needed to adapt the OAI request test:

AFAICT these are necessary – see CI failures above and current success status.

bertsky:
> LGTM, also I think environment variables are the right tool here. We need to document those, though. Perhaps we should add a section for that to the main `ocrd --help` output? And to the README.

Done:
```
Usage: ocrd [OPTIONS] COMMAND [ARGS]...

  Entry-point of multi-purpose CLI for OCR-D

Options:
  --version                       Show the version and exit.
  -l, --log-level [OFF|ERROR|WARN|INFO|DEBUG|TRACE]
                                  Log level
  --help                          Show this message and exit.

Commands:
  bashlib    Work with bash library
  log        Logging
  network    Managing network components
  ocrd-tool  Work with ocrd-tool.json JSON_FILE
  process    Process a series of tasks
  resmgr     Managing processor resources
  validate   All the validation in one CLI
  workspace  Working with workspace
  zip        Bag/Spill/Validate OCRD-ZIP bags

  Variables:
    PATH    Search path for processor executables
            (affects `ocrd process` and `ocrd resmgr`)
    HOME    Directory to look for `ocrd_logging.conf`,
            fallback for unset XDG variables.

    XDG_CONFIG_HOME
      Directory to look for `./ocrd/resources.yml`
      (i.e. `ocrd resmgr` user database) - defaults to
      `$HOME/.config`.
    XDG_DATA_HOME
      Directory to look for `./ocrd-resources/*`
      (i.e. `ocrd resmgr` data location) - defaults to
      `$HOME/.local/share`.

    OCRD_DOWNLOAD_RETRIES
      Number of times to retry failed attempts
      for downloads of workspace files.
    OCRD_DOWNLOAD_TIMEOUT
      Timeout in seconds for connecting or reading
      (comma-separated) when downloading.

    OCRD_METS_CACHING
      Whether to enable in-memory storage of OcrdMets
      data structures for speedup during processing or
      workspace operations.

    OCRD_MAX_PROCESSOR_CACHE
      Maximum number of processor instances
      (for each set of parameters) to be kept
      in memory (including loaded models) for
      processing workers or processor servers.

    OCRD_NETWORK_SERVER_ADDR_PROCESSING
      Default address of Processing Server to connect to
      (for `ocrd network client processing`).
    OCRD_NETWORK_SERVER_ADDR_WORKFLOW
      Default address of Workflow Server to connect to
      (for `ocrd network client workflow`).
    OCRD_NETWORK_SERVER_ADDR_WORKSPACE
      Default address of Workspace Server to connect to
      (for `ocrd network client workspace`).

    OCRD_PROFILE
      Whether to enable gathering runtime statistics
      on the `ocrd.profile` logger (comma-separated):
      - `CPU`: yields CPU and wall-time,
      - `RSS`: also yields peak memory (resident set size)
      - `PSS`: also yields peak memory (proportional set size)
    OCRD_PROFILE_FILE
      When profiling is enabled, data file to write to
      (for external analysis tools like snakeviz).

```

PR #1075, opened 2023-08-14
ocrd network extension

joschrew:
This are mainly only small additions to ocrd_network which I need for the slim containers.
Summary:
    - Possibility to create a queue for a worker when the worker is externally managed. Previously only the processing-server was able to create a queue for a worker. 
    - Retry for queue creation: When a worker is started at the same time as the rabbit-mq, the rabbitmq needs up to 2 seconds to be fully available so that a queue can be created.
    - allow starting a processing-server without any workers or processor-servers
    - limit pydantic to version 1. Version 2 was released in June 2023 and conflicts with current models which were created for pydantic v1

### DIFF
--- a/ocrd_network/ocrd_network/processing_worker.py
+++ b/ocrd_network/ocrd_network/processing_worker.py
@@ -18,6 +18,8 @@ import pika.adapters.blocking_connection
 
 from ocrd_utils import getLogger
 
+from time import sleep
+
 from .database import (
     sync_initiate_database,
     sync_db_get_workspace,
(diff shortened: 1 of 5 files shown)


ocrd_network/ocrd_network/processing_worker.py
```diff
@@ -269,3 +272,24 @@ def post_to_callback_url(self, callback_url: str, result_message: OcrdResultMess
...
+            attempts_left = connection_attempts if connection_attempts > 0 else 1
+            while attempts_left > 0:
+                try:
+                    self.connect_publisher()
+                    break
+                except AMQPConnectionError as e:
+                    if attempts_left <= 1:
+                        raise e
+                    attempts_left -= 1
+                    sleep(retry_delay)
```
kba, line 292: If this is a systematic problem, why not wait before the first connection?

kba (approved):

ocrd_network/ocrd_network/processing_worker.py
joschrew: Already merged but anyway: This is user controlled waiting. User specifies if and if so how long he/she wants to wait. There are cases when the queue is already running and no waiting is necessary when starting a worker. I have the case that queue and worker are started at the same time, this takes about 2 Secs and waiting is necessary.

kba: Yeah, good point, I thought as much, so it was more of an understanding question than a change request. Thanks for the explanation!

MehmedGIT:
This was already merged when I was on vacation, however, I thought we wanted to have an endpoint on the Processing Server for registering external (not deployed at start time) processing workers and processor servers. This way the additional `--create-queue` and `--queue-connect-attempts` flags for the processing worker are not needed.

MehmedGIT:
It was decided to remove the flags. The first flag `--create-queue` is not needed since queue creation is idempotent. The processing worker will always try to create the queue when started. The second flag could be just an environment variable.

Issue #323, opened 2019-09-24
mets:file URL handling: keep remote links


bertsky:
Currently with workspaces we can either keep images on the remote side by using http URLs in `mets:file/mets:FLocat/@xlink:href` (which means they have to be downloaded again and again during processing), or get local filesystem copies with relative paths by cloning with `download=True` or bagging and spilling (but then the source information will be lost forever).

When processing is finished and I want to make my workspace public, I now have to upload my shiny new results _in addition_ to the original images – which I might not even have the rights to publish myself. It would be much better, if the original remote URLs would be used again for that – _even if I used local copies in between_.

METS-XML allows that: A `mets:FLocat` has `xs:@maxoccurs=unbounded` within `mets:file`, with the following documented semantic:
> The file element provides access to content files for a METS object. A file element may contain one or more FLocat elements, which provide pointers to a content file, and/or an FContent element, which wraps an encoded version of the file. Note that ALL FLocat and FContent elements underneath a single file element should identify/contain identical copies of a single file.

So why don't we keep 2 `FLocat` elements in that case, one relative path for local processing and one remote URL for provenance/bookkeeping? When making results public, the local copies could be disposed of again, e.g. when bagging with `--manifestation-depth=partial`.

kba:
That's a great proposal and would also be an option to keep `@imageFilename` and `@xlink:href` in check.

The idea behind the `local_filename` property was much the same: To have a local copy of a potentially-remote URL. Another idea was to use multiple `FLocat` as you propose but there were reasons why we decided against it. @maria-federbusch @cneud @tboenig I cannot seem to find the discussion in the issues in core or spec, do you remember where we documented this? IIRC (and I might not), there was a limitation in Goobi/Kitodo or maybe in the ZVDD METS Profile to use only one FLocat?

Apart from that, I'm open for the idea, but it will take some time because we have to change file handling in a few places for this (much like your AlternativeImage work, with additional checks and new possible points of failure in the logic).

kba:
Yes, the [ZVDD guidelines](https://www.zvdd.de/fileadmin/AGSDD-Redaktion/METS_Anwendungsprofil_2.1.pdf) are pretty restrictive. FLocat isn't repeatable. But it also says `@xlink:href` must be a URL which I tried to defend for the longest time but we're not abiding by anymore. So maybe repeated FLocat would be less intrusive than changing `xlink:href` in a destructive way as we do now...

[...]

PR #1079, opened 2023-08-20
Keep local copies of files in a separate mets:FLocat

kba:
One early decision that has haunted us for years now is that we have been using a single `mets:FLocat` for both the original URL of a `mets:file` and the local copy in the workspace we use for processing.

This PR tries to solve #323 by changing `OcrdFile` and the download logic in `Resolver` and `Workspace`:

- the original URL (`OcrdFile.url`) remains in `mets:FLocat[@LOCTYPE="URL"]/xlink:href`
- the local filename (`OcrdFile.local_filename`) will now be written to an additional `mets:Flocat[@LOCTYPE="OTHER"][@OTHERLOCTYPE="FILE"]/xlink:href`
- The contract of `Workspace.download_file` is that after calling it with `OcrdFile f`, `f` will have a `local_filename` attribute and that is what processors should use rather than the `url`.
- The logic of when to download and when to copy files in `Resolver.download_to_directory` and `Workspace.download_file` has been adapted accordingly.

The goal here is to make the OCR-D processing non-invasive. Currently, once you do `ocrd workspace find --download`, the original URL will be gone. With this PR, `ocrd workspace find --download` will add an additional `mets:Flocat` which can then be removed after processing is finished (to be compliant with the DFG Viewer METS profile) with `ocrd workspace find --undo-download`.


MehmedGIT:
When the files are downloaded there is an output with relative file path per line. When I did undo the downloading it returned just `None` on each line. Since the local path is removed, this makes sense, however, how should the general user interpret that output?

kba:
> Since the local path is removed, this makes sense, however, how should the general user interpret that output

Good point, I hadn't thought about that. Should be fixed in 0f26809e9

For the kant_aufklaerung_1784 test asset:

```
ocrd workspace find --download
OCR-D-IMG/INPUT_0017.tif
OCR-D-IMG/INPUT_0020.tif
OCR-D-GT-PAGE/PAGE_0017_PAGE.xml
OCR-D-GT-PAGE/PAGE_0020_PAGE.xml
OCR-D-GT-ALTO/PAGE_0017_ALTO.xml
OCR-D-GT-ALTO/PAGE_0020_ALTO.xml
```

and the reverse:

```
ocrd workspace find --undo-download
Removed local_filename OCR-D-IMG/INPUT_0017.tif
Removed local_filename OCR-D-IMG/INPUT_0020.tif
Removed local_filename OCR-D-GT-PAGE/PAGE_0017_PAGE.xml
Removed local_filename OCR-D-GT-PAGE/PAGE_0020_PAGE.xml
Removed local_filename OCR-D-GT-ALTO/PAGE_0017_ALTO.xml
Removed local_filename OCR-D-GT-ALTO/PAGE_0020_ALTO.xml
```

MehmedGIT:
Yes, that output is more convenient. Great!

ocrd_models/ocrd_models/ocrd_file.py
```diff
@@ -13,40 +14,35 @@ class OcrdFile():
...
         self._el = el
         self.mets = mets
         self.ID = ID
         self.mimetype = mimetype
-        self.local_filename = local_filename
-        self.loctype = loctype
         self.pageId = pageId
 
+        if local_filename:
+            self.local_filename = Path(local_filename)
```
bertsky, line 42: Should have been `str(Path(local_filename))`, right? @kba

kba: No, that was a conscious decision, though not optimally communicated. Changing the type helped me spot areas where `url` and `local_filename` were used interchangeably but I think it is a correct implementation that `local_filename` is a Path.

kba: Reverted in #1182

PR #1080, opened 2023-08-20
Refactor logging

kba:
From the module description:

- Try to be less intrusive with OCR-D specific logging conventions to
      make it easier and less surprising to define logging behavior when
      using OCR-D/core as a library
- Change setOverrideLogLevel to only override the log level of the ``ocrd``
      logger and its descendants
- initLogging will set exactly one handler, for the root logger or for the
      ``ocrd`` logger.
- Child loggers should propagate to the ancestor logging (default
      behavior of the logging library - no more PropagationShyLogger)
- disableLogging only removes any handlers from the ``ocrd`` logger

In short:

  - if you want the OCR-D logging conventions, use `ocrd_utils.logging.{initLogging,getLogger,disableLogging}`.
  - if you don't want OCR-D to mess with the root logger, call `initLogging(basic_config=False)`
  - if you want the builtin logging behavior only, call `initLogging(builtin_only)` which will omit the configuration file lookup
  - if you don't want the OCR-D logging conventions, just do not use them. The only side effect of (transitively) importing `ocrd_utils.logging` now is that all handlers added to the `ocrd` logger will be removed. I'm open for even removing that but need to think about it some more.

Until now, we implemented `--log-level`/`setOverrideLoglevel` by changing the default logger to one that does not propagate and overriding the levels of every logger. Now, we keep the default behavior (all loggers propagate) and only change the level of the `ocrd` logger. We leave the root logger alone completely.

### DIFF
--- a/ocrd_utils/ocrd_utils/__init__.py
+++ b/ocrd_utils/ocrd_utils/__init__.py
@@ -161,7 +161,6 @@ from .logging import (
     getLevelName,
     getLogger,
     initLogging,
-    logging,
     setOverrideLogLevel,
     )
 
--- a/ocrd_utils/ocrd_utils/logging.py
+++ b/ocrd_utils/ocrd_utils/logging.py
@@ -10,15 +10,29 @@ Logging can be overridden either programmatically in code using the library or b
 - $PWD/ocrd_logging.py
 
 These files will be executed in the context of ocrd/ocrd_logging.py, with `logging` global set.
+
+Changes as of 2023-08-20:
+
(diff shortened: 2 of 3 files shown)


kba:
It's finally passing all the tests, let's merge before another one breaks ;)

bertsky:
@kba 26967052 broke the existing API because `setOverrideLogLevel` used to handle OCR-D level names like `OFF`, but now does not anymore. Can we get the old behaviour back please?

PR #1081, opened 2023-08-21
ocrd_utils.config to handle all environment variables

kba:
We've been accumulating quite a lot of environment variables that define behavior in addition to the CLI options and other explicit mechanisms in the codebase. @bertsky started documenting all of those in 2d8550398 and this PR goes a step further to bundle all the functionality and documentation in one place.

In `ocrd_utils.config`, there is a singleton instance `config` of `ocrd_utils.config.OcrdEnvConfig` that should be used for registering environment variables and parsing them.

For example to define a new environemnt variable `OCRD_LOGGING_CONFIG`, we register it:

```
from pathlib import Path
from ocrd_utils import config
config.add('OCRD_LOGGING_CONFIG',
description="Path to logging configuration",
validator: lambda val: Path.exists(),
parser: lambda val: Path(val))
```

This environment variable can now be accessed as `config.OCRD_LOGGING_CONFIG` which will return a `pathlib.Path` because the `parser`. Since we're defining `validator`, this will fail as soon as `config.OCRD_LOGGING_CONFIG` is accessed, if the value is not an existing file and `config.OCRD_LOGGING.CONFIG` will return a `Path`.

From the `description` field, we generate the epilog to the `ocrd --help` output listing all the variables.

In short, whenever we need a new environment variable `X`:
- register it in `ocrd_utils.config`
- add it to the CLI epilog
- use it as `config.X`


ocrd_utils/ocrd_utils/config.py
```diff
@@ -0,0 +1,167 @@
...
+config.add("OCRD_DOWNLOAD_TIMEOUT",
+    description="Timeout in seconds for connecting or reading (comma-separated) when downloading.",
+    parser=_ocrd_download_timeout_parser)
+
+config.add("OCRD_NETWORK_SERVER_ADDR_PROCESSING",
+        description="Default address of Processing Server to connect to (for `ocrd network client processing`).",
+        default=(True, ''))
+
+config.add("OCRD_NETWORK_SERVER_ADDR_WORKFLOW",
+        description="Default address of Workflow Server to connect to (for `ocrd network client processing`).",
```
MehmedGIT, line 147: Should be `ocrd network client workflow`?

ocrd_utils/ocrd_utils/config.py
```diff
@@ -0,0 +1,167 @@
...
+        default=(True, ''))
+
+config.add("OCRD_NETWORK_SERVER_ADDR_WORKSPACE",
+        description="Default address of Workspace Server to connect to (for `ocrd network client processing`).",
```
MehmedGIT, line 151: Should be `ocrd network client workspace`?

MehmedGIT (approved):
It looks more organized now, I like it.

ocrd_utils/ocrd_utils/config.py
kba: Fixed in 1e35a5074

ocrd_utils/ocrd_utils/config.py
kba: Fixed in 1e35a5074

Issue #1032, opened 2023-03-29
`ocrd network` CLI syntax


kba:
From https://github.com/OCR-D/core/pull/974#issuecomment-1483588724

@MehmedGIT:

> Okay, another note that will potentially be raised soon. Currently, the supported processing-worker CLI is still the old one:
> 
> ```shell
> # 1.
> $ ocrd processing-worker <processor-name> --queue=<queue-address> --database=<database-address>
> 
> # 2.
> $ <processor-name> --queue=<queue-address> --database=<database-address>
> ```
> 
> We (me and @joschrew) are aware that in the [spec](https://github.com/OCR-D/spec/blob/master/web_api.md) there is a change that is not adapted here, yet:
> 
> ```shell
> # 1. Use ocrd CLI bundled with OCR-D/core
> $ ocrd server <processor-name> --type=worker --queue=<queue-address> --database=<database-address>
> 
> # 2. Use processor name
> $ <processor-name> --server --type=worker --queue=<queue-address> --database=<database-address>
> ```
> 
> Where the `--type` can be either: `worker` (processing worker)
> 
> ```shell
> ocrd server <processor-name> --type=worker --queue=<queue-address> --database=<database-address>
> ```
> 
> or `server` (the REST API wrapper based on #884).
> 
> ```shell
> ocrd server <processor-name> --type=server --database=<database-address>
> ```
> 
> However, it's a bad idea to extend the processing worker code to support the REST API processing server (aka standalone processing worker that has nothing to do with the bigger Processing Server in the spec and does not need the queues).
> 
> Now, try to imagine a good way to explain to the OCR-D community without confusing them:
> 
> 1. why the `<processor-name>` is a server, but is not a server when `--type=worker` and is referred to with `processing-worker` and no direct requests can be sent
> 2. why the `<processor-name>` is a server, and actual server when `--type=server` and is referred to with `processing-server`
> 3. why both are grouped together under `ocrd server...` or `... --server ...` but potentially implemented together under `processing_worker.py`
> 4. why Processing Server (PS-big) and processing servers (PS-small), standalone ocrd processors, are different concepts but both referred to with `processing server`.
> 
> Sounds confusing? It's even more when it has to be implemented in a clean way. So there are 3 main reasons to not adapt that yet:
> 
> * The priority changed a bit and the higher priority now is to deploy working Processing Server, Workflow Server, and Workspace Server together on a live VM instance, so KITODO can use that to continue their development.
[...]

PR #1083, opened 2023-08-28
Workflow endpoint of the Processing Server

joschrew:
Endpoint to run a workflow with the processing-server

Usage described [here](https://github.com/joschrew/workflow-endpoint-usage-example)

### DIFF
--- a/ocrd_network/ocrd_network/database.py
+++ b/ocrd_network/ocrd_network/database.py
@@ -13,9 +13,11 @@ XXX: Currently the information is not preserved after the processing-server shut
 database (runs in docker) currently has no volume set.
 """
 from beanie import init_beanie
+from beanie.operators import In
 from motor.motor_asyncio import AsyncIOMotorClient
 from uuid import uuid4
 from pathlib import Path
+from typing import List
 
 from .models import (
     DBProcessorJob,
@@ -173,3 +175,25 @@ async def db_update_processing_job(job_id: str, **kwargs) -> DBProcessorJob:
 @call_sync
 async def sync_db_update_processing_job(job_id: str, **kwargs) -> DBProcessorJob:
     return await db_update_processing_job(job_id=job_id, **kwargs)
+
+
+async def db_get_workflow_job(job_id: str) -> DBWorkflowJob:
+    job = await DBWorkflowJob.find_one(DBWorkflowJob.job_id == job_id)
+    if not job:
+        raise ValueError(f'Workflow job with id "{job_id}" not in the DB.')
+    return job
+
+
+@call_sync
+async def sync_db_get_workflow_job(job_id: str) -> DBWorkflowJob:
+    return await db_get_workflow_job(job_id)
+
+
+async def db_get_processing_jobs(job_ids: List[str]) -> [DBProcessorJob]:
+    jobs = await DBProcessorJob.find(In(DBProcessorJob.job_id, job_ids)).to_list()
+    return jobs
+
+
+@call_sync
+async def sync_db_get_processing_jobs(job_ids: List[str]) -> [DBProcessorJob]:
+    return await db_get_processing_jobs(job_ids)
--- a/ocrd_network/ocrd_network/database.py
+++ b/ocrd_network/ocrd_network/database.py
@@ -41,7 +41,7 @@ async def db_create_workspace(mets_path: str) -> DBWorkspace:
     """ Create a workspace-database entry only from a mets-path
     """
     if not Path(mets_path).exists():
-        raise ValueError(f'Cannot create DB workspace entry, `{mets_path}` does not exist!')
+        raise FileNotFoundError(f'Cannot create DB workspace entry, `{mets_path}` does not exist!')
     try:
         return await db_get_workspace(workspace_mets_path=mets_path)
     except ValueError:
--- a/ocrd_network/ocrd_network/database.py
+++ b/ocrd_network/ocrd_network/database.py
@@ -118,8 +118,6 @@ async def db_update_workspace(workspace_id: str = None, workspace_mets_path: str
             workspace.bag_info_adds = value
         elif key == 'deleted':
             workspace.deleted = value
-        elif key == 'pages_locked':
-            workspace.pages_locked = value
         elif key == 'mets_server_url':
             workspace.mets_server_url = value
         else:
--- a/ocrd_network/ocrd_network/database.py
+++ b/ocrd_network/ocrd_network/database.py
@@ -19,6 +19,7 @@ from pathlib import Path
 
 from .models import (
     DBProcessorJob,
+    DBWorkflowJob,
     DBWorkspace
 )
 from .utils import call_sync
@@ -28,7 +29,7 @@ async def initiate_database(db_url: str):
     client = AsyncIOMotorClient(db_url)
     await init_beanie(
         database=client.get_default_database(default='ocrd'),
-        document_models=[DBProcessorJob, DBWorkspace]
+        document_models=[DBProcessorJob, DBWorkflowJob, DBWorkspace]
     )
 
 
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -564,4 +564,9 @@ class Deployer:
         )
         '''
 
+        # TODO: Reconsider this again
+        #  Not having this sleep here causes connection errors
+        #  on the last request processed by the processing worker.
+        #  Sometimes 3 seconds is enough, sometimes not.
+        sleep(5)
         stop_mets_server(mets_server_url=mets_server_url)
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -513,6 +513,7 @@ class Deployer:
         stdin, stdout = channel.makefile('wb'), channel.makefile('rb')
         cmd = f'{processor_name} server --address {agent_address} --database {database_url}'
         stdin.write(f"echo starting processor server with '{cmd}'\n")
+        stdin.write(f'{cmd} &\n')
         stdin.write('echo xyz$!xyz \n exit \n')
         output = stdout.read().decode('utf-8')
         stdout.close()
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -474,7 +474,7 @@ class Deployer:
         # printed with `echo $!` but it is printed inbetween other output. Because of that I added
         # `xyz` before and after the code to easily be able to filter out the pid via regex when
         # returning from the function
-        log_path = '/tmp/ocrd-processing-server-startup.log'
+        log_path = f'/tmp/deployed_{processor_name}.log'
         stdin.write(f"echo starting processing worker with '{cmd}' >> '{log_path}'\n")
         stdin.write(f'{cmd} >> {log_path} 2>&1 &\n')
         stdin.write('echo xyz$!xyz \n exit \n')
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -512,11 +512,7 @@ class Deployer:
         channel = ssh_client.invoke_shell()
         stdin, stdout = channel.makefile('wb'), channel.makefile('rb')
         cmd = f'{processor_name} server --address {agent_address} --database {database_url}'
-        port = agent_address.split(':')[1]
-        log_path = f'/tmp/server_{processor_name}_{port}_{getpid()}.log'
-        # TODO: This entire stdin/stdout thing is broken with servers!
-        stdin.write(f"echo starting processor server with '{cmd}' >> '{log_path}'\n")
-        stdin.write(f'{cmd} >> {log_path} 2>&1 &\n')
+        stdin.write(f"echo starting processor server with '{cmd}'\n")
         stdin.write('echo xyz$!xyz \n exit \n')
         output = stdout.read().decode('utf-8')
         stdout.close()
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -30,7 +30,11 @@ from .runtime_data import (
     DataProcessorServer,
     DataRabbitMQ
 )
-from .utils import validate_and_load_config
+from .utils import (
+    is_mets_server_running,
+    stop_mets_server,
+    validate_and_load_config
+)
 
 
 class Deployer:
@@ -521,13 +525,16 @@ class Deployer:
 
     # TODO: No support for TCP version yet
     def start_unix_mets_server(self, mets_path: str) -> str:
-        mets_server_url = f'/tmp/{safe_filename(mets_path)}'
-        if Path.exists(Path(mets_server_url)):
+        socket_file = f'{safe_filename(mets_path)}.sock'
+        log_path = f'/tmp/{safe_filename(mets_path)}.log'
+        mets_server_url = f'/tmp/{socket_file}'
+
+        if is_mets_server_running(mets_server_url=mets_server_url):
             self.log.info(f"The mets server is already started: {mets_server_url}")
             return mets_server_url
+
         cwd = Path(mets_path).parent
-        self.log.info(f'Starting native mets server: {mets_server_url}')
-        log_path = f'{mets_server_url}.log'
+        self.log.info(f'Starting UDS mets server: {mets_server_url}')
         sub_process = subprocess.Popen(
             args=['nohup', 'ocrd', 'workspace', '--mets-server-url', f'{mets_server_url}',
                   '-d', f'{cwd}', 'server', 'start'],
@@ -543,14 +550,18 @@ class Deployer:
         return mets_server_url
 
     def stop_unix_mets_server(self, mets_server_url: str) -> None:
-        self.log.info(f'Stopping native mets server: {mets_server_url}')
+        self.log.info(f'Stopping UDS mets server: {mets_server_url}')
         if mets_server_url in self.mets_servers:
             mets_server_pid = self.mets_servers[mets_server_url]
         else:
             raise Exception(f"Mets server not found: {mets_server_url}")
 
+        '''
         subprocess.run(
             args=['kill', '-s', 'SIGINT', f'{mets_server_pid}'],
             shell=False,
             universal_newlines=True
         )
+        '''
+
+        stop_mets_server(mets_server_url=mets_server_url)
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -39,7 +39,7 @@ from .utils import (
 
 class Deployer:
     def __init__(self, config_path: str) -> None:
-        self.log = getLogger(__name__)
+        self.log = getLogger('ocrd_network.deployer')
         config = validate_and_load_config(config_path)
 
         self.data_mongo: DataMongoDB = DataMongoDB(config['database'])
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -474,9 +474,12 @@ class Deployer:
         # printed with `echo $!` but it is printed inbetween other output. Because of that I added
         # `xyz` before and after the code to easily be able to filter out the pid via regex when
         # returning from the function
-        log_path = f'/tmp/deployed_{processor_name}.log'
-        stdin.write(f"echo starting processing worker with '{cmd}' >> '{log_path}'\n")
-        stdin.write(f'{cmd} >> {log_path} 2>&1 &\n')
+
+        # TODO: Check here again
+        # log_path = f'/tmp/deployed_{processor_name}.log'
+        # stdin.write(f"echo starting processing worker with '{cmd}' >> '{log_path}'\n")
+        # stdin.write(f'{cmd} >> {log_path} 2>&1 &\n')
+        stdin.write(f'{cmd} &\n')
         stdin.write('echo xyz$!xyz \n exit \n')
         output = stdout.read().decode('utf-8')
         stdout.close()
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -9,7 +9,7 @@ Each Processing Worker is an instance of an OCR-D processor.
 from __future__ import annotations
 from typing import Dict, List, Union
 from re import search as re_search
-from os import getpid, chmod
+from os import getpid
 from pathlib import Path
 import subprocess
 from time import sleep
@@ -528,7 +528,6 @@ class Deployer:
         cwd = Path(mets_path).parent
         self.log.info(f'Starting native mets server: {mets_server_url}')
         log_path = f'{mets_server_url}.log'
-        # cmd = f'nohup ocrd workspace --mets-server-url {mets_server_url} -d {cwd} server start &> {log_path} &'
         sub_process = subprocess.Popen(
             args=['nohup', 'ocrd', 'workspace', '--mets-server-url', f'{mets_server_url}',
                   '-d', f'{cwd}', 'server', 'start'],
@@ -538,12 +537,8 @@ class Deployer:
             cwd=cwd,
             universal_newlines=True
         )
-
-        # TODO: Remove chmod when the default permissions of the socket files are changed in core
-        from time import sleep
-        sleep(3)
-        chmod(mets_server_url, 0o0777)
-
+        # Wait for the mets server to start
+        sleep(2)
         self.mets_servers[mets_server_url] = sub_process.pid
         return mets_server_url
 
--- a/ocrd_network/ocrd_network/process_helpers.py
+++ b/ocrd_network/ocrd_network/process_helpers.py
@@ -1,5 +1,6 @@
 import json
 from typing import List, Optional
+from pathlib import Path
 
 from ocrd import Resolver
 from ocrd.processor.helpers import run_cli, run_processor
@@ -7,7 +8,6 @@ from ocrd.processor.helpers import run_cli, run_processor
 
 # A wrapper for run_processor() and run_cli()
 def invoke_processor(
-        logger,
         processor_class,
         executable: str,
         abs_path_to_mets: str,
@@ -22,13 +22,15 @@ def invoke_processor(
     input_file_grps_str = ','.join(input_file_grps)
     output_file_grps_str = ','.join(output_file_grps)
 
-    # TODO: Passing the `mets_server_url` is a must here.
-    #  Seems like just passing it inside `run_processor` is not enough.
-    #  Potentially a bug to be resolved in core.
+    if not mets_server_url.startswith('http://'):
+        if not Path.exists(Path(mets_server_url)):
+            raise RuntimeError(f'Socket file does not exist: {mets_server_url}')
+
     workspace = Resolver().workspace_from_url(
         mets_url=abs_path_to_mets,
         mets_server_url=mets_server_url
     )
+
     if processor_class:
         try:
             run_processor(
@@ -42,7 +44,7 @@ def invoke_processor(
                 mets_server_url=mets_server_url
             )
         except Exception as e:
-            raise RuntimeError(f"Python executable '{executable}' exited with: {e}")
+            raise RuntimeError(f"Python executable '{processor_class.__dict__}' exited with: {e}")
     else:
         return_code = run_cli(
             executable=executable,
--- a/ocrd_network/ocrd_network/process_helpers.py
+++ b/ocrd_network/ocrd_network/process_helpers.py
@@ -27,7 +27,7 @@ def invoke_processor(
 
     ctx_mgr = redirect_stderr_and_stdout_to_file(log_filename) if log_filename else nullcontext()
     with ctx_mgr:
-        initLogging()
+        initLogging(force_reinit=True)
         workspace = get_ocrd_workspace_instance(
             mets_path=abs_path_to_mets,
             mets_server_url=mets_server_url
--- a/ocrd_network/ocrd_network/process_helpers.py
+++ b/ocrd_network/ocrd_network/process_helpers.py
@@ -1,9 +1,11 @@
 import json
 from typing import List, Optional
 import logging
+from contextlib import nullcontext
 
 from ocrd.processor.helpers import run_cli, run_processor
 from .utils import get_ocrd_workspace_instance
+from ocrd_utils import redirect_stderr_and_stdout_to_file
 
 
 # A wrapper for run_processor() and run_cli()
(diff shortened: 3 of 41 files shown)


ocrd_network/ocrd_network/processing_server.py
```diff
@@ -670,3 +687,41 @@ async def list_processors(self) -> List[str]:
...
+                path_to_mets=workspace_path,
+                input_file_grps=task.input_file_grps,
+                output_file_grps=task.output_file_grps,
+                parameters=task.parameters,
+                callback_url=callback_url,
+                depends_on=[last_job_id] if last_job_id else [],
+            )
+            output = await self.push_processor_job(task.executable, data)
+            outputs.append(output)
+            last_job_id = output.job_id
```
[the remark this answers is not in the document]
kba: Also, the `page_id` attribute is missing completely right now. Is there a reason not to do page-wise processing?

Pseudo-algorithm:

```python
page_range = expand_page_ids(request.page_id) if request.page_id else workspace.physical_pages
depends_on = None
for task in tasks:
  for page in page_range:
                data = PYJobInput(
                agent_type='worker',
                path_to_mets=workspace_path,
                input_file_grps=task.input_file_grps,
                output_file_grps=task.output_file_grps,
                parameters=task.parameters,
                page_id=page
                callback_url=callback_url,
                depends_on=[depends_on] if depends_on else [],
            )
```

As I said, pseudo-code, just to illustrate the idea.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -670,3 +687,41 @@ async def list_processors(self) -> List[str]:
             unique_only=True
         )
         return processor_names_list
+
+    # TODO: think about providing arguments in another way
+    # TODO: this function "just" writes to the queue and returns. A network-client functionality
+    #       should be available to react to everys processors callback. With this feedback
+    #       a blocking mechanism could be provided to inform about starting the cain and waiting for
+    #       the processors to finish and printing when reponses are received from the processors
```
kba, line 695: Indeed, that is a tricky but important aspect- how can we manage the callbacks in such a way that it is useful to the user? We could calculate the number of jobs we'll create (i.e. the output of this to-be-refactored `task_sequence_to_processing_jobs` function I mention below) and send back the current state, i.e. `Job {uuid} [30 of 123] finished {"successful" if state == 'SUCCESS` else "with an eror"}.

But more importantly, could we design this in such a way that we only give a callback if a) one job has failed or b) the complete workflow succeeded? Can we perhaps add a "session" ID to the jobs which the processing server can track and call the callback only for the aforementioned cases?

ocrd_network/ocrd_network/server_utils.py
```diff
@@ -78,10 +80,29 @@ def validate_job_input(logger, processor_name: str, ocrd_tool: dict, job_input:
...
+    except ValueError:
+        workspace_db = DBWorkspace(
+            workspace_id=str(uuid.uuid4()),
+            workspace_path=Path(workspace_path).parent,
+            workspace_mets_path=workspace_path,
+            ocrd_identifier="",
+            bagit_profile_identifier="",
+        )
+        await workspace_db.save()
+        return workspace_db
```
kba, line 108: This should be moved to `.database`

kba (changes requested):

ocrd_network/ocrd_network/server_utils.py
MehmedGIT: Done

ocrd_network/ocrd_network/processing_server.py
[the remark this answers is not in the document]
MehmedGIT: Doing page-wise processing and just processing a sub-domain of pages is different, right? It feels like when the `page_id` is provided in the `run_workflow` method, only the range of `page_id` must be processed. This should not affect whether it is page-wise or not. I have implemented that now.

For page-wise processing, it would be better to introduce a flag that enables/disables page-wise processing.  If set, for each page a separate processing request would be created (as in your example), if not set, a single request with the `page_id` domain.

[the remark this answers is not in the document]
MehmedGIT: > We should put that into a separate function, that generates a list of dicts, so we can test the behavior properly.

A list of `PYJobInput` or `PYJobOutput`? If the former, then it is a bit tricky since the `job_id` is not produced yet by the Processing Server. Hence, we cannot do dependence chaining based on `job_id`. If the latter, then it is doable.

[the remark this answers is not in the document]
MehmedGIT: > For page-wise processing, it would be better to introduce a flag that enables/disables page-wise processing. If set, for each page a separate processing request would be created (as in your example), if not set, a single request with the page_id domain.

We should then also consider what happens if no `page_id` is provided, but the flag for page-wise processing is set. How do we get the total amount of pages of a workspace? Simply using `workspace.physical_pages` would be inefficient since we are interested only in the number of pages, not the specific page object.

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: I am not sure I fully understand the idea. The `callback_url` passed to the `run_workflow` method is there to be called only once - when the entire workflow succeeds or fails.

kba: Oh, okay, I thought the `callback_url` was to be passed on to every `PYInputJob` and I expected it to be called when every **job** is finished, not just if the `workflow` finished. If that is the case already, then disregard my comment :)

MehmedGIT: Maybe we should think about what response to return instead of returning just a list of `PYJobOutput`s to make it more useful for the user. The `callback_url` concept was planned to be used by the `Workflow Server` to report the workflow status. Not sure how this should be used to make it useful to the user as well.

ocrd_network/ocrd_network/processing_server.py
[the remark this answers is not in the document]
MehmedGIT: This is now implemented but with a few changes:
- for page-wise processing, it is a must for the user to specify the `page_id` when the `page_wise` flag is set. Utilizing `workspace.physical_pages` could have unpredictable side effects when used together with the mets server.
- the outer loop is based on the `page` not on the `task` since the other way around messes up the dependency relation among jobs. It creates a dependency both in pages and tasks, but there is no tight dependency among the pages since we already have an output file group locking mechanism for that.

MehmedGIT:
Another point is that a better response is needed for the workflow endpoint for user convenience. Just a list of processing job responses leads to a duplication of the fields (e.g., `mets_path`), and is hard to track the job dependencies.

ocrd_network/ocrd_network/database.py
```diff
@@ -35,6 +37,25 @@ async def sync_initiate_database(db_url: str):
     await initiate_database(db_url)
 
 
+async def db_create_workspace(mets_path: str) -> DBWorkspace:
+    """ Create a workspace-database entry only from a mets-path
+    """
+    if not Path(mets_path).exists():
+        raise ValueError(f'Cannot create DB workspace entry, `{mets_path}` does not exist!')
```
kba, line 44:
```suggestion
        raise FileNotFoundError(f'Cannot create DB workspace entry, `{mets_path}` does not exist!')
```

ocrd_network/ocrd_network/database.py
```diff
@@ -35,6 +37,25 @@ async def sync_initiate_database(db_url: str):
...
+    try:
+        return await db_get_workspace(workspace_mets_path=mets_path)
+    except ValueError:
```
kba, line 47: Why does this (potentially) raise a ValueError?

ocrd_network/ocrd_network/deployer.py
```diff
@@ -515,3 +518,39 @@ def start_native_processor_server(
         stdout.close()
         stdin.close()
         return re_search(r'xyz([0-9]+)xyz', output).group(1)  # type: ignore
+
+    # TODO: No support for TCP version yet
+    def start_unix_mets_server(self, mets_path: str) -> str:
+        mets_server_url = f'/tmp/{safe_filename(mets_path)}'
+        if Path.exists(Path(mets_server_url)):
```
kba, line 525: Perhaps try a request to `/workspace_name` to test whether it is running, that works for both UDS and HTTP server.

ocrd_network/ocrd_network/deployer.py
```diff
@@ -515,3 +518,39 @@ def start_native_processor_server(
...
+            cwd=cwd,
+            universal_newlines=True
+        )
+        # Wait for the mets server to start
+        sleep(2)
+        self.mets_servers[mets_server_url] = sub_process.pid
+        return mets_server_url
+
+    def stop_unix_mets_server(self, mets_server_url: str) -> None:
+        self.log.info(f'Stopping native mets server: {mets_server_url}')
```
kba, line 546:
```suggestion
        self.log.info(f'Stopping mets server: {mets_server_url}')
```

What does "native" mean here?

ocrd_network/ocrd_network/deployer.py
```diff
@@ -515,3 +518,39 @@ def start_native_processor_server(
...
+        if mets_server_url in self.mets_servers:
+            mets_server_pid = self.mets_servers[mets_server_url]
+        else:
+            raise Exception(f"Mets server not found: {mets_server_url}")
+
+        subprocess.run(
```
kba, line 552: Possible but probably cleaner to http GET `/stop`.

ocrd_network/ocrd_network/process_helpers.py
```diff
@@ -14,12 +15,23 @@ def invoke_processor(
...
+
+    if mets_server_url:
+        if not mets_server_url.startswith('http://'):
+            if not Path.exists(Path(mets_server_url)):
+                raise RuntimeError(f'Socket file does not exist: {mets_server_url}')
+
+    workspace = Resolver().workspace_from_url(
+        mets_url=abs_path_to_mets,
+        mets_server_url=mets_server_url
+    )
```
kba, line 33: This would not work for the hostname:port server. So if checking is necessary, why not create the workspace and try to request `/workspace_path`?

ocrd_network/ocrd_network/process_helpers.py
```diff
@@ -29,10 +41,11 @@ def invoke_processor(
                 output_file_grp=output_file_grps_str,
                 page_id=page_id,
                 parameter=parameters,
-                instance_caching=True
+                instance_caching=True,
+                mets_server_url=mets_server_url
             )
         except Exception as e:
-            raise RuntimeError(f"Python executable '{executable}' exited with: {e}")
+            raise RuntimeError(f"Python executable '{processor_class.__dict__}' exited with: {e}")
```
kba, line 48: Why that change? If for debugging, perhaps extend to include both executable and the class dict?

kba (commented):
First half of review, second half after our call.

ocrd_network/ocrd_network/database.py
MehmedGIT: When neither `workspace_mets_path` nor `workspace_id` matches some database document representing the Workspace.

ocrd_network/ocrd_network/deployer.py
MehmedGIT: It meant `UDS`. Replacing it now.

ocrd_network/ocrd_network/process_helpers.py
MehmedGIT: Yes, I will implement a method that checks that and use it in both the processing server and processing worker.

ocrd_network/ocrd_network/process_helpers.py
MehmedGIT: It was just overlooked in the previous implementation. `{executable}` for `run_cli`, but `processor_class.__dict__` for `run_processor`. It was just `{executable}` for both and always resolved to None in the `run_processor` case.

ocrd_network/ocrd_network/deployer.py
MehmedGIT: There is `/workspace_path` that returns `workspace.directory` as content. Should we implement another endpoint just for checking the running status of the server that returns just a status code of 200?

ocrd_network/ocrd_network/deployer.py
MehmedGIT: That's just sending a DELETE to `/`. Could you return some response before the shutdown so the one stopping the server knows if it was successful or not?

MehmedGIT:
I have just realized that the very last processing request always fails due to some connection error with the mets server. @kba, could you investigate? I have tried to put some sleep before the mets server shutdown to make sure `workspace.save()` gets enough time. Still, fails.

```
  File "/home/mm/Desktop/core/ocrd_network/ocrd_network/process_helpers.py", line 44, in invoke_processor
    mets_server_url=mets_server_url
  File "/home/mm/Desktop/core/ocrd/ocrd/processor/helpers.py", line 132, in run_processor
    raise err
  File "/home/mm/Desktop/core/ocrd/ocrd/processor/helpers.py", line 129, in run_processor
    processor.process()
  File "/home/mm/venv37-ocrd/lib/python3.7/site-packages/ocrd_tesserocr/recognize.py", line 459, in process
    self._process_existing_regions(tessapi, regions, page_image, page_coords, pcgts_mapping)
  File "/home/mm/venv37-ocrd/lib/python3.7/site-packages/ocrd_tesserocr/recognize.py", line 950, in _process_existing_regions
    self._process_existing_lines(tessapi, textlines, region_image, region_coords, mapping)
  File "/home/mm/venv37-ocrd/lib/python3.7/site-packages/ocrd_tesserocr/recognize.py", line 961, in _process_existing_lines
    line, region_image, region_coords)
  File "/home/mm/Desktop/core/ocrd/ocrd/workspace.py", line 964, in image_from_segment
    segment_image = self._resolve_image_as_pil(alternative_image.get_filename())
  File "/home/mm/Desktop/core/ocrd/ocrd/workspace.py", line 490, in _resolve_image_as_pil
    f = next(self.mets.find_files(local_filename=str(image_url)))
  File "/home/mm/Desktop/core/ocrd/ocrd/mets_server.py", line 127, in find_files
    r = self.session.request('GET', f'{self.url}/file', params={**kwargs})
  File "/home/mm/venv37-ocrd/lib/python3.7/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/mm/venv37-ocrd/lib/python3.7/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/home/mm/venv37-ocrd/lib/python3.7/site-packages/requests/adapters.py", line 547, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```

MehmedGIT:
Deciding when to shut down the METS Server is more complicated than I was expecting. The internal request cache for a workspace could become empty before all processing workers are done with the processing. So shutting the METS Server when the cache is empty is an unreliable solution.

kba:
>  From the error, it seems the server shuts down before the client request is complete.

If it is just about saving the mets file before the server is shut down (which should happen on shutdown but it is possible that it does not handle the `kill` command gracefully - I'll investigate), you can call `ClientSideOcrdMets.save()` or `http PUT /` on the endpoint.

But I guess the really tricky problem is indeed

> Deciding when to shut down the METS Server is more complicated than I was expecting.

Could we reuse the callback functionality for this and delay shutting down the server until the jobs are complete?

Alternatively, we could add a (long) timeout to the METS server itself, say 5 minutes. The timer resets whenever an operation is triggered but will shut down automatically if there was no activity for 5 minutes. I see no real drawback against having the METS server around longer, at least for the UNIX socket variant - while it's idle, there's very little consumption of resources.

MehmedGIT:
> If it is just about saving the mets file before the server is shut down (which should happen on shutdown but it is possible that it does not handle the `kill` command gracefully - I'll investigate), you can call `ClientSideOcrdMets.save()` or `http PUT /` on the endpoint.

No, that is not the problem. I have already implemented `http DELETE /` which invokes the `/stop` method in the METS Server and `kill` is no longer utilized. The problem is that the cache gets empty before all the processing is done. The last processor in the workflow fails for the last 8-10 pages depending on the workspace size.

> Could we reuse the callback functionality for this and delay shutting down the server until the jobs are complete?

How exactly? The issue here is that with the page-wise processing, we no longer have dependency among pages. The last step in the workflow for each page has the potential to be the last request that is finished processing. Since the last page theoretically (and I observed it practically as well) could be processed before other pages, we could not simply say the last page of the last processor will trigger another callback functionality to stop the METS Server.

> Alternatively, we could add a (long) timeout to the METS server itself, say 5 minutes. The timer resets whenever an operation is triggered but will shut down automatically if there was no activity for 5 minutes.

I am afraid that this is the only solution that could work with the `ocrd_network`. The METS Server timeout should be configurable through an env variable which defaults to 5 mins. However, we should keep in mind that the timeout value should be longer for pages that take longer than 5 minutes to process.

On the other hand, handling the start and stop of the METS Server is not an issue in the HPC environment, since the batch script itself handles the start/stop, respectively, before and after the Nextflow script finishes.

MehmedGIT:
Here is another idea that should work in theory: For each workspace, i.e., mets server, the processing server could also have a counter which increases every time a request for that workspace is pushed to the RabbitMQ and decreases when a result callback for that workspace is invoked. When the counter reaches 0, then the shutdown will happen. The internal queue size of a workspace becomes 0 when each request inside is pushed to the RabbitMQ queues, but that counter will become 0 only when all requests call the internal result callback.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -554,63 +587,59 @@ async def get_processor_job(self, processor_name: str, job_id: str) -> PYJobOutp
...
+        if result_job_state == StateEnum.failed:
+            if len(self.processing_requests_cache[workspace_key]):
+                self.log.debug(f"Cancelling jobs dependent to job_id: {result_job_id}")
+                await self.cancel_dependent_jobs(
+                    job_id=result_job_id,
+                    internal_queue=self.processing_requests_cache[workspace_key]
+                )
+
+        if result_job_state != StateEnum.success:
+            # TODO: Handle other potential error cases
```
kba, line 617: A log statement would be a good first step and helpful for debugging.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -554,63 +587,59 @@ async def get_processor_job(self, processor_name: str, job_id: str) -> PYJobOutp
...
+            self.log.exception(f"Processing job with id: {result_job_id} not found in DB")
+
+        # Unlock the output file group pages for the result processing request
+        await self.unlock_pages(
+            db_workspace=db_workspace,
+            output_file_grps=db_result_job.output_file_grps,
+            page_ids=expand_page_ids(db_result_job.page_id)
+        )
+
         if not len(self.processing_requests_cache[workspace_key]):
```
kba, line 631: Just to be sure: The logic of the function prevents that there could be stale jobs in the cache, right? I.e. they are (recursively) cancelled/removed in case of any failure?

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -554,63 +587,59 @@ async def get_processor_job(self, processor_name: str, job_id: str) -> PYJobOutp
...
             # The queue is empty - delete it
             try:
                 del self.processing_requests_cache[workspace_key]
             except KeyError:
                 self.log.warning(f"Trying to delete non-existing internal queue with key: {workspace_key}")
+
+            # Shut down the Mets Server for the workspace_key
```
kba, line 638: See main discussion. IIUC we have discussed three solutions

1. Highjack the callback mechanism to be notified when all jobs are finished
2. Have an inactivity timeout in the METS server to automatically shut down when no more files are added
3. Poll the number of jobs completed/failed and shutdown once it reaches 0

Solution 1) is difficult to implement, especially when doing page-wise processing because the pages do not depend on each other (?)

Solution 2) would make sense in any case because there can always be situations where it is inconvenient to explicitly shut down the server but for the processing-server-use-case it is inefficient

Solution 3) is as robust as the queue/cache are robust, so as long as there is no danger of stale jobs lingering, it might be the way to go.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -670,3 +700,116 @@ async def list_processors(self) -> List[str]:
...
+                depends_on=dependent_jobs,
+            )
+            response = await self.push_processor_job(
+                processor_name=job_input_data.processor_name,
+                data=job_input_data
+            )
+            for file_group in task.output_file_grps:
+                file_group_cache[file_group] = response.job_id
+            responses.append(response)
+        return responses
```
kba, line 738: Looks good, that should be general enough to handle even dependencies of non-contiguous jobs AFAICT.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -670,3 +700,116 @@ async def list_processors(self) -> List[str]:
...
+            print(e)
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                                detail=f"Error parsing tasks: {e}")
+
+        if page_id:
+            page_range = expand_page_ids(page_id)
+        else:
+            # If no page_id is specified, all physical pages are assigned as page range
+            page_range = get_ocrd_workspace_physical_pages(mets_path=mets_path)
+        compact_page_range = f'{page_range[0]}..{page_range[-1]}'
```
kba, line 771: That seems unnecessary, you can just pass `None` as page_id to get the same behavior.

kba: Ah, no wait, you're right, this is for the database, which does not know how many pages there are in the workspace.

ocrd_network/ocrd_network/processing_worker.py
```diff
@@ -51,6 +51,8 @@ def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict,
         file_handler.setFormatter(logging.Formatter(logging_format))
         file_handler.setLevel(logging.DEBUG)
         self.log.addHandler(file_handler)
+        # TODO: remove this when refactoring the logging
+        self.log.setLevel(logging.DEBUG)
```
kba, line 55: I thought I had removed that in #1101 - need to check again

ocrd_network/ocrd_network/utils.py
```diff
@@ -110,3 +111,59 @@ def post_to_callback_url(logger, callback_url: str, result_message: OcrdResultMe
...
+    physical_pages = get_ocrd_workspace_instance(
+        mets_path=mets_path,
+        mets_server_url=mets_server_url
+    ).mets.physical_pages
+
+    # This conversion is required since
+    # the type of physical pages is not str
+    str_physical_pages = []
+    for physical_page in physical_pages:
+        str_physical_pages.append(f'{physical_page}')
```
kba, line 138: Why is that conversion necessary, `physical_pages` should return a list of `str`?

ocrd_network/ocrd_network/utils.py
```diff
@@ -110,3 +111,59 @@ def post_to_callback_url(logger, callback_url: str, result_message: OcrdResultMe
...
+        response = requests.get(url=f'{mets_server_url}/workspace_path', headers=headers)
+    else:
+        unix_url = f'http+unix://{mets_server_url.replace("/", "%2F")}'
+        try:
+            response = requests_unixsocket.Session().get(url=f'{unix_url}/workspace_path')
+        except Exception:
+            return False
+    if response.status_code == 200:
+        return True
+    return False
```
kba, line 154: A `log.debug` of the `response` might be useful here, since `/workspace_path` should always return 200 and the path if the server is running.

ocrd_network/ocrd_network/utils.py
```diff
@@ -110,3 +111,59 @@ def post_to_callback_url(logger, callback_url: str, result_message: OcrdResultMe
...
+
+
+def stop_mets_server(mets_server_url: str) -> bool:
+    if mets_server_url.startswith('http://') or mets_server_url.startswith('https://'):
+        headers = {"Content-Type": "application/json"}
+        response = requests.delete(url=f'{mets_server_url}/', headers=headers)
+    else:
+        unix_url = f'http+unix://{mets_server_url.replace("/", "%2F")}'
+        try:
+            response = requests_unixsocket.Session().delete(f'{unix_url}/')
```
kba, line 164: You can slightly simplify this with using a session for both UDS and TCP server, like in the constructor of ClientSideOcrdMets:

```
    def __init__(self, url):                                                                      
        protocol = 'tcp' if url.startswith('http://') else 'uds'                                  
        self.log = getLogger(f'ocrd.mets_client[{url}]')                                          
        self.url = url if protocol == 'tcp' else f'http+unix://{url.replace("/", "%2F")}'         
        self.session = requests_session() if protocol == 'tcp' else requests_unixsocket_session() 
```

Also why is the Content-Type necessary for the host:port variant?

kba (approved):
Looks good, the only thing missing is a proper way to shut down the METS server AFAICT.

ocrd_network/ocrd_network/processing_server.py
kba: Forgot "solution 4" which is to just not handle shutdown of the METS server at all and hope that everything is shut down gracefully when the processing server is stopped.

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: The resulting `job_id` and `job_state` are already printed (at the beginning) when the callback method is invoked.

MehmedGIT: But I could refactor it, yes.

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: I have implemented Solution 3). It seems to work just fine.

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: Yes, that's the case.

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: Yes. With the `comapct_page_rage` the `page_id` is now always passed to lower levels with all pages when `page_id` is not set.

ocrd_network/ocrd_network/processing_worker.py
MehmedGIT: The merge of master with #1101 to #1083 completely broke the logging in `ocrd_network`. Thus, I reverted the merge.

ocrd_network/ocrd_network/utils.py
MehmedGIT: It does not. I spent some time debugging that. The `physical_pages` returns a list of XML tree elements. Then when the `OcrdProcessingMessage` is encoded to bytes by the Processing Server, and decoded from bytes by the Processing Worker, it creates an XML tree element object, not `str`.

ocrd_network/ocrd_network/utils.py
MehmedGIT: I will have to add an extra parameter `logger` so the caller could pass their own logger type. Is that fine?

ocrd_network/ocrd_network/utils.py
MehmedGIT: Good point. I could further simplify both functions.

> Also why is the Content-Type necessary for the host:port variant?

Maybe it is not necessary, I will remove it.

MehmedGIT:
Okay, seems when there are a lot of processing requests in a page-wise manner the locking/unlocking of output file group pages is sometimes inconsistent. Consider the example below:
```bash
DEBUG:ocrd_network.processing_server:Result job_id: b1673bdc-9bef-4017-b3a7-badde7743835, state: SUCCESS
DEBUG:ocrd_network.processing_server:Result job_id: 55961039-7e75-449b-8585-bbeb06b36171, state: SUCCESS
DEBUG:ocrd_network.processing_server:Unlocking pages of `OCR-D-SEG-PAGE-ANYOCR`: ['PHYS_0010']
DEBUG:ocrd_network.processing_server:Remaining locked pages of `OCR-D-SEG-PAGE-ANYOCR`: ['PHYS_0009']
DEBUG:ocrd_network.processing_server:Unlocking pages of `OCR-D-SEG-PAGE-ANYOCR`: ['PHYS_0009']
DEBUG:ocrd_network.processing_server:Remaining locked pages of `OCR-D-SEG-PAGE-ANYOCR`: ['PHYS_0010']

# PHYS_0010 remains still locked in the DB, although it is not anymore.
```

Simply locking/unlocking the DB access did not help. Will investigate more.

ocrd_network/ocrd_network/utils.py
kba: This is strange and surprising. Can you share the workspace where this happens? We do explicitly test this behavior in the `tests/model/test_ocrd_mets.py`:

```python
def test_physical_pages(sbb_sample_01):                               
    assert len(sbb_sample_01.physical_pages) == 3, '3 physical pages' 
    assert isinstance(sbb_sample_01.physical_pages, list)             
    assert isinstance(sbb_sample_01.physical_pages[0], str)           
```

If the return is not a `List[str]`, this is a bug and needs fixing.

ocrd_network/ocrd_network/utils.py
kba: Sure, but you could also just do `getLogger(logger_name)`, whatever works best in your opinion.

ocrd_network/ocrd_network/utils.py
MehmedGIT: It happens with any workspace I have tried. Check it for yourself, by running:

```python
from ocrd import Resolver, Workspace

METS_PATH = '/path/to/mets'
physical_pages = Resolver().workspace_from_url(mets_url=METS_PATH).mets.physical_pages

print(physical_pages[0])
print(type(physical_pages[0]))
print(type(f'{physical_pages[0]}'))
```

Which prints:
```bash
PHYS_0001
<class 'lxml.etree._ElementUnicodeResult'>
<class 'str'>
```

MehmedGIT: I think this happens because `lxml.etree._ElementUnicoreResults` extends the `str` which the `isinstance` sees as a string, but it is not.

MehmedGIT:
Okay, the DB is no longer utilized for page locking/unlocking. I have created another internal variable to keep track of locked/unlocked pages for each workspace. This fixes the issue above. We should consider creating a separate data class to handle the low-level data access to make the code more readable.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -106,6 +106,12 @@ def __init__(self, config_path: str, host: str, port: int) -> None:
         # but no internal callback was yet invoked
         self.processing_counter_cache = {}
 
+        # Used for keeping track of locked pages for a workspace
+        # Key: `workspace_id` or `path_to_mets` depending on which is provided
```
kba, line 110: Is that still accurate? Because you later changed

```
- workspace_key = data.workspace_id if data.workspace_id else data.path_to_mets
+ workspace_key = data.path_to_mets
```

ocrd_network/ocrd_network/utils.py
kba: I see, thanks, that explains it, I'll prepare a quick fix for that.

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: Good catch! The key is currently just the `path_to_mets`. I have added a TODO to fix it when refactoring. A more proper resolution of `workspace_id` to `path_to_mets` is required.

ocrd_network/ocrd_network/utils.py
kba: https://github.com/OCR-D/core/pull/1102

MehmedGIT: Approved.

MehmedGIT:
Okay, the issues seem resolved now. The issues were:
1) Not calling the QoS method in the wrapper RabbitMQ library when connecting a consumer. This led to consumers consuming a lot of requests from the queue instead of 1 per time.
2) Iterating over a list and trying to remove elements in the same iteration - a dummy mistake to be made.
3) Relying on processing request indices when caching instead of actual processing requests. This led to the cancellation of jobs that were not supposed to be canceled. 

To confirm, I have tested with a workspace of 140 pages. First, I ran with 2 instances per OCR-D processor. Then 4 instances per OCR-D processor. In both executions, the same results were obtained. As well as the complete `OCR-D-OCR` file group.

MehmedGIT:
Things to be considered yet:
1) Ease to the user for checking job statuses, i.e., currently the processing server supports `GET /processor/{processor_name}/{job_id}` for checking the status of the processing job. However, this is inconvenient for the general user. The user should be able to check the processing step job status without knowing the name of the processor. The `job_id` is a globally unique id. 
Suggestion: `GET /processor/{processor_name}/{job_id}` -> `GET /processor/job_status/{job_id}`
2) Since there is no way to check the workflow job status, a separate endpoint is needed. In a similar manner to checking the processing job status. There is already a DB and Response model for the Workflow.
Suggestion: provide an endpoint `GET /workflow/job_status/{job_id}`
3) Find an optimal way to update the job status of a workflow based on the job statuses of the building processing steps.

ocrd_network/ocrd_network/processing_worker.py
MehmedGIT: This is now resolved with #1105.

MehmedGIT:
> For the page-wise processing, I was also considering a way to reduce the amount of total requests. With a workflow that utilizes 9 processors and a workspace of 100 pages that makes 900 requests per workflow... We could also provide something like `page_step`. Instead of creating a one-page request, the `page_step` amount of pages per request will be created. This adds more complexity for testing but is also more flexible and convenient for the user. For now, I would just keep it as an idea - till we have everything tested and working.
> 
> EDIT: Feature postponed since it will bring more complexity and further increase the required timeout value for the METS Server.

Adding as an extra note so it does not disappear somewhere above.

bertsky:
We still need a [network client CLI](https://github.com/OCR-D/core/issues/1032#issuecomment-1511482113) for the new workflow endpoint. Should we track this in a new issue?

bertsky:
> Anything you see in the [spec](https://github.com/OCR-D/spec/blob/eb51dfaee54313826404ebdcb60425c4682d8b9b/openapi.yml) should be up-to-date and matching with core v2.67.2. If that is not the case, let me know.

oops, sry, I did not look at master – everything is in order between spec and core, as you stated.

> Regarding the network client CLI - that idea was left behind at some point and needs to be implemented. A separate issue makes sense.

https://github.com/OCR-D/core/issues/1264

Issue #1085, opened 2023-08-31
`ocrd network processing-worker` can only invoke cli


joschrew:
For starting workers with docker-compose I use commands like this: 
`command: ocrd network processing-worker --database $MONGODB_URL --queue $RABBITMQ_URL --create-queue --queue-connect-attempts 5 ocrd-anybaseocr-crop`.
That creates a `ProcessingWorker` class with `processor_class` set to `None`. This finally leads to call `run_cli` in process_helpers.py.invoke_processor().
Maybe it is possible to provide the **class** (string with class-name is not sufficient) somehow with the `ocrd network processing-worker` command.

kba:
Recap of the discussion yesterday:

  - We do need a generic solution to add the worker CLI to the processor CLI because that is the only way we can reasonably access the processor class which allows for caching etc.
  - The processor API should be extended in such a way that the worker CLI is programmatically added without requiring changes to the processors themselves
  - If that is not possible without hacking click's internals, then we need to adapt all the processors to add the additional CLI themselves.

kba:
There is no way for `ocrd network processing-worker` to be able to reliably deduce the python class of a processor just from the CLI name. For pythonic processors, `ocrd-* --agent-type {worker,server}` (or `ocrd-* [server|worker]` is the way to go, for bashlib processors we need `ocrd network processing-worker` is the only way.

PR #1087, opened 2023-09-06
replace --agent-type with subcommand, #1032, #1085

kba:
- Instead of an option `--agent-type`, implement the behavior as subcommands, i.e.

```
ocrd-foo --agent-type worker ...
```

becomes

```
ocrd-foo worker ...
```

Also removes the exception handling because it does not actually handle the exceptions, just stringifies them and we want full stack traces if something goes wrong.

### DIFF
--- a/ocrd/ocrd/cli/ocrd_tool.py
+++ b/ocrd/ocrd/cli/ocrd_tool.py
@@ -6,6 +6,7 @@ OCR-D CLI: ocrd-tool.json management
     :nested: full
 
 """
+from inspect import getmodule
 from json import dumps
 import codecs
 import sys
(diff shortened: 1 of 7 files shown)


MehmedGIT:
Initially, I did not see a reason why we would need that change. However, now when I think about it, this refines the way we handle the `worker` and `server` subcommands. The `--queue` option is exclusive for the `worker` and the `--address` option is exclusive for the `server`.

So `ocrd-* worker --help` and `ocrd-* server --help` could have a more meaningful output for the general user. 

Currently, `ocrd-* --help` is still lacking the `--address` flag and the additional processing worker flags (`--create-queue` and `--queue-connect-attempts`) added in #1075.

```
Options for Processing Worker server:
  --queue                         The RabbitMQ server address in format
                                  "amqp://{user}:{pass}@{host}:{port}/{vhost}"
                                  [amqp://admin:admin@localhost:5672]
  --database                      The MongoDB server address in format
                                  "mongodb://{host}:{port}"
                                  [mongodb://localhost:27018]
  --type                          type of processing: either "worker" or "server"

```

MehmedGIT:
I could already push here the adaptation changes needed in the `ocrd network` to use the subcommands.

MehmedGIT:
The bash library should still be adapted appropriately. Currently, it still accepts `--type`.

kba:
> So `ocrd-* worker --help` and `ocrd-* server --help` could have a more meaningful output for the general user.

OK, so how about three `--help` variants. One for the general CLI which lists the `worker`/`server` subcommands but omits the sections `Options for Processing Worker:` and `Options for Processor Server:`. Then one for `ocrd-* worker --help` which lists only those options and one for `ocrd-* server` which lists those options plus the general `Options for processing`?

MehmedGIT:
That would be even better!

kba:
> now to fix the errors I've introduced apparently :/

Debug echo&exit was still in there 🙃 time for weekend.

MehmedGIT (approved):

PR #1089, opened 2023-09-08
raise ValueError if --page-id is provided but leads to empty result

kba:


MehmedGIT (approved):

kba:
I had to downgrade this to a warning, to not break existing code.

PR #1093, opened 2023-09-12
Make ProcessingWorker create its queue by default

joschrew:
With this changes the ProcessingWorker always creates its queue on instantiation. This is an idempotent operation. The reason to implement this is to make it possible to create and use a worker started externally via docker-compose/docker.

### DIFF
--- a/ocrd_network/ocrd_network/processing_worker.py
+++ b/ocrd_network/ocrd_network/processing_worker.py
@@ -39,6 +39,7 @@ from .utils import (
     verify_database_uri,
     verify_and_parse_mq_uri
 )
+from ocrd_utils import config
 
 
 class ProcessingWorker:
@@ -79,6 +80,11 @@ class ProcessingWorker:
         # The publisher is connected when the `result_queue` field of the OcrdProcessingMessage is set for first time
         # Used to publish OcrdResultMessage type message to the queue with name {processor_name}-result
         self.rmq_publisher = None
+        # Always create a queue (idempotent)
+        self.create_queue(
+            connection_attempts=config.OCRD_NETWORK_WORKER_QUEUE_CONNECT_ATTEMPTS,
+            retry_delay=2
+        )
 
     def connect_consumer(self) -> None:
         self.log.info(f'Connecting RMQConsumer to RabbitMQ server: '
@@ -261,7 +267,7 @@ class ProcessingWorker:
             message=encoded_result_message
         )
 
-    def create_queue(self, connection_attempts=1, retry_delay=1):
+    def create_queue(self, connection_attempts: int = 1, retry_delay: int = 1):
         """Create the queue for this worker
 
         Originally only the processing-server created the queues for the workers according to the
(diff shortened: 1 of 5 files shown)


kba (approved):
Looks good and much cleaner. Good to go AFAICT.

ocrd_utils/ocrd_utils/config.py
```diff
@@ -151,6 +151,11 @@ def _ocrd_download_timeout_parser(val):
         description="Default address of Workspace Server to connect to (for `ocrd network client workspace`).",
         default=(True, ''))
 
+config.add("OCRD_NETWORK_WORKER_QUEUE_CONNECT_ATTEMPTS",
+    description="Number of attempts for a worker to create its queue. Helpfull if the rabbitmq-server needs time to be fully started",
+    parser=int,
+    default=(True, 1))
```
MehmedGIT, line 157: The default should be 3?

ocrd_network/ocrd_network/processing_worker.py
```diff
@@ -261,7 +267,7 @@ def publish_to_result_queue(self, result_queue: str, result_message: OcrdResultM
             message=encoded_result_message
         )
 
-    def create_queue(self, connection_attempts=1, retry_delay=1):
+    def create_queue(self, connection_attempts: int = 1, retry_delay: int = 1):
```
MehmedGIT, line 270: The default for connection attempts should be `config.OCRD_NETWORK_WORKER_QUEUE_CONNECT_ATTEMPTS`. The retry delay should also be a bit higher - maybe 3 seconds?

ocrd_network/ocrd_network/processing_worker.py
```diff
@@ -79,6 +80,11 @@ def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict,
         # The publisher is connected when the `result_queue` field of the OcrdProcessingMessage is set for first time
         # Used to publish OcrdResultMessage type message to the queue with name {processor_name}-result
         self.rmq_publisher = None
+        # Always create a queue (idempotent)
+        self.create_queue(
```
MehmedGIT, line 84: We then do not need to pass parameters here but use the defaults assigned in the method signature.

MehmedGIT (changes requested):

ocrd_utils/ocrd_utils/config.py
joschrew: I thought not to wait by default only when "needed" and that could be configured via ENV

ocrd_network/ocrd_network/processing_worker.py
joschrew: I think it is better not to use the ENV as a function-parameter default value. Why do you think this is better?

MehmedGIT (approved):

ocrd_utils/ocrd_utils/config.py
joschrew: Makes sense, I committed accordingly.

ocrd_network/ocrd_network/processing_worker.py
joschrew: Makes sense, I committed accordingly.

ocrd_network/ocrd_network/processing_worker.py
joschrew: I committed accordingly.

PR #1100, opened 2023-09-15
:fire: mets server did not support local_filename and url

kba:


MehmedGIT:
Looks good

MehmedGIT (approved):
Looks good.

tests/test_mets_server.py
```diff
@@ -32,11 +32,11 @@ def _start_mets_server(*args, **kwargs):
     if exists(WORKSPACE_DIR):
         rmtree(WORKSPACE_DIR, ignore_errors=True)
 
-    copytree(assets.path_to('kant_aufklaerung_1784/data'), WORKSPACE_DIR)
+    copytree(assets.path_to('SBB0000F29300010000/data'), WORKSPACE_DIR)
     workspace = Workspace(Resolver(), WORKSPACE_DIR)
     p = Process(target=_start_mets_server, kwargs={'workspace': workspace, 'url': request.param})
     p.start()
-    sleep(2)  # sleep to start up server
+    sleep(.1)  # sleep to start up server
```
MehmedGIT, line 39: I would still keep this high enough - 1 second.

MehmedGIT (changes requested):
Since the python 3.7 test failed, I think this is because of the low sleep in the test.

tests/test_mets_server.py
kba: Indeed, that was just for testing, corrected in 97a05facf

PR #1102, opened 2023-09-20
OcrdMets.physical_pages should return "proper" str

kba:
`lxml.xpath` returns `lxml.etree._ElementUnicodeResult` which are also `str`, but not quite which led to problems in #1083.


MehmedGIT (approved):
Okay, tested and it returns proper strings now.

MehmedGIT:
Merged to #1083.

Issue #1103, opened 2023-09-26
Processor not failing when the required page id is not found


MehmedGIT:
The Processor class should raise an Exception but does not when a page id is not found. To be precise - when the whole input file group is missing from the `mets.xml`.

From the `ocrd_network` module workflow endpoint in #1083:
```python
DEBUG:ocrd_network.processing_worker:Consumer tag: ctag1.9aaac84eeb3d41e7aaef5742a39a3075, message delivery tag: 1, redelivered: False
DEBUG:ocrd_network.processing_worker:Message headers: {'OCR-D WebApi Header': 'OCR-D WebApi Value'}
DEBUG:ocrd_network.processing_worker:Trying to decode processing message with tag: 1
INFO:ocrd_network.processing_worker:Starting to process the received message: {'job_id': '2eabd3b7-e177-4953-b095-c4d5aa6b44b0', 'processor_name': 'ocrd-cis-ocropy-binarize', 'created_time': 1695708719, 'input_file_grps': ['OCR-D-IMG'], 'output_file_grps': ['OCR-D-BIN'], 'path_to_mets': '/home/mm/Desktop/ocrd_network_files/example_ws3/data/mets.xml', 'page_id': 'PHYS_0001', 'internal_callback_url': 'http://localhost:8080/result_callback', 'parameters': {}}
DEBUG:ocrd_network.processing_worker:Invoking processor: ocrd-cis-ocropy-binarize
10:11:59.267 WARNING ocrd.processor.base - Could not find any files for --page-id PHYS_0001 - compare 'PHYS_0001' with the output of 'orcd workspace list-page'.
WARNING:ocrd.processor.base:Could not find any files for --page-id PHYS_0001 - compare 'PHYS_0001' with the output of 'orcd workspace list-page'.
10:11:59.267 INFO ocrd.process.profile - Executing processor 'ocrd-cis-ocropy-binarize' took 0.003558s (wall) 0.000931s (CPU)( [--input-file-grp='OCR-D-IMG' --output-file-grp='OCR-D-BIN' --parameter='{"method": "ocropy", "threshold": 0.5, "grayscale": false, "maxskew": 0.0, "noise_maxsize": 0, "dpi": 0, "level-of-operation": "page"}' --page-id='PHYS_0001']
[...]

PR #1104, opened 2023-09-26
Fix issue #1103

MehmedGIT:
This is a fix for issue #1103. PR to #1083 since the logging in the master branch still needs rework.

### DIFF
--- a/ocrd/ocrd/processor/base.py
+++ b/ocrd/ocrd/processor/base.py
@@ -353,10 +353,9 @@ class Processor():
             # Warn if no files found but pageId was specified because that
             # might be because of invalid page_id (range)
             if self.page_id and not files_:
-                LOG.warning(f"Could not find any files for --page-id {self.page_id} - compare '{self.page_id}' with the output of 'orcd workspace list-page'.")
-                if on_error == 'abort':
-                    raise ValueError(f"Could not find any files for --page-id {self.page_id} - compare "
-                                     f"'{self.page_id}' with the output of 'orcd workspace list-page'.")
+                msg = f"Could not find any files for --page-id {self.page_id} - compare '{self.page_id}' with the output of 'orcd workspace list-page'."
+                raise ValueError(msg) if on_error == 'abort' else LOG.warning(msg)
+                    
             for file_ in files_:
                 if not file_.pageId:
                     continue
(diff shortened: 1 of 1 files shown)


ocrd/ocrd/processor/base.py
```diff
@@ -354,6 +354,9 @@ def zip_input_files(self, require_first=True, mimetype=None, on_error='skip'):
             # might be because of invalid page_id (range)
             if self.page_id and not files_:
                 LOG.warning(f"Could not find any files for --page-id {self.page_id} - compare '{self.page_id}' with the output of 'orcd workspace list-page'.")
+                if on_error == 'abort':
+                    raise ValueError(f"Could not find any files for --page-id {self.page_id} - compare "
+                                     f"'{self.page_id}' with the output of 'orcd workspace list-page'.")
```
kba, line 359:
```suggestion
                msg = f"Could not find any files for --page-id {self.page_id} - compare '{self.page_id}' with the output of 'orcd workspace list-page'."
                raise ValueError(msg) if on_error == 'abort' else LOG.warning(msg)
                    
```

kba (commented):
I forgot about the `on_error` parameter, yes, this is the correct behavior.

kba (approved):

MehmedGIT:
```python
raise ValueError(msg) if on_error == 'abort' else LOG.warning(msg)
```
is a wrong syntax since Python tries to `raise LOG.warning(msg)` when the flag is not `abort`.

ocrd/ocrd/processor/base.py
MehmedGIT:
```python
raise ValueError(msg) if on_error == 'abort' else LOG.warning(msg)
```
is a wrong syntax since Python tries to `raise LOG.warning(msg)` when the flag is not `abort`.

kba: Ah, yeah, sry, not tested, was just a proposal.

kba: > ```python
> (raise ValueError(msg)) if on_error == 'abort' else LOG.warning(msg)
> ```

might work. but better not to be too clever for my own good. I just meant that the message could be a variable to be used for both warning log and raised exception.

PR #1105, opened 2023-09-26
Workflow endpoint logging 2023 09 26

kba:
This includes #1101 and updates the default config, builtin and config file, to offer the same behavior as the hard-coded logging `ocrd_network`.

I did add a file handler for the processing serverlog, as it was before, so you get the idea on how to adapt the server cache logging as well.

I did not remove the hard-coded logging configuration in ocrd_network though, since @MehmedGIT & @joschrew know best what the exact behavior should be.

The easies way to fine-tune the logging setup is to copy `ocrd_utils/ocrd_logging.conf` to `$PWD` (or `$HOME`), then remove the coded logging setup and translate it into the config file syntax and add your changes to this PR, then we can merge it to #1083 (as this PR proposes) or into master if we decide to merge #1089 into master before finishing the logging stuff.

Ideally, we should change the root logger from `ocrd_network` to `ocrd.network`. That way, we only need one logger definition but it's not urgent.

### DIFF
--- a/ocrd/ocrd/cli/network.py
+++ b/ocrd/ocrd/cli/network.py
@@ -23,9 +23,6 @@ def network_cli():
     Managing network components
     """
     initLogging()
-    # TODO: Remove after the logging fix in core
-    logging.getLogger('paramiko.transport').setLevel(logging.INFO)
-    logging.getLogger('ocrd_network').setLevel(logging.DEBUG)
 
 
 network_cli.add_command(client_cli)
--- a/ocrd_network/ocrd_network/processing_worker.py
+++ b/ocrd_network/ocrd_network/processing_worker.py
@@ -44,14 +44,13 @@ from ocrd_utils import config
 
 class ProcessingWorker:
     def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict, processor_class=None) -> None:
-        self.log = getLogger('ocrd_network.processing_worker')
-        # TODO: Provide more flexibility for configuring file logging (i.e. via ENV variables)
-        file_handler = logging.FileHandler(f'/tmp/worker_{processor_name}_{getpid()}.log', mode='a')
-        logging_format = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
-        file_handler.setFormatter(logging.Formatter(logging_format))
+        logging_suffix = f'{processor_name}.{getpid()}'
+        self.log = getLogger(f'ocrd_network.processing_worker')
+        file_handler = logging.FileHandler(f'/tmp/worker_{logging_suffix}.log', mode='a')
+        file_handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
         file_handler.setLevel(logging.DEBUG)
         self.log.addHandler(file_handler)
-        # TODO: remove this when refactoring the logging
+        # TODO: removing this line still leads to empty output
         self.log.setLevel(logging.DEBUG)
 
         try:
--- a/ocrd_network/ocrd_network/processing_worker.py
+++ b/ocrd_network/ocrd_network/processing_worker.py
@@ -44,7 +44,7 @@ from ocrd_utils import config
 
 class ProcessingWorker:
     def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict, processor_class=None) -> None:
-        self.log = getLogger(__name__)
+        self.log = getLogger('ocrd_network.processing_worker')
         # TODO: Provide more flexibility for configuring file logging (i.e. via ENV variables)
         file_handler = logging.FileHandler(f'/tmp/worker_{processor_name}_{getpid()}.log', mode='a')
         logging_format = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
--- a/ocrd_network/ocrd_network/processing_worker.py
+++ b/ocrd_network/ocrd_network/processing_worker.py
@@ -43,11 +43,13 @@ from ocrd_utils import config
 
 
 class ProcessingWorker:
-    def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict, processor_class=None) -> None:
+    def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict, processor_class=None, log_filename:str=None) -> None:
         initLogging()
-        logging_suffix = f'{processor_name}.{getpid()}'
         self.log = getLogger(f'ocrd_network.processing_worker')
-        file_handler = logging.FileHandler(f'/tmp/ocrd_worker_{logging_suffix}.log', mode='a')
+        if not log_filename:
+            log_filename = f'/tmp/ocrd_worker_{processor_name}.{getpid()}.log'
+        self.log_filename = log_filename
+        file_handler = logging.FileHandler(log_filename, mode='a')
         file_handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
         self.log.addHandler(file_handler)
 
@@ -219,6 +221,7 @@ class ProcessingWorker:
                 input_file_grps=input_file_grps,
                 output_file_grps=output_file_grps,
                 page_id=page_id,
+                log_filename=self.log_filename,
                 parameters=processing_message.parameters,
                 mets_server_url=mets_server_url
             )
--- a/ocrd_network/ocrd_network/processing_worker.py
+++ b/ocrd_network/ocrd_network/processing_worker.py
@@ -45,7 +45,6 @@ from ocrd_utils import config
 class ProcessingWorker:
     def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict, processor_class=None) -> None:
         initLogging()
-        self.test_logging()
         logging_suffix = f'{processor_name}.{getpid()}'
         self.log = getLogger(f'ocrd_network.processing_worker')
         file_handler = logging.FileHandler(f'/tmp/ocrd_worker_{logging_suffix}.log', mode='a')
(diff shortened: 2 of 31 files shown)


kba:
Test failure is due to connection issue, tests pass.

ocrd_network/ocrd_network/processing_worker.py
```diff
@@ -44,7 +44,7 @@
 
 class ProcessingWorker:
     def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict, processor_class=None) -> None:
-        self.log = getLogger(__name__)
+        self.log = getLogger('ocrd_network.processing_worker')
         # TODO: Provide more flexibility for configuring file logging (i.e. via ENV variables)
         file_handler = logging.FileHandler(f'/tmp/worker_{processor_name}_{getpid()}.log', mode='a')
```
MehmedGIT, line 49: How could we create that in the configuration file when a `processor_name` and `pid` is involved? For other modules such as the Processing Server or Server Cache, this is not an issue and the pattern with the configuration file could be followed. However, for the dynamic creation of log files, it will be a pain to manually put things in the log config file.

ocrd_network/ocrd_network/rabbitmq_utils/consumer.py
```diff
@@ -25,9 +24,6 @@ def __init__(self, host: str = HOST, port: int = PORT, vhost: str = VHOST,
         if not logger_name:
             logger_name = __name__
         logger = logging.getLogger(logger_name)
```
MehmedGIT, line 26:
```suggestion
        logger = getLogger('ocrd_network.rabbitmq_utils.consumer')
```

I should take care of that and test if the logging happens properly though - since the `logger_name` comes from the constructor. However, we could and should remove the `logger_name` parameter as well, since we have a separate file logging for different modules now.

ocrd_network/ocrd_network/rabbitmq_utils/publisher.py
```diff
@@ -30,9 +28,6 @@ def __init__(self, host: str = HOST, port: int = PORT, vhost: str = VHOST,
         if logger_name is None:
             logger_name = __name__
         logger = logging.getLogger(logger_name)
```
MehmedGIT, line 30: The same as in the `consumer.py` case. This reminds me that the same change should happen in the `connector.py` as well when the `logger_name` parameter is removed.

ocrd_utils/ocrd_logging.conf
```diff
@@ -54,28 +54,23 @@ handlers=consoleHandler
...
-handlers=consoleHandler
-qualname=ocrd_utils
-[logger_ocrd_ocrd_network]
-level=INFO
-handlers=consoleHandler
-qualname=ocrd_network
-[logger_ocrd_ocrd_exif]
-level=INFO
-handlers=consoleHandler
-qualname=ocrd_exif
```
MehmedGIT, line 77: Why is `ocrd_models` still involved? Isn't this already handled by `logger_ocrd` above?

ocrd_utils/ocrd_utils/logging.py
```diff
@@ -46,19 +46,23 @@
...
+    'ocrd_network': logging.DEBUG,
     # 'ocrd.resolver': logging.INFO,
     # 'ocrd.resolver.download_to_directory': logging.INFO,
     # 'ocrd.resolver.add_files_to_mets': logging.INFO,
     # To cut back on the `Self-intersection at or near point` INFO messages
     'shapely.geos': logging.ERROR,
     'tensorflow': logging.ERROR,
     'PIL': logging.INFO,
+    'paramiko.transport': logging.INFO
 }
```
MehmedGIT, line 66: The FastAPI and Uvicorn loggers are not reflected here. This leads to their log suppression - I no longer see output from them.

I think that should work and be enough:
```python
'uvicorn.access': logging.DEBUG,
'uvicorn.error': logging.DEBUG,
'uvicorn': logging.INFO
```
Have a look here: https://github.com/tiangolo/fastapi/issues/1508

MehmedGIT (changes requested):
When I ran a workflow to check if everything was logged properly I saw that only 2 main things were missing:
1) The FastAPI and Uvicorn logging - ~already suggested a solution~ suggestion not working.
2) The logging that comes from the OCR-D processors is suppressed. I am not sure about the fix. This is the logging that is redirected by the Deployer with the `xyz` hack to get the `pid`. You can check the log files with the pattern `/tmp/deployed_*` to see they have just 2 lines each - the initialization part.

ocrd/ocrd/cli/network.py
```diff
@@ -23,9 +23,6 @@ def network_cli():
     Managing network components
     """
     initLogging()
```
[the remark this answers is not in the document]
kba: They are required for the CLI because without it, no logging is set up.

The idea is to only use `initLogging` if we're in "application mode" not in "library mode" where there might be logging set up in a different way. E.g. in ODEM, ocrd_kitodo or OPERANDI, there likely is a separate logging setup, with which we don't want to interfere.

ocrd_network/ocrd_network/processing_worker.py
kba: You don't have to put every logger you're using in the config file, only in cases where the logger's level deviates from the ancestor in the logger hierarchy.

```
getLogger(f'ocrd_network.worker.{processor_name}.{pid}')
```

As it is currently implemented, this would mean that it should inherit the `DEBUG` level from

```conf
[logger_ocrd_network]                          
level=DEBUG                                    
qualname=ocrd_network                          
```

But you could override the `ocrd_network.worker` logger in the config file to change the log level to something else.

ocrd_network/ocrd_network/rabbitmq_utils/consumer.py
kba: Agree, feel free to commit this and similar changes to this branch.

ocrd_network/ocrd_network/processing_worker.py
MehmedGIT: Got it, that makes sense, yes.

ocrd_utils/ocrd_logging.conf
kba: It was to be removed in the the part of #1101 that I had to abandon because of the bizarre test failures. In fact that was the point of that PR, to make all the loggers consistently inherit from `ocrd`. I will still do that, but let's first finish this and #1083.

ocrd_utils/ocrd_utils/logging.py
kba: I only copied those from the hard-coded definition in `ocrd/ocrd/cli/network.py`, feel free to adapt the way you see fit.

MehmedGIT: That suggestion did not fix it.

kba: And you're using the `ocrd_logging.conf`? These must be set in the config file as well, the `LOGGING_DEFAULTS` are only set for the builtin logging (i.e. no ocrd_logging.conf).

If that does not help, it might be dependant on when those loggers are first accessed. Can you try

```python.diff
-    if config_file:                                                            
-        logging.config.fileConfig(config_file) 
+    if config_file:                                                            
+        logging.config.fileConfig(config_file, disable_existing_loggers=False) 
```

in `ocrd_utils/ocrd_utils/logging.py`?

MehmedGIT: > And you're using the `ocrd_logging.conf`? These must be set in the config file as well, the `LOGGING_DEFAULTS` are only set for the builtin logging (i.e. no ocrd_logging.conf).

Good point. I indeed use a logging config and they were not set there. Will test again.

MehmedGIT: Resolved in eebc5cda683e9c41a76c27cb9940a95b96113891.

MehmedGIT:
What is still missing/not working properly regarding the logging:
1) The `Processor Servers` cannot log to a file (under `/tmp/ocrd_server_{processor_name}.{pid}.log`) both their and the output of the specific OCR-D Processor is missing. Note that the `Processing Server` which has similar configurations to the `Processor Server` can in fact log its output to a file (under `/tmp/ocrd_processing_server.log`). The separate log files for the caching mechanisms are also fine (under `/tmp/ocrd_processing_server_cache_locked_pages.log` and `/tmp/ocrd_processing_server_cache_processing_requests.log`).
2) The `Processing Workers` can log their output (under `/tmp/ocrd_worker_{processor_name}.{pid}.log`) regarding the RabbitMQ message consumption and final job status of the triggered OCR-D processor. 
3) The Deployer module, when deploying the `Processing Workers` redirects the deployed process/es stdout matching the `processor_name` to the `/tmp/deployed_{processor_name}.log` file. That log file contains both the `Processing Workers` output and the separate processing steps of OCR-D processors output.

kba:
> * It would be easier to wrap our heads around if the `run_processor` and `run_cli` methods provided a way to redirect their stdout/stderr to a file path.

For `run_cli`, it is easy to do and makes sense.

For `run_processor`, it's more complicated. Since these are called in the same process, redirecting STDOUT7STDERR is tricky. It is certainly possible to mess with `sys.stdout` but there is a mechanism specifically for this - logging.

How about we add a processor-wide `log` attribute that holds the logger the processor uses in `process`, i.e. have

```
def __init__(...):
  self.log = getLogger('ocrd.ocrd-tesserocr-recognize')
```

Once we have instantiated the processor, we can add a `logging.FileHandler` to `processor.log` which points to a provided file path.

MehmedGIT:
> Once we have instantiated the processor, we can add a logging.FileHandler to processor.log which points to a provided file path.

Not sure if that would be enough for what we are trying to achieve. Sure adding a `logging.FileHandler` will capture everything related to that logger but there will be no differentiation among multiple initializations of that specific processor. 3 Processing Workers initializing the same OCR-D processor will end up logging to the same file since the logger name will be the same. This will not be the case if the `logging.FileHandler` is initialized inside the `run_processor` method. However, even then we still do not fully get what we are trying to achieve. Attaching the file handler to that specific logger will omit other relevant logging output from other loggers say `ocrd.workspace`. Is that what we want to achieve though?

MehmedGIT:
After #1109, now it is clear how we can achieve processing job level file logging.

ocrd_utils/ocrd_logging.conf
kba: `ocrd_models` is gone now.

PR #1107, opened 2023-09-28
generate_page_range: warn, not raise, if start==end, fix #1106

### LINKED ISSUE #1106: Allow single page ranges
Page ranges of the kind `PHYS_0001..PHYS_0001` currently raises an error, which interferes with testing #1083 because this is now a valid use case - testing a single-page workspace. Such a range should be valid and at most there should be a warning only.

kba:


tests/test_utils.py
```diff
@@ -317,7 +317,7 @@ def test_generate_range(self):
         assert generate_range('PHYS_0001', 'PHYS_0005') == ['PHYS_0001', 'PHYS_0002', 'PHYS_0003', 'PHYS_0004', 'PHYS_0005']
         with self.assertRaisesRegex(ValueError, 'could not find numeric part'):
             generate_range('NONUMBER', 'ALSO_NONUMBER')
-        with self.assertRaisesRegex(ValueError, 'evaluates to the same number'):
+        with warns(UserWarning, match='same number'):
             generate_range('PHYS_0001_123', 'PHYS_0010_123')
```
MehmedGIT, line 321: Hmm, how does this evaluate on the same page? The page_id here is just `123` for both the start and end?

MehmedGIT (commented):
Seems to fix the issue on my side. I am just not sure whether having a warning for each processing request with a single page will not be too much but that is a small detail. So good to be merged!

tests/test_utils.py
kba: Because the assumption is that the variable number is the last one in the string. For example, Deutsches Textarchiv has the naming convention `<author>_<work>_<year>`, e.g. `kant_aufklaerung_1784` and a possible page ID naming scheme could be `kant_aufklaerung_1784_0001` - in this case it would be wrong to iterate over the `1784` part.

In this case this evaluates to `PHYS_0001_123`, because we take the start id as the basis.

kba:
> I am just not sure whether having a warning for each processing request with a single page will not be too much but that is a small detail.

We can either redirect warnings to logging https://docs.python.org/3/library/logging.html#logging.captureWarnings and handle/hide them that way.

There's also methods to filter warnings directly, e.g. https://docs.python.org/3/library/warnings.html#the-warnings-filter

PR #1108, opened 2023-10-03
Additions to the workflow-endpoint

joschrew:
This contains a new endpoint to query the status of a workflow run and improved error handling for the workflow-enpoint. This were some of the parts we realized in our last meeting we need improvements in.

### DIFF
--- a/ocrd_network/ocrd_network/processing_server.py
+++ b/ocrd_network/ocrd_network/processing_server.py
@@ -758,6 +758,8 @@ class ProcessingServer(FastAPI):
             workflow_callback_url=workflow_callback_url
         )
         await db_workflow_job.insert()
+        # Do not return the processing_job_ids to the user
+        db_workflow_job[processing_job_ids] = {}
         return db_workflow_job.to_job_output()
 
     async def get_workflow_info(self, workflow_job_id) -> Dict:
--- a/ocrd_network/ocrd_network/processing_server.py
+++ b/ocrd_network/ocrd_network/processing_server.py
@@ -183,6 +183,11 @@ class ProcessingServer(FastAPI):
             tags=['workflow', 'processing'],
             status_code=status.HTTP_200_OK,
             summary='Run a workflow',
+            response_model=PYWorkflowJobOutput,
+            response_model_exclude=["processing_job_ids"],
+            response_model_exclude_defaults=True,
+            response_model_exclude_unset=True,
+            response_model_exclude_none=True
         )
 
         self.router.add_api_route(
@@ -758,8 +763,6 @@ class ProcessingServer(FastAPI):
             workflow_callback_url=workflow_callback_url
         )
         await db_workflow_job.insert()
-        # Do not return the processing_job_ids to the user
-        db_workflow_job[processing_job_ids] = {}
         return db_workflow_job.to_job_output()
 
     async def get_workflow_info(self, workflow_job_id) -> Dict:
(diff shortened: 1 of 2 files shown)


MehmedGIT (approved):
Works just great on my end! Could and should be merged to master together with #1083 and #1105 in the next core release.

MehmedGIT:
Okay, I thought of a few other small suggestions. Since there is an endpoint for checking the workflow job status, potentially we could further improve by:

1) The user is no longer interested in getting a big list with all processing job IDs when a workflow is triggered. 
Suggestion: Simplify the response to contain just the workflow job ID.

2) When a few pages fail in the workflow, currently there is no way for the user to know exactly which processing tasks, i.e. pages, have failed. 
Suggestion: When returning the dictionary with states, in case some page ids have a state "FAILED", then a more detailed list inside the response dictionary should be filled with all the `DBProcessorJob` parameters for the failed tasks.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -735,3 +759,21 @@ async def run_workflow(
         )
         await db_workflow_job.insert()
         return db_workflow_job.to_job_output()
```
MehmedGIT, line 761: For suggestion 1:

```suggestion
        await db_workflow_job.insert()
        # Do not return the processing_job_ids to the user
        db_workflow_job[processing_job_ids] = {}
        return db_workflow_job.to_job_output()
```

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -735,3 +759,21 @@ async def run_workflow(
...
+                                detail=f"Workflow-Job with id: {workflow_job_id} not found")
+        job_ids: List[str] = [id for lst in workflow_job.processing_job_ids.values() for id in lst]
+        jobs = await db_get_processing_jobs(job_ids)
+        res = {}
+        for job in jobs:
+            res.setdefault(job.processor_name, {})
+            res[job.processor_name].setdefault(job.state.value, 0)
+            state_count = res[job.processor_name][job.state.value]
+            res[job.processor_name][job.state.value] = state_count + 1
+        return res
```
MehmedGIT, line 779:
```suggestion
        res = {}
        failed_tasks = {}
        for job in jobs:
            res.setdefault(job.processor_name, {})
            res[job.processor_name].setdefault(job.state.value, 0)
            state_count = res[job.processor_name][job.state.value]
            if job.state.value == "FAILED":
                failed_tasks[job.id] = job
            res[job.processor_name][job.state.value] = state_count + 1
        res[failed_tasks] = failed_tasks
        return res
```

Just to give you an idea - I have not tested it!

ocrd_network/ocrd_network/processing_server.py
joschrew: I committed first and will check later/now, and I am going to resolve conversation then.

ocrd_network/ocrd_network/processing_server.py
joschrew: same as above

ocrd_network/ocrd_network/processing_server.py
joschrew: I have modified the code. Same outcome, different approach, I just excluded the the property "processing_job_ids" from the output via a Flag. See: https://github.com/OCR-D/core/pull/1108/commits

MehmedGIT (approved):

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -735,3 +764,27 @@ async def run_workflow(
...
+        failed_tasks_key = "failed-processor-tasks"
+        for job in jobs:
+            res.setdefault(job.processor_name, {})
+            res[job.processor_name].setdefault(job.state.value, 0)
+            res[job.processor_name][job.state.value] += 1
+            if job.state == "FAILED":
+                if failed_tasks_key not in res:
+                    res[failed_tasks_key] = failed_tasks
+                failed_tasks.setdefault(job.processor_name, [])
+                failed_tasks[job.processor_name].append(job.job_id)
```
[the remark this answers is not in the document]
joschrew: Ok, I didn't realize that you wanted it that way. I think then it would be better if we e.g. I at least add the page_id for now so that it is easier to adapt later and that it is more obvious that more fields would potentially be available.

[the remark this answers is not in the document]
joschrew: I have now added the page_id. I agree that the logfile-path would be helpful. We should add that sometime when it is possible.

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: Yes, that's better.

kba (approved):
LGTM. We should continue the discussion about what best to return to users in a separate issue after release.

PR #1111, opened 2023-10-11
Improve ocrd_network logging

MehmedGIT:
This PR improves the logging mechanism of the `ocrd_network` module.

Additions:
- A separate logging dir tree structure for the modules (processing servers, processing workers, processor servers, mets servers, processing jobs). Configurable with an env variable.
- Processing job-level logging - each job is logged into a separate file with format `{job_id}.log`
- Processing job-level logging file paths are added to the Job models and preserved in the database.
- The `ocrd_network` logging is based on the format provided in `ocrd_utils`
- Support env variables for setting the root directory for logging/sockets
- An endpoint for getting the log file of a processing job of a processor

Feel free to recommend suggestions.

### DIFF
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -9,7 +9,6 @@ Each Processing Worker is an instance of an OCR-D processor.
 from __future__ import annotations
 from typing import Dict, List, Union
 from re import search as re_search
-from os import getpid
 from pathlib import Path
 import subprocess
 from time import sleep
@@ -22,7 +21,7 @@ from .deployment_utils import (
     verify_mongodb_available,
     verify_rabbitmq_available,
 )
-
+from .logging import get_mets_server_logging_file_path
 from .runtime_data import (
     DataHost,
     DataMongoDB,
@@ -468,18 +467,15 @@ class Deployer:
         self.log.info(f'Starting native processing worker: {processor_name}')
         channel = ssh_client.invoke_shell()
         stdin, stdout = channel.makefile('wb'), channel.makefile('rb')
-        cmd = f'{processor_name} worker --database {database_url} --queue {queue_url}'
+        cmd = f'{processor_name} worker --database {database_url} --queue {queue_url} &'
         # the only way (I could find) to make it work to start a process in the background and
         # return early is this construction. The pid of the last started background process is
         # printed with `echo $!` but it is printed inbetween other output. Because of that I added
         # `xyz` before and after the code to easily be able to filter out the pid via regex when
         # returning from the function
 
-        # TODO: Check here again
-        # log_path = f'/tmp/deployed_{processor_name}.log'
-        # stdin.write(f"echo starting processing worker with '{cmd}' >> '{log_path}'\n")
-        # stdin.write(f'{cmd} >> {log_path} 2>&1 &\n')
-        stdin.write(f'{cmd} &\n')
+        self.log.debug(f'About to execute command: {cmd}')
+        stdin.write(f'{cmd}\n')
         stdin.write('echo xyz$!xyz \n exit \n')
         output = stdout.read().decode('utf-8')
         stdout.close()
@@ -514,9 +510,9 @@ class Deployer:
         self.log.info(f"Starting native processor server: {processor_name} on {agent_address}")
         channel = ssh_client.invoke_shell()
         stdin, stdout = channel.makefile('wb'), channel.makefile('rb')
-        cmd = f'{processor_name} server --address {agent_address} --database {database_url}'
-        stdin.write(f"echo starting processor server with '{cmd}'\n")
-        stdin.write(f'{cmd} &\n')
+        cmd = f'{processor_name} server --address {agent_address} --database {database_url} &'
+        self.log.debug(f'About to execute command: {cmd}')
+        stdin.write(f'{cmd}\n')
         stdin.write('echo xyz$!xyz \n exit \n')
         output = stdout.read().decode('utf-8')
         stdout.close()
@@ -525,9 +521,8 @@ class Deployer:
 
     # TODO: No support for TCP version yet
     def start_unix_mets_server(self, mets_path: str) -> str:
-        socket_file = f'{safe_filename(mets_path)}.sock'
-        log_path = f'/tmp/{safe_filename(mets_path)}.log'
-        mets_server_url = f'/tmp/{socket_file}'
+        log_file = get_mets_server_logging_file_path(mets_path=mets_path)
+        mets_server_url = f'/tmp/{safe_filename(mets_path)}.sock'
 
         if is_mets_server_running(mets_server_url=mets_server_url):
             self.log.info(f"The mets server is already started: {mets_server_url}")
@@ -539,8 +534,8 @@ class Deployer:
             args=['nohup', 'ocrd', 'workspace', '--mets-server-url', f'{mets_server_url}',
                   '-d', f'{cwd}', 'server', 'start'],
             shell=False,
-            stdout=open(log_path, 'w'),
-            stderr=open(log_path, 'a'),
+            stdout=open(file=log_file, mode='w'),
+            stderr=open(file=log_file, mode='w'),
             cwd=cwd,
             universal_newlines=True
         )
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -12,6 +12,7 @@ from re import search as re_search
 from pathlib import Path
 import subprocess
 from time import sleep
+from tempfile import gettempdir
 
 from ocrd_utils import getLogger, safe_filename
 
@@ -522,7 +523,7 @@ class Deployer:
     # TODO: No support for TCP version yet
     def start_unix_mets_server(self, mets_path: str) -> str:
         log_file = get_mets_server_logging_file_path(mets_path=mets_path)
-        mets_server_url = f'/tmp/{safe_filename(mets_path)}.sock'
+        mets_server_url = f'{gettempdir()}/{safe_filename(mets_path)}.sock'
 
         if is_mets_server_running(mets_server_url=mets_server_url):
             self.log.info(f"The mets server is already started: {mets_server_url}")
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -535,7 +535,7 @@ class Deployer:
                   '-d', f'{cwd}', 'server', 'start'],
             shell=False,
             stdout=open(file=log_file, mode='w'),
-            stderr=open(file=log_file, mode='w'),
+            stderr=open(file=log_file, mode='a'),
             cwd=cwd,
             universal_newlines=True
         )
--- a/ocrd_network/ocrd_network/deployer.py
+++ b/ocrd_network/ocrd_network/deployer.py
@@ -12,9 +12,8 @@ from re import search as re_search
 from pathlib import Path
 import subprocess
 from time import sleep
-from tempfile import gettempdir
 
-from ocrd_utils import getLogger, safe_filename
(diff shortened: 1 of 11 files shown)


ocrd_network/ocrd_network/logging.py
```diff
@@ -0,0 +1,48 @@
...
+    "processor_servers"
+]
+
+
+def get_root_logging_dir(module_name: str) -> str:
+    if module_name not in OCRD_NETWORK_MODULES:
+        raise ValueError(f"Invalid module name: {module_name}, should be one of: {OCRD_NETWORK_MODULES}")
+    # TODO: Utilize env variable to set the root
+    module_log_dir = join("/tmp/ocrd_network_logs", module_name)
+    makedirs(name=module_log_dir, exist_ok=True)
```
kba, line 19: I would use [pathlib](https://docs.python.org/3/library/pathlib.html) for all new fs-related code.
```suggestion
    module_log_dir = Path(tempfile.gettempdir(), "ocrd_network_logs", module_name)
    module_log_dir.mkdir(exist_ok=True)
```

ocrd_network/ocrd_network/logging.py
```diff
@@ -0,0 +1,48 @@
...
+    return module_log_dir
+
+
+def get_cache_locked_pages_logging_file_path() -> str:
+    return join(get_root_logging_dir("processing_servers"), f"cache_locked_pages.log")
```
kba, line 24:
```suggestion
    return get_root_logging_dir("processing_servers") / "cache_locked_pages.log"
```

Also for the other calls.

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -82,6 +85,11 @@ def __init__(self, config_path: str, host: str, port: int) -> None:
             description='OCR-D Processing Server'
         )
         self.log = getLogger('ocrd_network.processing_server')
+        log_file = get_processing_server_logging_file_path(pid=getpid())
+        file_handler = logging.FileHandler(filename=log_file, mode='a')
+        file_handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
```
kba, line 90:
```suggestion
        from ocrd_utils import LOG_FORMAT
        file_handler.setFormatter(logging.Formatter(LOG_FORMAT))
```

kba (commented):
LGTM, just a quick read-through, I'll contribute ideas once I have some more time for it.

ocrd_network/ocrd_network/deployer.py
```diff
@@ -539,8 +534,8 @@ def start_unix_mets_server(self, mets_path: str) -> str:
             args=['nohup', 'ocrd', 'workspace', '--mets-server-url', f'{mets_server_url}',
                   '-d', f'{cwd}', 'server', 'start'],
             shell=False,
-            stdout=open(log_path, 'w'),
-            stderr=open(log_path, 'a'),
+            stdout=open(file=log_file, mode='w'),
+            stderr=open(file=log_file, mode='w'),
```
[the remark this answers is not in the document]
MehmedGIT: Yes, my bad. We have the same `log_file` there. Fixed.

ocrd_network/ocrd_network/deployer.py
```diff
@@ -525,9 +521,8 @@ def start_native_processor_server(
 
     # TODO: No support for TCP version yet
     def start_unix_mets_server(self, mets_path: str) -> str:
-        socket_file = f'{safe_filename(mets_path)}.sock'
-        log_path = f'/tmp/{safe_filename(mets_path)}.log'
-        mets_server_url = f'/tmp/{socket_file}'
+        log_file = get_mets_server_logging_file_path(mets_path=mets_path)
+        mets_server_url = f'/tmp/{safe_filename(mets_path)}.sock'
```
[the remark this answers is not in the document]
MehmedGIT: Good point.

ocrd_network/ocrd_network/logging.py
kba: This does the right concatenation by platform, also Windows. The `/` operator is just overloaded to allow writing `Path('a') / 'b' instead of `Path('a', 'b'), it is semantically equivalent to `join('a', 'b')` and under the hood, it uses `os.join` IIRC. cf https://docs.python.org/3/library/pathlib.html#operators

MehmedGIT: Did not know that. Thanks!

ocrd_network/ocrd_network/logging.py
MehmedGIT: Resolved and using Path, however, I am still not using `tempfile.gettempdir()`. For the logging root, I don't think it's a good idea to use a temporary directory. Especially, for the dockerized version where volume mapping should be persistent in the docker compose file. Let me know if you have a better idea.

ocrd_network/ocrd_network/deployer.py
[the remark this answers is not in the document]
joschrew: I think we should not use `tempfile.gettempdir()` here. The reason tmp is used is that it allows root and the "normal" user (i.e. 1000) to write and read here. The processing server and the workers (possible deployed in docker and /tmp is a volume mounted dir from the host) must be able to read and write here. I don't see any point in using tempfile. Tbh I would prefer this to be env-configurable and I would rather like to see it in /data instead of /tmp.

ocrd_network/ocrd_network/logging.py
```diff
@@ -0,0 +1,48 @@
...
+    "processing_servers",
+    "processing_workers",
+    "processor_servers"
+]
+
+
+def get_root_logging_dir(module_name: str) -> Path:
+    if module_name not in OCRD_NETWORK_MODULES:
+        raise ValueError(f"Invalid module name: {module_name}, should be one of: {OCRD_NETWORK_MODULES}")
+    # TODO: Utilize env variable to set the root
```
joschrew, line 17: I thought about that too, not only for logging but for the mets server too. I think this could be part of the PR

joschrew, line 17: I thought about that too, not only for logging but for the mets server too. I think this could be part of the PR

ocrd_network/ocrd_network/logging.py
```diff
@@ -0,0 +1,48 @@
...
+    module_log_dir = Path("/tmp", "ocrd_network_logs", module_name)
```
joschrew, line 18: I thought about that too, not only for logging but for the mets server too. I think this could be part of the PR

joschrew, line 18: I thought about that too, not only for logging but for the mets server too. I think this could be part of the PR

ocrd_network/ocrd_network/logging.py
```diff
@@ -0,0 +1,48 @@
...
+    module_log_dir.mkdir(parents=True, exist_ok=True)
+    return module_log_dir
+
+
+def get_cache_locked_pages_logging_file_path() -> Path:
+    return get_root_logging_dir("processing_servers") / "cache_locked_pages.log"
```
joschrew, line 24: I think this should use the constants defined instead of hard coded. I know this is checked in `get_root_logging_dir` but still. Other methods as well.

joschrew (commented):
I have tested and it seems to work fine. I think it is a huge improvement, all structured now and much better to find the right logfile now.
What imho could be added is: 
- make logfile and and mets-server basedir configurable with env. For example when I test locally I use /data for the mets-server (this is better for my local setup with docker-compose) and when this was configurable it would be easier for me to set it up. And I could use /data in the docker-compose-setup as well and would not have to volume mount two directories (now I am mounting /data and /tmp).
- add a logging endpoint for the processors: as now we have the log-file-path in the database it would be very easy to add and endpoint to the processing server to retrieve the log of a processor through http

ocrd_network/ocrd_network/deployer.py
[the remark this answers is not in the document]
MehmedGIT: @joschrew, you are right. I paid attention for the root log folder to be consistent since it will be volume mapped in the docker environment. So I did not use the temp dir there. However, I did not realize that the same issue is here for the socket files as well.

ocrd_network/ocrd_network/logging.py
MehmedGIT: It will be.

ocrd_network/ocrd_network/deployer.py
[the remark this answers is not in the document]
kba: `tempfile.gettempdir()` is `/tmp` in most cases. I agree we should make this overrideable (cf. https://github.com/OCR-D/core/pull/1111#discussion_r1356583376).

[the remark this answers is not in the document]
joschrew: ok, thanks for clarifying, I thought it would create `/tmp/<random-name>` ...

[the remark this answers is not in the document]
MehmedGIT: @kba, `tempfile.gettempdir()` indeed creates a temporary directory under `/tmp`, so it is still `/tmp/<random-name>`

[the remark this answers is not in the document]
kba: How do you mean?

```python
>>> import tempfile
>>> tempfile.gettempdir()
'/tmp'
```

What does that return for you?

ocrd_network/ocrd_network/logging.py
MehmedGIT: Resolved in https://github.com/OCR-D/core/pull/1111/commits/464f8eaf8c2c696668699a880f885dea14847f1b

ocrd_network/ocrd_network/logging.py
MehmedGIT: Resolved in https://github.com/OCR-D/core/pull/1111/commits/464f8eaf8c2c696668699a880f885dea14847f1b

ocrd_network/ocrd_network/deployer.py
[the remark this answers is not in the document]
MehmedGIT: @kba, sorry for the confusion I have created - you're right. The issue I had was related to another bug.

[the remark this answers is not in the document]
MehmedGIT: Resolved in https://github.com/OCR-D/core/pull/1111/commits/464f8eaf8c2c696668699a880f885dea14847f1b

ocrd_network/ocrd_network/logging.py
MehmedGIT: > I think this should use the constants defined

Which defined constants? There's just a defined list with the modules. Do you mean we need an enum here?

joschrew: You are right, probably I thought it was an enum, my bad. I would keep it then as it is now.

kba (approved):
Looks good, much cleaner and being able to get the log via the API is excellent.

Ready for release IMHO. Can you summarize the changes for the release notes?

MehmedGIT:
> Can you summarize the changes for the release notes?

It's on top. I have updated the initial post.

MehmedGIT:
@kba, I forgot to update the `ocrd --help` section - thanks for adding that.

Considering this, I was wondering why should we add the environment variables manually to the help section. I think it would be possible to iterate over all variables inside the `OcrdEnvConfig` and print their descriptions. Based on prefix values we could also decide what to print, e.g., `OCRD_NETWORK` prefix would print only those environment variables.

joschrew (approved):

kba:
> Considering this, I was wondering why should we add the environment variables manually to the help section. I think it would be possible to iterate over all variables inside the `OcrdEnvConfig` and print their descriptions. Based on prefix values we could also decide what to print, e.g., `OCRD_NETWORK` prefix would print only those environment variables.

The main reasons I did the `--help` section manually is order of the output and that I found it difficult to get the `\b` escapes right. It is a source of inconsistency though, I'll think about finding a better way.

PR #1116, opened 2023-10-18
remove dump_json flag

MehmedGIT:
Related to https://github.com/OCR-D/ocrd_tesserocr/issues/196

### DIFF
--- a/ocrd/ocrd/decorators/__init__.py
+++ b/ocrd/ocrd/decorators/__init__.py
@@ -154,7 +154,6 @@ def check_and_run_network_agent(ProcessorClass, subcommand: str, address: str, d
 
     processor = ProcessorClass(workspace=None)
     if subcommand == 'worker':
-        # TODO: Passing processor_name and ocrd_tool is redundant
         processing_worker = ProcessingWorker(
             rabbitmq_addr=queue,
             mongodb_addr=database,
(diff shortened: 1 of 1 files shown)


ocrd/ocrd/decorators/__init__.py
```diff
@@ -152,9 +152,9 @@ def check_and_run_network_agent(ProcessorClass, subcommand: str, address: str, d
         if not queue:
             raise ValueError(f"Option '--queue' required for subcommand {subcommand}")
 
-    processor = ProcessorClass(workspace=None, dump_json=True)
+    processor = ProcessorClass(workspace=None)
     if subcommand == 'worker':
-        # TODO: Passing processor_name and ocrd_tool is reduntant
+        # TODO: Passing processor_name and ocrd_tool is redundant
```
kba, line 157: It is not redundant though, because we don't always (bashlib processors) have a ProcessorClass to derive ocrd-tool.json from.

Minor note while checking this: maybe we can use consistent variable name `processor_class` instead of `processorClass`, `ProcessorClass` and `processor_class`.

kba (approved):

ocrd/ocrd/decorators/__init__.py
MehmedGIT: > It is not redundant though, because we don't always (bashlib processors) have a ProcessorClass to derive ocrd-tool.json from.

You are right. The comment is old - removed.

> Minor note while checking this

Sure, that should be generalized in the `ocrd_network` package. However, I would postpone it for now since there are more pressing tasks such as testing the mets server with Operandi.

Issue #1119, opened 2023-10-18
Error starting "bashlib-worker"


joschrew:
I tried to start a worker with bashlib (ocrd-olena) with this command (docker-compose): `ocrd-olena-binarize worker --database $MONGODB_URL --queue $RABBITMQ_URL` and I get `/build/core/ocrd/ocrd/lib.bash: line 172: ocrd__argv[log_filename]: unbound variable`.
Then I tried `ocrd-olena-binarize worker --database $MONGODB_URL --queue $RABBITMQ_URL --logfile-name something.log` and then I get:
```
Usage: ocrd network processing-worker [OPTIONS] PROCESSOR_NAME
Try 'ocrd network processing-worker --help' for help.

Error: No such option: --log-filename
```
I would just remove the `--logfile-name` thing from lib.bash again which was introduced [here](https://github.com/OCR-D/core/commit/6fc606027aee93efd60da55e70fe8967c508da2#diff-7826d88e116f00e2c0a23335b864dcce617770e57ced1877b14d5ca74748b5ea
but I don't know if that was put there with something in mind.

MehmedGIT:
#1120 fixes it.

PR #1120, opened 2023-10-18
remove logfile from bashlib

MehmedGIT:
Fix for #1119

### DIFF
--- a/ocrd/ocrd/lib.bash
+++ b/ocrd/ocrd/lib.bash
@@ -145,7 +145,6 @@ ocrd__parse_argv () {
             -I|--input-file-grp) ocrd__argv[input_file_grp]=$2 ; shift ;;
             -w|--working-dir) ocrd__argv[working_dir]=$(realpath "$2") ; shift ;;
             -m|--mets) ocrd__argv[mets_file]=$(realpath "$2") ; shift ;;
-            --log-filename) ocrd__argv[log_filename]="$2" ; shift ;;
             --mets-server-url) ocrd_argv[mets_server_url]="$2" ; shift ;;
             --overwrite) ocrd__argv[overwrite]=true ;;
             --profile) ocrd__argv[profile]=true ;;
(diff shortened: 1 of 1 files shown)


kba (approved):
Yes, that was an intermediate state I forgot to fix. LGTM

bertsky:
@kba @MehmedGIT the `--log-filename` is still in the general CLI though. It is still used for the processing worker (and the server), but comes in via configuration option `NetworkLoggingDirs.PROCESSING_JOBS` instead of the CLI. Since it was on the main processor CLI, I thought it was a mistake and added it to #1240 in https://github.com/OCR-D/core/pull/1240/commits/4f88f1d209bc86cdaea031ae9b2bda685e4f8fba. Do we really want that, though? If yes, then I must reinstante the bashlib side as well.

MehmedGIT:
@bertsky The `ocrd_network` logging should have nothing to do with the `--log-filename` flag coming from the general CLI. That is not passed down to the `ProcessingWorker` instance. I think that is a leftover from old code and should be removed. Unless the user wants to forward their logs to a separate file with the CLI - in that case it is fine.

We have structured logging in the `ocrd_network`. `NetworkLoggingDirs.PROCESSING_JOBS` is just a directory constant for processing job logs. Each processing job log file is stored inside with a unique name: `{job_id}.log`. When the ProcessingWorker invokes a processing job with `run_processor` or `run_cli`, the ProcessingWorker passes the unique job log file name as a parameter. Each ProcessingWorker also has its log file (`f"{AgentType.PROCESSING_WORKER}.{pid}.{processor_name}.log"`) and these are stored under `NetworkLoggingDirs.PROCESSING_WORKERS`.

bertsky:
I'll take that as a `yes` and will re-instante the option for bashlib (but only for the processing case, not networking).

MehmedGIT:
> (but only for the processing case, not networking)

Yes!

PR #1121, opened 2023-10-19
Fix loglevel type error when starting bashlib worker

joschrew:
The log-level is provided as an Integer but a String, `'DEBUG' ` for example, is needed.

### DIFF
--- a/ocrd/ocrd/processor/helpers.py
+++ b/ocrd/ocrd/processor/helpers.py
@@ -6,12 +6,13 @@ from time import perf_counter, process_time
 from functools import lru_cache
 import json
 import inspect
-from subprocess import run, PIPE
-from typing import List, Type
+from subprocess import run
+from typing import List
 
 from click import wrap_text
 from ocrd.workspace import Workspace
-from ocrd_utils import freeze_args, getLogger, pushd_popd, config, setOverrideLogLevel
+from ocrd_utils import freeze_args, getLogger, config, setOverrideLogLevel
+from logging import getLevelName
 
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


kba (changes requested):
I would fix the problem at the source of the call in ocrd_network/ocrd_network/process_helpers.py

```diff
- return_code = run_cli(                   
-     executable=executable,
-     # [...]
-     log_level=logging.DEBUG              
- )
+ return_code = run_cli(                   
+     executable=executable,
+     # [...]
+     log_level='DEBUG'
+ )                                        
```

joschrew:
I have already fixed that (just now) with accepting both str and int (e.g. `loglevel.DEBUG`). Test should pass now

ocrd/ocrd/processor/helpers.py
```diff
@@ -198,7 +199,10 @@ def run_cli(
     args = [executable, '--working-dir', workspace.directory]
     args += ['--mets', mets_url]
     if log_level:
-        args += ['--log-level', log_level]
+        if isinstance(log_level, str):
+            args += ['--log-level', log_level]
+        else:
+            args += ['--log-level', getLevelName(log_level)]
```
kba, line 205:
```suggestion
        args += ['--log-level', log_level if isinstance(log_level, str) else getLevelName(log_level)]
```

ocrd/ocrd/processor/helpers.py
```diff
@@ -6,12 +6,13 @@
...
-from subprocess import run, PIPE
-from typing import List, Type
+from subprocess import run
+from typing import List
 
 from click import wrap_text
 from ocrd.workspace import Workspace
-from ocrd_utils import freeze_args, getLogger, pushd_popd, config, setOverrideLogLevel
+from ocrd_utils import freeze_args, getLogger, config, setOverrideLogLevel
+from logging import getLevelName
```
kba, line 15:
```suggestion
from ocrd_utils import freeze_args, getLogger, config, setOverrideLogLevel, getLevelName
```

ocrd/ocrd/processor/helpers.py
kba: We could also change getLevelName to handle that, so we don't need the `isinstance` check here.

kba (approved):

PR #1122, opened 2023-10-19
Fix lib.bash receiving mets-server-url option

joschrew:
Typo in lib.bash when receiving the parameter mets-server-url. The mets-server-url has to be used in the bashlib-processor itself (in my case ocrd-olena) so that the mets-server is eventually used, because this is not handled by lib.bash e.g. core.
TODO: link and create ocrd-olena issue for using the mets-server


joschrew:
duplicate of https://github.com/OCR-D/core/pull/1123 (same fix is present, together with additional stuff)

PR #1123, opened 2023-10-19
Fix logging to a file with bashlib based processors

MehmedGIT:
This PR fixes the logging errors of type:
```
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib/python3.7/logging/__init__.py", line 1028, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
```

EDIT: Forcing InitLogging() just for Python processors but not for bashlib processors was the simple fix.

PR #1124, opened 2023-10-19
METS server: POST /reload to reload METS from disk, #1123

kba:
Implements an endpoint `POST/reload` and method `ClientSideOcrdMets.reload()` to reload the METS from disk.

This is in support of use cases where not all the processors in a workflow go through the METS server (we currently lack proper support in bashlib processors).

Running a processor without --mets-server-url should happen like this:
- Save the METS of the server to disk: `PUT /`
- Run the processor
- Reload the METS: `POST /reload`

### DIFF
--- a/ocrd/ocrd/mets_server.py
+++ b/ocrd/ocrd/mets_server.py
@@ -114,6 +114,9 @@ class ClientSideOcrdMets():
     def workspace_path(self):
         return self.session.request('GET', f'{self.url}/workspace_path').text
 
+    def reload(self):
+        return self.session.request('POST', f'{self.url}/reload').text
+
     @deprecated_alias(ID="file_id")
     @deprecated_alias(pageId="page_id")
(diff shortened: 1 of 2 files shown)


MehmedGIT (approved):
For the current time being should we potentially force the processing worker to always reload if the processor used is a bashlib processor? Could you name other bashlib processors where a such force should also apply?

kba:
> For the current time being should we potentially force the processing worker to always reload if the processor used is a bashlib processor? Could you name other bashlib processors where a such force should also apply?

`ocrd-fileformat` is the most important other bashlib processor but since by its nature, it is the last processor in the workflow normally, not really pertinent. Same for `ocrd-pagetopdf`.

`ocrd-page-transform` and `ocrd-im6convert` are two other processors where this is relevant.

I will implement `--mets-server-url` for olena asap, so this workaround should not be necessary at least in this case anymore.

PR #1127, opened 2023-10-19
add file handler for bashlib

MehmedGIT:
This PR implements the file handling envisioned in #1123. Unfortunately, the simple solution there is not working when the bashlib processors are natively deployed by the processing server.


kba (approved):
I would have prefered the simpler solution of course, but as it was not working in combination with the deployer, let's go for this one and hope we'll be able to revert to it later on. For now, getting it shipped is more important.

Issue #1133, opened 2023-11-13
`ocrd workspace add`: ORDER-attribute is not set during writing/updating 


csidirop:
If updating a METS without (for whatever reason) the structMap PHYSICAL entry of the page, the ORDER label is not set.

The ORDER attribute is mandatory, isn't it?

In addition, consider whether it would make sense to use this as an additional parameter.

kba:
Can you share the workspace and sequence of calls that lead to this issue? For the most part, the METS handling in core should not change what was input, except for newly generated elements.

I.e. if I take an example like https://content.staatsbibliothek-berlin.de/dc/PPN631277528.mets.xml, there are `@ORDER` and `@ORDERLABEL` on the `mets:div` and remain so when serializing.

I'll be happy to add better support for `@ORDER` once I understand what goes wrong.

csidirop:
Thanks for your response! I'm adding single ALTOs to an existing METS-file.

I'm calling `ocrd --log-level INFO workspace add --force --file-grp FULLTEXT --file-id "fulltext-$pageId" --page-id="$pageNum" --mimetype text/xml "$pageId.xml" `. This creates a `<mets:fileGrp USE="FULLTEXT">` node and a new `structMap PHYSICAL`-entry with `<mets:div TYPE="page" ID="XXX">` is created and the original untouched. 

If the ID is known, the `structMap' entry will be updated. This may only be a problem if you use it in a script or with a lot of METS automatically.

kba:
Thanks, I understand now. We need a way to either add `@ORDER`, `@ORDERLABEL` and `@TYPE` (any other attributes) to newly created page-specific `mets:div`. 

Here's two possibilities to add that:
1. `ocrd workspace update-page --order ... --order-label ... --type ... $PAGE_ID` which changes/adds the attributes to the `mets:div` with `@ID` `$PAGE_ID`
2. Add these attributes to `ocrd workspace add/bulk-add`.

Solution 1. is not too complicated to implement. If it is enough for your use case, I would prefer to leave it at that: You add your files with `ocrd workspace add` and then once `ocrd workspace update-page` to set the attributes.

If it is important for your use case to be able to do it as part of `ocrd workspace add/bulk-add`, then we'll need to find a way to make this as unobtrusive as possible because both `add` and `bulk-add` already have a pretty convoluted CLI and API.

kba:
[...]

Issue #821, opened 2022-03-18
workspace list-page: show label


bertsky:
Sometimes physical page identifiers look like `phys_fcd565cd-d170-4774-bb37-9aa25b68370b` – which is impossible to type/memorize. But even names like `PHYS_0001` are not as helpful as the `@ORDERLABEL` and `@ORDER` attributes.

Since presentation systems usually combine the latter two (`@ORDER [@ORDERLABEL]` IINM) we should offer something of the sort, too.

So `ocrd workspace list-page` could use an option (for backwards compatibility) to display these labels in extra columns.

But also, the page range operator could be made aware (somehow) of these labels, allowing me to specify, say `-g 20` instead of `-g phys_fcd565cd-d170-4774-bb37-9aa25b68370b`.

bertsky:
> We would have to refactor `OcrdMets.physical_pages` to return the complete `mets:div` element (instead of just the `@ID` string), so all attributes can be queried from it.

Oh, I completely forgot about #1063, which already solves that first part (by providing an extra kwarg to `get_physical_pages`).

kba:
All of these features are now merged with #1063:

* `OcrdMets.get_physical_pages` can return the `mets:div` elements instead of just the `ID`
* `ocrd workspace list-page` supports `ORDER`, `ORDERLABEL`, `LABEL` and `CONTENTIDS` in addition to `ID`
* `ocrd workspace update-page --set` allows setting them
* `-g` ranges can refer to any of them

PR #1134, opened 2023-11-13
implement ocrd workspace update-page

kba:
As discussed in #1133

### DIFF
--- a/ocrd/ocrd/cli/workspace.py
+++ b/ocrd/ocrd/cli/workspace.py
@@ -629,6 +629,20 @@ def set_id(ctx, id):   # pylint: disable=redefined-builtin
     workspace.mets.unique_identifier = id
     workspace.save_mets()
 
+@workspace_cli.command('update-page')
+@click.option('--order', help="@ORDER attribute for this mets:div", metavar='ORDER')
+@click.option('--orderlabel', help="@ORDERLABEL attribute for this mets:div", metavar='ORDERLABEL')
+@click.option('--contentids', help="@CONTENTIDS attribute for this mets:div", metavar='ORDERLABEL')
+@click.argument('PAGE_ID')
+@pass_workspace
+def update_page(ctx, order, orderlabel, contentids, page_id):
+    """
+    Update the @ORDER, @ORDERLABEL o @CONTENTIDS attributes of the mets:div with @ID=PAGE_ID
+    """
+    workspace = Workspace(ctx.resolver, directory=ctx.directory, mets_basename=ctx.mets_basename, automatic_backup=ctx.automatic_backup)
+    workspace.mets.update_physical_page_attributes(page_id, order=order, orderlabel=orderlabel, contentids=contentids)
+    workspace.save_mets()
+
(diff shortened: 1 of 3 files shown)


csidirop:
Works for me. Thank you!

kba:
> Should there be a check if there already exists an element with the same `ORDER`? Or is only the `ID` the "unique" key and `ORDER` only an additional information. If I understand the standard correctly all attributes are optional, so it should not matter if there are multiple ORDER-attributes with the same value, right?

If `@ID` is provided, it is required to be unique.

As for `@ORDER` and `@ORDERLABEL`:

```xml
<xsd:attribute name="ORDER" type="xsd:integer" use="optional">
<xsd:annotation>
<xsd:documentation xml:lang="en">ORDER (integer/O): A representation of the element's order among its siblings (e.g., its absolute, numeric sequence). For an example, and clarification of the distinction between ORDER and ORDERLABEL, see the description of the ORDERLABEL attribute. </xsd:documentation>
</xsd:annotation>
</xsd:attribute>
<xsd:attribute name="ORDERLABEL" type="xsd:string" use="optional">
<xsd:annotation>
<xsd:documentation xml:lang="en">ORDERLABEL (string/O): A representation of the element's order among its siblings (e.g., “xii”), or of any non-integer native numbering system. It is presumed that this value will still be machine actionable (e.g., it would support ‘go to page ___’ function), and it should not be used as a replacement/substitute for the LABEL attribute. To understand the differences between ORDER, ORDERLABEL and LABEL, imagine a text with 10 roman numbered pages followed by 10 arabic numbered pages. Page iii would have an ORDER of “3”, an ORDERLABEL of “iii” and a LABEL of “Page iii”, while page 3 would have an ORDER of “13”, an ORDERLABEL of “3” and a LABEL of “Page 3”. </xsd:documentation>
</xsd:annotation>
</xsd:attribute>
```

(from https://www.loc.gov/standards/mets/mets.xsd)

So, no, I don't think we should enforce unique `@ORDER`.

But I will check again whether the generic solution here produces valid order values before merging, these should be `xsd:integer`.

bertsky:
> (from https://www.loc.gov/standards/mets/mets.xsd)
> 
> So, no, I don't think we should enforce unique `@ORDER`.

@kba, but IMO we have to respect DFG's application profile for METS here, too. Version 2.3.1 says regarding physical structMap in section 2.2.2.1 (on p.14):
> - The ORDER attribute has to contain a numerical sorting value,
which is used to bring the individual pages into their correct
physical sequence.
> - The ORDERLABEL attribute can contain the pagination of the
individual page according to the original.
> - The LABEL attribute can include the foliation of the individual
page or the specification of record side and track number.

So regarding the options you presented in https://github.com/OCR-D/core/issues/1133,

> 1. `ocrd workspace update-page --order ... --order-label ... --type ... $PAGE_ID` which changes/adds the attributes to the `mets:div` with `@ID $PAGE_ID`
> 2. Add these attributes to `ocrd workspace add/bulk-add`,

I would have preferred option 2 for setting.

See #821 for getting that info on the CLI.

PR #1136, opened 2023-11-15
Change api processor paths in ocrd_network

joschrew:
This is to simplify the api for handling processor requests. `processor_name` is removed where it is not needed and the remaining paths are adapted to that.
As discussed on Monday. Separate PR for workflow will follow later.

### DIFF
--- a/ocrd_network/ocrd_network/processing_server.py
+++ b/ocrd_network/ocrd_network/processing_server.py
@@ -135,7 +135,7 @@ class ProcessingServer(FastAPI):
         )
 
         self.router.add_api_route(
-            path='/processor/{processor_name}',
+            path='/processor/run/{processor_name}',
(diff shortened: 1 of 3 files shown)


MehmedGIT (approved):
Looks good. What about the endpoints of the Processor Server? We have not talked about it on Monday, but IMO it makes sense to adapt the same changes there as well. E.g., `/{job_id}` would become `/info/{job_id}`.

MehmedGIT:
I should not forget to update the workflow endpoint guide once we have this merged.

kba (approved):
LGTM, thanks @joschrew. I second @MehmedGIT, let's also change the processor server endpoints while we're at it.

joschrew:
good point, I will add it

MehmedGIT (approved):

Issue #1129, opened 2023-10-30
ocrd workspace bag: downloaded files are overwritten


MehmedGIT:
The download of remote images is currently broken with core v2.58.1. Consider the following part of a mets file.

```xml
<mets:file ID="FILE_0001_DEFAULT" MIMETYPE="image/jpg">
  <mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" LOCTYPE="URL" xlink:href="https://content.staatsbibliothek-berlin.de/dc/PPN631277528-00000001/full/max/0/default.jpg"/>
<mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="DEFAULT/FILE_0001_DEFAULT.jpg" LOCTYPE="OTHER" OTHERLOCTYPE="FILE"/></mets:file>
<mets:file ID="FILE_0002_DEFAULT" MIMETYPE="image/jpg">
  <mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" LOCTYPE="URL" xlink:href="https://content.staatsbibliothek-berlin.de/dc/PPN631277528-00000002/full/max/0/default.jpg"/>
<mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="DEFAULT/FILE_0002_DEFAULT.jpg" LOCTYPE="OTHER" OTHERLOCTYPE="FILE"/></mets:file>
<mets:file ID="FILE_0003_DEFAULT" MIMETYPE="image/jpg">
  <mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" LOCTYPE="URL" xlink:href="https://content.staatsbibliothek-berlin.de/dc/PPN631277528-00000003/full/max/0/default.jpg"/>
<mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="DEFAULT/FILE_0003_DEFAULT.jpg" LOCTYPE="OTHER" OTHERLOCTYPE="FILE"/></mets:file>
<mets:file ID="FILE_0004_DEFAULT" MIMETYPE="image/jpg">
```

After downloading the images, there is only a single image (of the last page) named `default.jpg`. It seems the wrong `<mets:FLocat>` is used to parse the name of the images to be downloaded.

I have not investigated in depth yet but I think https://github.com/OCR-D/core/pull/1113 broke it.

kba:
This might be because of bad defaults for the filename - can you share the workspace so I can debug further?

[...]

PR #1137, opened 2023-11-15
ocrd workspace bag: ensure filenames are unique

kba:
40fb94ede introduced a subtle but consequential bug: If files were to be downloaded, we used the `OcrdFile.basename` attribute blindly for downloading, which in the case of a remote URL might not be unique (e.g. IIIF-URLs that all end in `/default.jpg`). This in turn led to the bagger overwriting the file repeatedly, so that a workspace with 10 such links would result in an OCRD-ZIP with only one file.

This is the cause of the confounding behavior described in #1129.

This PR fixes this by falling back to the `mets:file/@ID` as the basename for downloading.


ocrd/ocrd/workspace_bagger.py
```diff
@@ -71,14 +72,11 @@ def _bag_mets_files(self, workspace, bagdir, ocrd_mets, processes):
                 if not file_grp_dir.is_dir():
```
MehmedGIT, line 72: If I understand correctly, the `_bag_mets_files` method would still try to download every remote file if not locally available. How about providing a way to exclude some file groups here to disable automatic downloading for them? That way I would no longer have to remove problematic file groups (Goobi in my case) entries from the mets file.

MehmedGIT: It may not be an efficient implementation but adding an extra argument `skip_filegroups` to `_bag_mets_files` which is a list of file groups and checking whether the current file `f` belongs to any of these would allow that. Then, the `bag` method should also accept that argument to allow control to the caller as well.

MehmedGIT: Another option would be `download_filegroups` which is the opposite, and only downloads the specified file groups. If empty, should download all - the current behaviour.

MehmedGIT (approved):
Looks good.

ocrd/ocrd/workspace_bagger.py
kba: We did not originally allow incomplete OCRD-ZIP, which were supposed to be completely self-contained for long-term archiving purposes.

I started a general solution to that problem [years ago](https://github.com/OCR-D/core/compare/master...clone-filter) but never finished it. This would have included both whitelisting and blacklisting of search queries, i.e. a more general solution to what you propose, also for `ocrd workspace clone` and related functionality.

However, I see the need as well (these goobi/kitodo `file:/` URLs are a problem in our collection as well), so I am open to at least support fileGrp filtering for the bagger. I'll prepare a PR.

PR #1139, opened 2023-11-17
Bagger filegrp filter

kba:
There are use cases, where it might be beneficial to only bag a subset of the file groups.

This PR introduces an inclusion/exclusion mechanism to filter which fileGrps whould be included.

E.g. to only include the `DEFAULT` and `TESS` fileGrps:

```
ocrd zip bag -I TESS -I DEFAULT
```

Or to include all fileGrps except `PRESENTATION` (which has the `goobi` URLs in our collections):

```
ocrd zip bag -X PRESENTATION
```

Programmatically, the relevant kwargs are `include_file_grps` and `exclude_file_grps`.


ocrd/ocrd/cli/zip.py
```diff
@@ -38,13 +38,15 @@ def zip_cli():
               default="mets.xml",
               help='Basename of the METS file.',
               show_default=True)
+@click.option('-q', '--include-file-grps', 'include_fileGrp', help="fileGrps to include", default=[], multiple=True)
+@click.option('-Q', '--exclude-file-grps', 'exclude_fileGrp', help="fileGrps to include", default=[], multiple=True)
```
MehmedGIT, line 42:
```suggestion
@click.option('-Q', '--exclude-file-grps', 'exclude_fileGrp', help="fileGrps to exclude", default=[], multiple=True)
```

ocrd/ocrd/decorators/mets_find_options.py
```diff
@@ -6,6 +6,8 @@ def mets_find_options(f):
         option('-m', '--mimetype', help="Media type to look for", metavar='FILTER'),
         option('-g', '--page-id', help="Page ID", metavar='FILTER'),
         option('-i', '--file-id', help="ID", metavar='FILTER'),
+        option('-q', '--include-file-grps', 'include_fileGrp', help="fileGrps to include", default=[], multiple=True),
+        option('-Q', '--exclude-file-grps', 'exclude_fileGrp', help="fileGrps to include", default=[], multiple=True),
```
MehmedGIT, line 10:
```suggestion
        option('-Q', '--exclude-file-grps', 'exclude_fileGrp', help="fileGrps to exclude", default=[], multiple=True),
```

ocrd/ocrd/cli/zip.py
```diff
@@ -38,13 +38,15 @@ def zip_cli():
...
 @click.option('-i', '--identifier', '--id', help="Ocrd-Identifier", required=True)
 @click.option('-m', '--mets', help="location of mets.xml in the bag's data dir", default="mets.xml")
 @click.option('-b', '--base-version-checksum', help="Ocrd-Base-Version-Checksum")
 @click.option('-t', '--tag-file', help="Add a non-payload file to bag", type=click.Path(file_okay=True, dir_okay=False, readable=True, resolve_path=True), multiple=True)
 @click.option('-Z', '--skip-zip', help="Create a directory but do not ZIP it", is_flag=True, default=False)
 @click.option('-j', '--processes', help="Number of parallel processes", type=int, default=1)
-def bag(directory, mets_basename, dest, identifier, mets, base_version_checksum, tag_file, skip_zip, processes):
+def bag(directory, mets_basename, dest, include_fileGrp, exclude_fileGrp, identifier, mets, base_version_checksum, tag_file, skip_zip, processes):
```
MehmedGIT, line 49: Maybe worth adding an extra `s` to the parameters `include/exclude_fileGrps` to make it more obvious it is a list of file groups. Same for all places.

MehmedGIT (approved):
Looks good, I have not tested using the CLI, but the tests already cover that.

ocrd/ocrd/cli/zip.py
kba: Thanks for spotting this, I originally had these as `{in,ex}clude_file_grps` but then thought it would be more consistent to have `fileGrp`, `include_fileGrp` and `exclude_fileGrp` for `OcrdMets.find_files` in the same style. So if we decide to extend this white/blacklist mechanism to other attributes such as `mimetype`, you don't have to think about the suffix, just prepend `include_` or `exclude_`. We also tend to use the singular in the CLI (`e.g. `list-page`, not `list-pages`).So I think I'll keep it as-is for consistency.

PR #1143, opened 2023-11-21
Workflow extension

joschrew:
Change the two existing workflow paths to run a workflow and to query a workflow job. 
Add endpoints for uploading, updating and downloading a workflow script. The scripts are stored in the mongodb.
Workflows can be run with a previously uploaded script or with providing a script when starting the workflow.

### DIFF
--- a/ocrd_network/ocrd_network/processing_server.py
+++ b/ocrd_network/ocrd_network/processing_server.py
@@ -226,7 +226,7 @@ class ProcessingServer(FastAPI):
             endpoint=self.upload_workflow,
             methods=['POST'],
             tags=['workflow'],
-            status_code=status.HTTP_200_OK,
+            status_code=status.HTTP_201_CREATED,
             summary='Upload/Register a new workflow script',
         )
         self.router.add_api_route(
--- a/ocrd_network/ocrd_network/processing_server.py
+++ b/ocrd_network/ocrd_network/processing_server.py
@@ -195,7 +195,7 @@ class ProcessingServer(FastAPI):
         )
 
         self.router.add_api_route(
-            path='/workflow',
+            path='/workflow/run',
             endpoint=self.run_workflow,
             methods=['POST'],
             tags=['workflow', 'processing'],
@@ -209,7 +209,7 @@ class ProcessingServer(FastAPI):
         )
 
         self.router.add_api_route(
-            path='/workflow/{workflow_job_id}',
+            path='/workflow/job/{workflow_job_id}',
             endpoint=self.get_workflow_info,
             methods=['GET'],
             tags=['workflow', 'processing'],
--- a/ocrd_network/ocrd_network/processing_server.py
+++ b/ocrd_network/ocrd_network/processing_server.py
@@ -14,7 +14,7 @@ from fastapi import (
     UploadFile
 )
 from fastapi.exceptions import RequestValidationError
-from fastapi.responses import FileResponse, JSONResponse
+from fastapi.responses import FileResponse, JSONResponse, PlainTextResponse
 
 from pika.exceptions import ChannelClosedByBroker
 from ocrd.task_sequence import ProcessorTask
@@ -29,13 +29,15 @@ from .database import (
     db_get_workflow_job,
     db_get_workspace,
     db_update_processing_job,
-    db_update_workspace
+    db_update_workspace,
+    db_get_workflow_script,
 )
 from .deployer import Deployer
 from .logging import get_processing_server_logging_file_path
 from .models import (
     DBProcessorJob,
     DBWorkflowJob,
+    DBWorkflowScript,
     PYJobInput,
     PYJobOutput,
     PYResultMessage,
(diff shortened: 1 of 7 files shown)


ocrd_network/ocrd_network/processing_server.py
```diff
@@ -209,14 +213,39 @@ def __init__(self, config_path: str, host: str, port: int) -> None:
...
             status_code=status.HTTP_200_OK,
             summary='Get information about a workflow run',
         )
 
+        self.router.add_api_route(
+            path='/workflow',
+            endpoint=self.upload_workflow,
+            methods=['POST'],
+            tags=['workflow'],
+            status_code=status.HTTP_200_OK,
```
kba, line 229:
```suggestion
            status_code=status.HTTP_201_CREATED,
```

MehmedGIT (approved):

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -817,3 +859,51 @@ async def get_workflow_info(self, workflow_job_id) -> Dict:
...
+        content = (await workflow.read()).decode("utf-8")
+        if not validate_workflow(content):
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                                detail="Provided workflow script is invalid")
+
+        db_workflow_script = DBWorkflowScript(
+            workflow_id=workflow_id,
+            content=content,
+        )
+        await db_workflow_script.insert()
```
kba, line 876: Maybe we could check in the database first whether this exact workflow already exists, to avoid redundancy.

ocrd_network/ocrd_network/utils.py
```diff
@@ -147,3 +148,20 @@ def stop_mets_server(mets_server_url: str) -> bool:
...
     return False
+
+
+def validate_workflow(workflow: str, logger=None) -> bool:
+    """ Check that workflow is not empty and parseable to a lists of ProcessorTask
+    """
+    if workflow.strip() == "":
+        if logger:
+            logger.info("Workflow is invalid (empty string)")
+        return
```
kba, line 159:
```suggestion
    if not workflow.strip():
        if logger:
            logger.info("Workflow is invalid (empty string)")
        return False
```

ocrd_network/ocrd_network/utils.py
```diff
@@ -147,3 +148,20 @@ def stop_mets_server(mets_server_url: str) -> bool:
...
+    try:
+        tasks_list = workflow.splitlines()
+        [ProcessorTask.parse(task_str) for task_str in tasks_list if task_str.strip()]
+    except BaseException as e:
```
kba, line 163:
```suggestion
    except Exception as e:
```

But even better, let `ProcessorTask.parse` raise (more specific) `ValueError`.

kba (commented):
LGTM, not yet tested though.

kba:
`ocrd_network/ocrd_network/models/workflow.py` seems missing - forgot to commit @joschrew?

ocrd_network/ocrd_network/processing_server.py
MehmedGIT: > Maybe we could check in the database first whether this exact workflow already exists, to avoid redundancy.

This would degrade the DB, wouldn't it? We should not iterate over the content of every workflow and compare it with the content of the currently uploaded workflow.

joschrew:
thx, that explains the test error I am currently looking trying to find. Not the first time this happens to me ...

ocrd_network/ocrd_network/processing_server.py
joschrew: I could add a hash for a workflow and compare that, this way the database should not be too busy with it. I'll give it a try.

kba: Or just use `md5sum(content)` as the mongo `_id`?

MehmedGIT: I would not alter the default handling of `_id` by assigning a user-defined variable to it. That `_id` is used internally for versioning from my understanding, and changing it manually may be problematic. Moreover, a single space in the same workflow would still lead to different hashes. But I am fine if we still go for the hash solution.

kba: OK, I have not worked intensively with MongoDB for many years, so pls disregard the `_id` hack idea.

A separate field with `hash(content.strip())` would still be nice though. Even if it does not catch all semantically equivalent workflows, it's still helpful I think to deduplicate.

ocrd_network/ocrd_network/utils.py
joschrew: I have implemented your suggestion to [raise ValueError](https://github.com/OCR-D/core/pull/1143/commits/d695f4955dfbc7d4be166ebb2db9520a1d7b7524)

ocrd_network/ocrd_network/processing_server.py
joschrew: I have implemented a check for duplication. I use md5 to hash the file before saving it. The hash is stored in the database too and before saving it is checked that the same hash does not exist yet. I thought about removing trailing whitespace and similar stuff but decided against it.
I am not too sure about the benefit of this, but I think the content check does not hurt much, so I would keep it for now.

MehmedGIT (approved):

kba (changes requested):

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -869,9 +871,19 @@ async def upload_workflow(self, workflow: UploadFile) -> Dict:
             raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
                                 detail="Provided workflow script is invalid")
 
+        content_hash = md5(content.encode("utf-8")).hexdigest()
+        try:
+            db_workflow_script = await db_find_first_workflow_script_by_content(content_hash)
+            if db_workflow_script:
+                raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail="The same workflow"
```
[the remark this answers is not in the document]
joschrew: Together with the 409 a detail-message is send with the id of the already existing workflow (one line below).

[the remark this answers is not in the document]
kba: Right, sorry for the noise, did not read properly :) thanks

ocrd_network/ocrd_network/processing_server.py
```diff
@@ -869,9 +871,19 @@ async def upload_workflow(self, workflow: UploadFile) -> Dict:
...
+                                    f"-script exists with id '{db_workflow_script.workflow_id}'")
+        except ValueError:
+            pass
```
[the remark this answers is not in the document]
kba: Yeah, fine in this instance.

PR #1147, opened 2023-12-10
Update the RabbitMQ Version to 3.12

MehmedGIT:
Update the RabbitMQ version to 3.12 to allow usage of [poison message handling](https://www.rabbitmq.com/quorum-queues.html#poison-message-handling). Quorum queues allow to set the `delivery-limit` policy for queues. I.e., how many times a message could fail before being dropped.

Currently, the processing worker drops the messages after they fail **once** (unlike what we would like to have - 3 times). The older versions of RabbitMQ supported the `redelivered` feature for checking whether a message was redelivered or not, however, there was no way to know how many times it was redelivered. 

TODO: I am still experimenting and will extend the processing worker implementation to achieve a redelivery count of 3 or ideally any number configured with an env variable, say, 'OCRD_NETWORK_WORKER_QUEUE_MESSAGE_RETRIES'.

### DIFF
--- a/ocrd_network/ocrd_network/constants.py
+++ b/ocrd_network/ocrd_network/constants.py
@@ -1,4 +1,6 @@
 NETWORK_AGENT_SERVER = 'server'
 NETWORK_AGENT_WORKER = 'worker'
-DOCKER_IMAGE_RABBIT_MQ = 'rabbitmq:3.12-management'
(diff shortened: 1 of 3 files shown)


MehmedGIT:
The changes in this PR are already included in #1191.

PR #1153, opened 2023-12-12
CD: push tag latest to Dockerhub, too

### LINKED ISSUE #1152: docker deploy does not produce latest anymore
https://github.com/OCR-D/core/blob/742906e330d5ef1139fd18c86b73c154c0a67eae/.circleci/config.yml#L90-L93

causes this:

dockerhub-ocrd-core

We should still provide an alias `latest` for the most recent version.

bertsky:
fixes #1152


kba:
While it was a somewhat conscious decision to only release versioned images (esp. to force the ocrd_all bulds to derive from well defined version rather than `latest`), we obviously don't want to break any deployments, so merging now.

PR #1156, opened 2023-12-14
defaults for mets_basename and mets_server_url

kba:
As discussed in https://github.com/slub/mets-mods2tei/pull/68, the `cli.workspace.WorkspaceCtx` class and `Resolver.resolve_mets_arguments` method now have defaults for `mets_basename` and `mets_server_url`.

Also, I replaced all hard-coded `metx.xml` with a constant `DEFAULT_METS_BASENAME` to avoid typos and make it easier to spot where we assume the default basename.

### DIFF
--- a/ocrd/ocrd/cli/bashlib.py
+++ b/ocrd/ocrd/cli/bashlib.py
@@ -14,6 +14,7 @@ import click
 from ocrd.constants import BASHLIB_FILENAME
 import ocrd.constants
 import ocrd_utils.constants
+from ocrd_utils.constants import DEFAULT_METS_BASENAME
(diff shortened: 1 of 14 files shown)


MehmedGIT (approved):
Looks good. I am just wondering if the `mets.xml` literals in `test_workspace_bagger.py` are intentional?

kba:
> Looks good. I am just wondering if the `mets.xml` literals in `test_workspace_bagger.py` are intentional?

Good point, I did not change the tests because I think it makes sense to hardcode constants in tests, and also to make sure I noticed if any tests break. I think it's best to leave the tests as they are for now.

Issue #1154, opened 2023-12-13
ocrd workspace bag: use attribute CONTENTIDS if present for image resouces


M3ssman:
# Description

In extension of #1137 please look first for attribute `CONTENTIDS` instead of falling for `ID` when trying to download image resources (or provide parameter to switch to `CONTENTIDS`).

In case of ULB's `CONTENTIDS` (or to any other value which is urn-like) additional processing is required: please exchange any `:` with a plus-sign (`+`), so `urn:nbn:de:gbv:3:1-113129-p0007-8` becomes `urn+nbn+de+gbv+3+1-113129-p0007-8`. 

One might find many samples in the [GT-Testexports](https://github.com/M3ssman/gt-test/blob/main/data/corpus-odem-ger-256/mets.xml) using [GT-Repo-Template](https://github.com/OCR-D/gt-repo-template)

Note:

Besides ULB Sachsen-Anhalt, digital objects from SBB provide this information, too, when issuing OAI-Records, c.f. [OAI-REcord Theoretischer Anarchismus](https://oai.sbb.berlin/?verb=GetRecord&metadataPrefix=mets&identifier=oai:digital.staatsbibliothek-berlin.de:PPN891267093), although it's looks more like a URL than an URN, for example: 
```
<mets:div CONTENTIDS="http://resolver.staatsbibliothek-berlin.de/SBB000202D300000095" ID="PHYS_0095" ORDER="95" ORDERLABEL="91" TYPE="page" >
```
which resolves straight to the expected page `https://digital.staatsbibliothek-berlin.de/werkansicht/?PPN=PPN891267093&PHYSID=PHYS_0095`.

PR #1157, opened 2023-12-14
WorkspaceBagger: Use, in order of preference, f.basename, f.contentids and f.ID for filenames

kba:
As requested in #1154, this PR introduces a `contentids` attribute for `OcrdFile`, which delegates to `OcrdMets.get_contentids_for_file`, which looks up the `CONTENTIDS` attribute of the `mets:div[@TYPE="page"]` that a file belongs to.

The bagger uses this information to set the filenames of the bagged files.

E.g. for this `mets:file`

```xml
<mets:file ID="FILE_0009_DEFAULT" MIMETYPE="image/tiff">                                                                                                             
  <mets:FLocat xmlns:xlink="http://www.w3.org/1999/xlink" LOCTYPE="URL" xlink:href="http://content.staatsbibliothek-berlin.de/dms/PPN85249078X/800/0/00000010.tif"/> 
</mets:file>                                                                                                                                                         
...
<mets:div CONTENTIDS="http://resolver.staatsbibliothek-berlin.de/SBB0001CA7900000010" ID="PHYS_0010" ORDER="10" ORDERLABEL="2" TYPE="page">
  <mets:fptr FILEID="FILE_0009_DEFAULT"/>                                                                                                  
</mets:div>                                                                                                                                
```

This file will be bagged as `DEFAULT/http_resolver_staatsbibliothek_berlin_de_SBB0001CA7900000010.tif`

If there was no `@CONTENTIDS` for the corresponding `mets:div[@TYPE="PAGE"]`, then the filename would be `DEFAULT/FILE_0009_DEFAULT.tif`.

A quick proof-of-concept to make sure this is the desired behavior, to be polished (e.g. adding setters for `contentids` and potentially also for `@ORDER`, `@ORDERLABEL` and make sure that we're consistent in all places where files are written out.

@M3ssman cannot use the "request review" feature because you're not in the OCR-D organization but would appreciate you providing one very much, thanks!

### DIFF
--- a/ocrd/ocrd/workspace_bagger.py
+++ b/ocrd/ocrd/workspace_bagger.py
@@ -18,6 +18,7 @@ from ocrd_utils import (
     unzip_file_to_dir,
     DEFAULT_METS_BASENAME,
     MIMETYPE_PAGE,
+    safe_filename,
     VERSION,
 )
 from ocrd_validators.constants import BAGIT_TXT, TMP_BAGIT_PREFIX, OCRD_BAGIT_PROFILE_URL
@@ -80,7 +81,10 @@ class WorkspaceBagger():
                     file_grp_dir.mkdir()
 
                 attr = 'local_filename' if f.local_filename else 'url'
-                basename = f.basename if f.basename else f"{f.ID}{MIME_TO_EXT.get(f.mimetype, '.xml')}"
+                if f.local_filename and f.basename:
+                    basename = f.basename
+                else:
+                    basename = safe_filename(f.contentids if f.contentids else f.ID) + MIME_TO_EXT.get(f.mimetype, '.xml')
                 _relpath = join(f.fileGrp, basename)
                 self.resolver.download_to_directory(file_grp_dir, getattr(f, attr), basename=basename)
                 changed_local_filenames[str(getattr(f, attr))] = _relpath
--- a/ocrd_models/ocrd_models/ocrd_mets.py
+++ b/ocrd_models/ocrd_models/ocrd_mets.py
@@ -722,6 +722,25 @@ class OcrdMets(OcrdXmlDocument):
         if len(ret):
             return ret[0]
 
+    def get_contentids_for_file(self, ocrd_file):
+        """
+        Get the ``@CONTENTIDS` attribute of the physical page (``@CONTENTIDS`` of the ``mets:structMap[@TYPE="PHYSICAL"]//mets:div[@TYPE="PAGE"]`` entry)
+        corresponding to the ``mets:file`` :py:attr:`ocrd_file`.
+        """
+        ret = []
+        if self._cache_flag:
+            for pageId in self._fptr_cache.keys():
+                if ocrd_file.ID in self._fptr_cache[pageId].keys():
+                    ret.append(self._page_cache[pageId].get('CONTENTIDS'))
+        else:
+            ret = self._tree.getroot().xpath(
+                '/mets:mets/mets:structMap[@TYPE="PHYSICAL"]/mets:div[@TYPE="physSequence"]/mets:div[@TYPE="page"][./mets:fptr[@FILEID="%s"]]/@CONTENTIDS' %
+                ocrd_file.ID, namespaces=NS)
+
+        # To get rid of the python's FutureWarning
+        if len(ret):
+            return ret[0]
+
     def remove_physical_page(self, ID):
         """
         Delete page (physical ``mets:structMap`` ``mets:div`` entry ``@ID``) :py:attr:`ID`.
--- a/ocrd_models/ocrd_models/ocrd_file.py
+++ b/ocrd_models/ocrd_models/ocrd_file.py
@@ -27,6 +27,7 @@ class OcrdFile():
             local_filename (Path): ``@xlink:href`` pointing to the locally cached version of the file in the workspace
             ID (string): ``@ID`` of this ``mets:file``
             loctype (string): DEPRECATED do not use
+            contentids (string): ``@CONTENTIDS`` of the ``mets:div`` in the ``mets:structMap[@TYPE="PHYSICAL]`` this file manifests
         """
         if el is None:
             raise ValueError("Must provide mets:file element this OcrdFile represents")
@@ -136,6 +137,15 @@ class OcrdFile():
         self.mets.set_physical_page_for_file(pageId, self)
 
     @property
+    def contentids(self):
+        """
+        Get the ``@CONTENTIDS`` of the physical ``mets:structMap`` entry corresponding to this ``mets:file``.
+        """
+        if self.mets is None:
+            raise Exception("OcrdFile %s has no member 'mets' pointing to parent OcrdMets" % self)
+        return self.mets.get_contentids_for_file(self)
+
+    @property
     def loctypes(self):
         """
         Get the ``@LOCTYPE``s of the ``mets:file``.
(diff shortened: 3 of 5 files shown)


ocrd/ocrd/workspace_bagger.py
```diff
@@ -80,7 +81,10 @@ def _bag_mets_files(
                     file_grp_dir.mkdir()
 
                 attr = 'local_filename' if f.local_filename else 'url'
-                basename = f.basename if f.basename else f"{f.ID}{MIME_TO_EXT.get(f.mimetype, '.xml')}"
+                if f.local_filename and f.basename:
+                    basename = f.basename
+                else:
+                    basename = safe_filename(f.contentids if f.contentids else f.ID) + MIME_TO_EXT.get(f.mimetype, '.xml')
```
bertsky, line 87: I would advise against direct use of `@CONTENTIDS` as file name. The URL prefix almost always is not what you want. How about stripping the host-name part (if in fact it is a URL), and then using `makedirs` for all remaining path prefixes?

ocrd_models/ocrd_models/ocrd_mets.py
```diff
@@ -722,6 +722,25 @@ def get_physical_page_for_file(self, ocrd_file):
         if len(ret):
             return ret[0]
 
+    def get_contentids_for_file(self, ocrd_file):
```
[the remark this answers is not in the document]
kba: That is a much better solution, I agree. As I said, this is just a proof-of-concept so we have a discussion basis for the naming. Implementation can be thrown away and replaced with a fine-tuned #1063.

It might be sensible to have an `OcrdMetsPage(s)` class similar to `OcrdFile` to provide a uniform interface. But that's best discussed in #1063.

ocrd/ocrd/workspace_bagger.py
kba: Fair points. Agree that representing the URL path as directory is a neat way to do it, though it deviates from our general flat directory structure below the `fileGrp` dirs. Removing the host is also prettier.

But I'm wondering how that would work for @M3ssman's use case - how much info do you need to still be able to debug your workflows?

M3ssman: It is not intended to be used as-it-is, simply because an URN/URI contains chars that are not valid as part in local filenames. Therefore - at least in ULB and semantics workflows - colon is exchanged with a plus sign.

Consider a page container like this

```
<mets:div ID="phys1278993" TYPE="page" CONTENTIDS="urn:nbn:de:gbv:3:1-113129-p0007-8" ORDER="1">
        <mets:fptr FILEID="IMG_MAX_1278993"/>
        <mets:fptr FILEID="OCR-D-GT-FULLTEXT-1"/>
</mets:div>

```
with GT linked

```
<mets:file ID="OCR-D-GT-FULLTEXT-1" MIMETYPE="application/vnd.prima.page+xml">
	<mets:FLocat xlink:href="GT-PAGE/urn+nbn+de+gbv+3+1-113129-p0007-8_ger.gt.xml" LOCTYPE="OTHER" OTHERLOCTYPE="FILE"/>
</mets:file>
```

but the actual imsge via URL like this

```
<mets:file MIMETYPE="image/jpeg" ID="IMG_MAX_1278993">
	<mets:FLocat xlink:href="https://opendata.uni-halle.de/retrieve/eeeee05d-c7cd-4e89-9607-5d1ac175afa1/00000007.jpg" LOCTYPE="URL"/>
</mets:file>
```

The goal is to match both files, GT and Image, by their names alone without any extensions. 
If this can be achieved, both can be used out-of-the-box for further GT-works in Tools like Transkribus or Larex.
(I have to handle 1.600 GT-ODEM-files, nearly 100 newspaper-GT-files, 101 GT arabic and about 400 pages GT persian. And if our next digi-project will be granted, the GT will increase further.)

Proposal: Instead of using the `CONTENTIDS` attribute it'd be sufficient to rename the images locally like the corresponding GT-file, whatever it's name was. My main concern is therefore to have equal names, not to name something like some attribute.

Maybe this way one hopefully avoids additional processing?





(c.f. https://github.com/M3ssman/gt-test/blob/a370f3a691506f4eab1b91226a76a7c1f461ba10/data/corpus-odem-ger-256/mets.xml#L8905)

kba: > It is not intended to be used as-it-is, simply because an URN/URI contains chars that are not valid as part in local filenames. Therefore - at least in ULB and semantics workflows - colon is exchanged with a plus sign.

It is not used as-is but passed through `safe_filename` which does

```python
def safe_filename(url):                                              
    """                                                              
    Sanitize input to be safely used as the basename of a local file.
    """                                                              
    ret = re.sub(r'[^\w]+', '_', url)                                
    ret = re.sub(r'^\.*', '', ret)                                   
    ret = re.sub(r'\.\.*', '.', ret)                                 
    #  print('safe filename: %s -> %s' % (url, ret))                 
    return ret                                                       
```

Adapting this to make the replacement (`_`) configurable as `+` is not an issue. However

> Proposal: Instead of using the `CONTENTIDS` attribute it'd be sufficient to rename the images locally like the corresponding GT-file, whatever it's name was. My main concern is therefore to have equal names, not to name something like some attribute.

I still don't understand how to achieve that. For your example

```xml
<mets:file ID="OCR-D-GT-FULLTEXT-1" MIMETYPE="application/vnd.prima.page+xml">
	<mets:FLocat xlink:href="GT-PAGE/urn+nbn+de+gbv+3+1-113129-p0007-8_ger.gt.xml" LOCTYPE="OTHER" OTHERLOCTYPE="FILE"/>
</mets:file>
...
<mets:file MIMETYPE="image/jpeg" ID="IMG_MAX_1278993">
	<mets:FLocat xlink:href="https://opendata.uni-halle.de/retrieve/eeeee05d-c7cd-4e89-9607-5d1ac175afa1/00000007.jpg" LOCTYPE="URL"/>
</mets:file>
...
<mets:div ID="phys1278993" TYPE="page" CONTENTIDS="urn:nbn:de:gbv:3:1-113129-p0007-8" ORDER="1">
        <mets:fptr FILEID="IMG_MAX_1278993"/>
        <mets:fptr FILEID="OCR-D-GT-FULLTEXT-1"/>
</mets:div>
```

What should the bagger write as the filenames of `IMG_MAX_1278993` and `OCR-D-GT-FULLTEXT-1`?

M3ssman: I'd wish to save `urn+nbn+de+gbv+3+1-113129-p0007-8_ger.gt.xml` as filename for the GT, since this is the file which `OCR-D-GT-FULLTEXT-1` points to, and for the image where container `IMG_MAX_1278993` refers to, a corresponding name like `urn+nbn+de+gbv+3+1-113129-p0007-8_ger.jpg` , if it's possible.

M3ssman: The point at least in my understanding - ALIMU- concerning ULB is, that the GT-files shall be published (and probably edited further afterwards) but will be tied to the GT-repository. There is nothing about to change with the images, they only need to be referenced and resolvable, for example, for later generation of training data using the GT-files.

M3ssman:
I'd like to provide additional testing (and test data as well) - is this possible in terms of this PR? 

(And I'm uncertain where to add tests. There's actually one in `tests/model/test_ocrd_mets.py` but I thought this was about the bagger, which has test in `tests/validator/test_workspace_bagger.py`?)

PR #1159, opened 2023-12-15
docker: we need .git during build for setuptools_scm

kba:
We now derive the version from the git repo, so we need to add it during the build.

BTW we don't need to add README.md and LICENSE, only to delete it in the last step, but not urgent.

Will merge and release right away so we can make sure dockerhub images get built.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@
 !Makefile
 !LICENSE
 !README.md
+!.git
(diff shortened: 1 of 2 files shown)


bertsky:
Ouch – did not think of this. 

`COPY` means that the repo data inevitably will stay in the image. But since we ultimately decided the same thing for ocrd_all, let's stick with this.

PR #1160, opened 2023-12-19
ocrd_utils.is_oai_content: downgrade loglevel

bertsky:
Perhaps we should exhaustively search for all such places...

IMHO core logging should not be gabby by default. 

You can of course argue that the user could increase their default logger `ocrd*` filter levels. But that's already an expert step.

In that light, I really don't understand what motivated https://github.com/OCR-D/core/commit/ed33ac69b4fd3c6ffd7033245ab2ff37f6d80b16 (maybe @kba can elaborate) and recommend reverting (i.e. raising the root and `ocrd` loggers, used both on the default console and file handler, to `INFO` again).

### DIFF
--- a/ocrd_models/ocrd_models/utils.py
+++ b/ocrd_models/ocrd_models/utils.py
@@ -49,7 +49,7 @@ def is_oai_content(data):
     log = getLogger('ocrd.models.utils.is_oai_content')
     xml_root = ET.fromstring(data)
     root_tag = xml_root.tag
-    log.info("response data root.tag: '%s'" % root_tag)
+    log.debug("response data root.tag: '%s'" % root_tag)
     return str(root_tag).endswith('OAI-PMH')
 
 


kba:
> In that light, I really don't understand what motivated ed33ac6 

This was part of a refactoring the logging to make it consistent and easier to debug while developing the network implementation. See #1101, #1109 and #1117 for the discussion.

> IMHO core logging should not be gabby by default.

That's also a valid point of view. Behavior can always be overridden, but we must decide on the defaults of course. I am agnostic here, since I always use an ocrd_logging.conf these days, so default behavior is less important to me. I think @MehmedGIT is on "team verbose" - could you live with a default level of `INFO`?

bertsky:
Thanks! 

BTW, can the global log level override embedded into the multi-CLI (`-l DEBUG`) still be used to get a debug log if needed – even for the servers?

PR #1164, opened 2024-01-10
ocrd_network base integration tests

MehmedGIT:
This PR is just a basic beginning for the `ocrd_network` integration tests.

What does this PR include:
- Some extensions to MongoDB methods in `ocrd_network` for easier testing of the DB functionalities
- New state `UNSET` for jobs for easier testing instead of relying on `None`
- New endpoints to the Processing Server:
  + endpoint `/` for just testing the connectivity to the Processing Server
  + endpoint `/workflow/job-simple/{workflow_job_id}` - simplified endpoint for checking the overall status of a workflow
- Processing Server adds `ocrd-dummy` processor and queue by default when started:
  + This is a hack since there is no endpoint for worker registration yet
  + This will be refactored in the next PR when refactoring the `ocrd_network` and adding a dedicated endpoint
- A `conftest.py` provided for using fixtures for the MongoDB and RabbitMQ
- MongoDB integration tests
- RabbitMQ integration tests
- A dummy workflow consisting of 3 `ocrd-dummy` processors
- Basic Processing Server integration tests:
  + processing endpoint test with `ocrd-dummy` (1 processing worker deployed)
  + workflow endpoint execution and status check test with the dummy workflow
- Separate `ocrd_logging.conf` for the ocrd_network integration tests for easier catching of bugs when tests are failing
- Fixing `.dockerignore` to include relevant parts for the tests
- Extension to core Dockerfile for creating an `ocrd_core_test` image to be used by the testing containers
- Extension to core Makefile to allow `make integration-test` call
- Integration test workflow for Github Actions.


tests/conftest.py
```diff
@@ -0,0 +1,16 @@
+from src.ocrd_utils.config import config
```
kba, line 1: We might want to use a separate `OcrdEnvConfig` instance, not the default `config` instance to avoid interfering with other tests

tests/network/docker-compose.yml
```diff
@@ -0,0 +1,140 @@
...
+    ports:
+      - "6701:27017"
+    environment:
+      - MONGO_INITDB_ROOT_USERNAME=network_test
+      - MONGO_INITDB_ROOT_PASSWORD=network_test
+    healthcheck:
+      test: echo 'db.runCommand("ping").ok' | mongosh localhost:27017/test --quiet
+      interval: 1s
+      timeout: 3s
+      retries: 30
```
kba, line 25: Nice, that is a useful pattern we should apply widely.

tests/network/docker-compose.yml
```diff
@@ -0,0 +1,140 @@
...
+            username: network_test
+            password: network_test
+        database:
+          address: mongodb-docker-host.ocrd_network_test
+          port: 27017
+          skip_deployment: true
+          credentials:
+            username: network_test
+            password: network_test
+        hosts: []\" > ./ps_config.yaml && \
```
kba, line 99: Could we make that somehow part of the `docker-compose.yml`. I see a potential for syntax errors and difficulty adapting when the config schema changes.

kba (approved):
Excellent, great work. a few minor questions, but ready for merge AFAICS. I'm running the tests and making sure that the multi-stage setup does not interfere with our other Docker deployments.

macos: I'll try to get it to work but it's not a blocker.

tests/conftest.py
MehmedGIT: True. I will add that.

tests/network/docker-compose.yml
MehmedGIT: If that works, I prefer using that instead of sleeping here and there to wait. So far I have not seen issues with that health check here and in Operandi.

tests/network/docker-compose.yml
MehmedGIT: Instead of doing that workaround, we could just volume map the `ps_config.yml`. However, the idea is that the credentials will be swapped with ENV variables. I have not found another optimal way to use environment variables inside the yaml file.

Dockerfile
```diff
@@ -33,13 +33,25 @@ RUN apt-get update && apt-get -y install software-properties-common \
...
+COPY Makefile .
+RUN make assets
+COPY tests ./tests
+COPY .gitmodules .
+COPY requirements_test.txt .
+RUN pip install -r requirements_test.txt
+RUN mkdir /ocrd-data && chmod 777 /ocrd-data
+
+CMD ["yes"]
+# CMD ["make", "test", "integration-test"]
```
joschrew, line 57: Can someone explain this to me? My understanding currently is that the core image is changed with this and now pip contains the requirements-test and executes `yes` by default. How can the core-Image still be build with that?

tests/network/test_rabbitmq.py
```diff
@@ -0,0 +1,83 @@
...
+def test_rmq_consume_ocrd_message(rabbitmq_consumer):
+    method_frame, header_frame, message = rabbitmq_consumer.get_one_message(
+        queue_name=DEFAULT_QUEUE,
+        auto_ack=True
+    )
+    assert method_frame.message_count == 0  # messages left in the queue
+    decoded_message = loads(message)
+    assert decoded_message["job_id"] == "Test_job_id"
+    assert decoded_message["workflow_id"] == "Test_workflow_id"
+    assert decoded_message["workspace_id"] == "Test_workspace_id"
```
joschrew, line 83: It seems to me some of the tests depend on each other or rather that they must be executed in order. It might be better to combine them into one test because I think the order of the test execution might not be from top to bottom always

Makefile
```diff
@@ -213,9 +215,17 @@ test: assets
...
 		$(TESTDIR)
 	cd ocrd_utils ; $(PYTHON) -m pytest --continue-on-collection-errors -k TestLogging -k TestDecorators $(TESTDIR)
 
+INTEGRATION_TEST_IN_DOCKER = docker exec core_test
+integration-test:
+	$(DOCKER_COMPOSE) --file tests/network/docker-compose.yml up -d
+	sleep 20
+	$(INTEGRATION_TEST_IN_DOCKER) pytest -k 'test_rmq or test_db or test_processing_server' -v
+	$(DOCKER_COMPOSE) --file tests/network/docker-compose.yml down --remove-orphans
+
```
joschrew, line 228: Some tests are failing currently on my machine: `test_db_workspace_update` (fails only sometimes), `test_processing_server_processing_request` and `test_processing_server_workflow_request`. 
When the tests are failing, the containers started by the Makefile are not stopped afterwards. I still have to find out why the tests fail. I will comment when I know more

joschrew (commented):
This are my first comments. I still have to find out why the tests fail on my side and then I will comment again.

Dockerfile
MehmedGIT: My understanding is that the `ocrd_core_base` layer that was built above is cached and continues from there to make assets, install test requirements, and so on. This is useful when only the test code is changed, it's enough to rebuild just the `ocrd_core_test`. However, if the source code changes as well, then `ocrd_core_base` is affected as well and is rebuilt again.

tests/network/test_rabbitmq.py
MehmedGIT: You are right. I will combine these.

Dockerfile
joschrew: But what comes out at the bottom is the core image containing the test stuff. Previously ocrd-core was build with this Dockerfile. But now when it is used to build core the final image contains all the unwanted test code (unwanted in production). I would think we need an extra Dockerfile for the Tests and one for creating the ocrd-core image (the state of the dockerfile before the PR).

Makefile
MehmedGIT: Unfortunately, I have not found a way yet to stop and remove the containers when some tests fail. What I use currently is manually stopping and removing all containers with 2 aliases for:
```bash
docker stop $(docker ps --format="{{.ID}}")
docker rm $(docker ps -a --format="{{.ID}}")
```

The reason why some DB tests fail is that you executed the tests again without stopping the DB container. I did not implement dropping DB collections when re-executing tests. Hence, some tests fail when entries exist in the DB. In case you know how to achieve dropping DB collections simply without refactoring the Processing Server and without introducing a second MongoDB client - feel free to commit.

Another reason why some Processing Server tests may fail is because the socket files for the mets server are still preserved in the host machine. Make sure no sockets are left from previous test executions.

Dockerfile
kba: We're using the multi-stage build feature https://docs.docker.com/build/building/multi-stage/

Essentially:
- Every `FROM` line starts a new stage
- Stage name is set via `FROM ... AS stage-name`
- The target stage can be provided with `--target stage-name`

In the `make docker` makefile target, we specify

```make
docker docker-cuda:                                                                                                                                     
  docker build --progress=plain -f $(DOCKER_FILE) -t $(DOCKER_TAG) --target ocrd_core_base --build-arg BASE_IMAGE=$(DOCKER_BASE_IMAGE) $(DOCKER_ARGS) . 
```

So, it only builds the `ocrd_core_base` stage, which is unchanged. So no need for two Dockerfiles.

joschrew: Thank you

tests/network/docker-compose.yml
MehmedGIT: I have decided to volume map the processing server configuration for now and keep adapting in two places. The processing server is no longer responsible for deploying RabbitMQ and MongoDB, according to the new spec, the config file is not needed in that case. The processing server can just get the MongoDB and RabbitMQ URLs as a parameter as processing workers do. However, the `ocrd_network` still needs to be refactored to reflect the latest spec changes.

Makefile
```diff
@@ -213,9 +215,17 @@ test: assets
...
 		--ignore-glob="$(TESTDIR)/**/*bench*.py" \
+		--ignore-glob="$(TESTDIR)/network/*.py" \
 		$(TESTDIR)
 	cd ocrd_utils ; $(PYTHON) -m pytest --continue-on-collection-errors -k TestLogging -k TestDecorators $(TESTDIR)
 
+INTEGRATION_TEST_IN_DOCKER = docker exec core_test
+integration-test:
+	$(DOCKER_COMPOSE) --file tests/network/docker-compose.yml up -d
+	sleep 20
+	$(INTEGRATION_TEST_IN_DOCKER) pytest -k 'test_rmq or test_db or test_processing_server' -v
```
joschrew, line 226:
```suggestion
	-$(INTEGRATION_TEST_IN_DOCKER) pytest -k 'test_rmq or test_db or test_processing_server' -v
```
With this an error (return != 0) in the line should be ignored and the stopping should be executed regardless if tests fail. But I am not sure if we really want that or if it has side effects.

MehmedGIT: I have intentionally failed some tests to check. This seems to work!

Makefile
MehmedGIT: This is now fully resolved. The socket files left from previous test executions are deleted before the tests.

joschrew: Yes, test are working now

joschrew (approved):

MehmedGIT:
Good to be shipped.

PR #1167, opened 2024-01-19
make OcrdFile.local_filename str again

bertsky:
Fixes a regression in #1079 which changed `.local_filename` type from str to Path

see https://github.com/OCR-D/core/pull/1079/files#r1459238204

### DIFF
--- a/ocrd_models/ocrd_models/ocrd_file.py
+++ b/ocrd_models/ocrd_models/ocrd_file.py
@@ -39,7 +39,7 @@ class OcrdFile():
         self.pageId = pageId
 
         if local_filename:
-            self.local_filename = Path(local_filename)
+            self.local_filename = local_filename
         if url:
             self.url = url
 


bertsky:
Not a good idea – too many related changes. Better directly revert https://github.com/OCR-D/core/commit/58bb31b52c716a774288b07ec559a5dd96f919f3

kba:
I tried reverting and starting to fix test errors but I think this is not worth the effort.

True, we should have communicated this more clearly in the changelog.

But I think the benefit of reverting this is not worth the effort of potentially breaking already-adapted code.

So unless there is no workaround, I would prefer to retain `isinstance(local_filename, Path)`.

bertsky:
> But I think the benefit of reverting this is not worth the effort of potentially breaking already-adapted code.
> 
> So unless there is no workaround, I would prefer to retain `isinstance(local_filename, Path)`.

Well, ... here's my arguments in **favour** of reverting to `str`:

- it's called `local_filename`, not `local_path`
- it is **set** via `str`, not via `Path`, so why should one not expect to **get** a `str` back?
- `local_filename` is also used as field/parameter on the CLI (which necessarily means str)
- I'm pretty sure it's easier to fix in core than in a multitude of external code bases (and we can never be sure we did not miss one, think Eynollah).
- Already adapted (external) code can easily be identified: search for `local_filename.` and change there. That's way more specific than the string `local_filename` in the other direction (adapting to new). Any `str(*.local_filename)` can be kept, because `str(str(.)) == str(.)`.

kba:
> I'm pretty sure it's easier to fix in core than in a multitude of external code bases

OK, then I will do the revert. And add typing information to avoid such a situation in the future.

kba:
Superseded by https://github.com/OCR-D/core/pull/1182

PR #1169, opened 2024-01-21
deps-cuda: add workaround for tf-keras#62

bertsky:


kba (approved):
I'm preparing a hotfix because docker build is broken, so I'll merge this right away. Thanks!

PR #1170, opened 2024-01-23
README: fix links to sub-readmes, move bashlib into separate README

bertsky:


kba (approved):
Thanks, also separating the bashlib info makes sense.

PR #1177, opened 2024-01-26
METS Server: try to remove UDS socket at sys.exit

kba:
Basically, just add `atexit(self.shutdown)`, so the socket file is removed when the process exits. This should cover most of the things that can go wrong, except `kill -9`/`C-\` which cannot be caught.

### DIFF
--- a/src/ocrd/mets_server.py
+++ b/src/ocrd/mets_server.py
@@ -2,24 +2,24 @@
 # METS server functionality
 """
 import re
-from os import environ, _exit, chmod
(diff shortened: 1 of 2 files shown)


MehmedGIT (approved):
This PR still does not resolve the issue. Nevertheless, it is a good addition.

After the Processing Server docker container is stopped and removed, the UDS socket stays where it was volume mapped on the host machine.

MehmedGIT:
I already took care of that in #1164 with the last commit :)

PR #1181, opened 2024-01-31
workspace: downgrade log messages from info to debug

bertsky:
We now have the INFO level default, but many log messages unfortunately have been promoted from DEBUG to INFO – this tries to revert that.

Perhaps we need to discuss further. (I kept the INFO level for `save_image_file`. Perhaps we should also have `save_mets` with INFO level.)


kba:
No objections from my side, let's reduce the log spam.

PR #1184, opened 2024-02-08
Test for ocrd-network

joschrew:
This PR adds a test for ocrd-network. Purpose is that the test is available in the docker-image and can then be used with ocrd_all. It does not run with just core itself ( I am not sure if it is good to do it this way, because it is not supposed to run with "just" plain core).

This PR is required for: https://github.com/OCR-D/ocrd_all/pull/407


bertsky (commented):
I don't see a problem integrating this into core proper.

Even the workflow file should be here (besides dummy-workflow.txt), not in ocrd_all.

And core/tests/network/docker-compose.yml could already bind-mount it to the right location. (This compose file in turn can be included in ocrd_all's compose file.)

Since that workflow test cannot be run without ocrd_all, it should simply be skipped by default.

kba:
> I don't see a problem integrating this into core proper.
> 
> Even the workflow file should be here (besides dummy-workflow.txt), not in ocrd_all.
> 
> And core/tests/network/docker-compose.yml could already bind-mount it to the right location. (This compose file in turn can be included in ocrd_all's compose file.)
> 
> Since that workflow test cannot be run without ocrd_all, it should simply be skipped by default.

I see the benefit of DRY and composing the behavior from existing/proven `docker-compose.yml`. However, I think in this case I think it would be better to not depend on OCR-D/core for the intergration test:

- it requires some hard-to-read constructs, like `build: !reset null` and inconsistent relative paths in the derived `docker-compose.yml`
- Changing the workflow to run requires updating core
- The reusable parts, i.e. up RabbitMQ, MongoDB and processing server is generic enough that the benefit of a simple setup (one directory in ocrd_all with the workflow(s), docker-compose.yml and pytest scripts) outweighs the drawback of having some redundancy.

bertsky:
> * Changing the workflow to run requires updating core

ok, good point!



> * The reusable parts, i.e. up RabbitMQ, MongoDB and processing server is generic enough that the benefit of a simple setup (one directory in ocrd_all with the workflow(s), docker-compose.yml and pytest scripts) outweighs the drawback of having some redundancy.

there's a lot of stuff in there, which will still likely evolve a lot. This would have to be synced with ocrd_all then.

If we don't want to inherit/include from core, perhaps we could just have a makefile rule `cp` from core, and have a few git-controlled override files in place (like docker-compose.override.yml)?

Dockerfile
```diff
@@ -44,9 +44,11 @@ WORKDIR /data
 CMD ["/usr/local/bin/ocrd", "--help"]
 
 FROM ocrd_core_base as ocrd_core_test
+# Optionally skip make assets with this arg
+ARG SKIP_ASSETS
 WORKDIR /build-ocrd
 COPY Makefile .
-RUN make assets
+RUN if test -z "$SKIP_ASSETS" || test $SKIP_ASSETS -eq 0 ; then make assets ; fi
```
kba, line 51: I would expect `SKIP_ASSETS=0` to disable the behavior.  Did you mean `test $SKIP_ASSETS -eq 1`?

tests/network/test_ocrd_all_workflow.py
```diff
@@ -0,0 +1,22 @@
...
+    post_ps_workflow_request,
+)
+
+PROCESSING_SERVER_URL = test_config.PROCESSING_SERVER_URL
+
+
+def test_ocrd_all_workflow():
+    # This test is supposed to run with ocrd_all not with just core on its own
+    # Note: the used workflow path is volume mapped
+    path_to_wf = "/ocrd-data/assets/ocrd_all-test-workflow.txt"
```
kba, line 14: Why the `ocrd-data` prefix, why not just use `/data`?

kba (approved):

Dockerfile
MehmedGIT: I think `SKIP_ASSETS` should be renamed to `MAKE_ASSETS` to reverse the logic.

`SKIP_ASSETS` itself has a negative meaning. False to False makes True.

SKIP_ASSETS=1 (true) -> do not make assets
SKIP_ASSETS=0 (false) -> make assets

tests/network/test_ocrd_all_workflow.py
MehmedGIT: I used `/ocrd-data` instead of `/data` as well, [here](https://github.com/OCR-D/core/blob/refactor-network/tests/network/docker-compose.yml). My reasoning at least was to avoid potential accidental overwriting of an already existing `/data` folder coming from the `ocrd_all` image.

Dockerfile
joschrew: for me 0 is false so if `SKIP_ASSETS` is 0 then it is not skipped.

tests/network/test_ocrd_all_workflow.py
joschrew: I use a mounted docker volume (not host directory) to transfer the test-workspace. I didn't want to interfere with /data because that is usally used to mount a host folder to the container. But maybe it is better to use data, I didn't try that yet.

PR #1193, opened 2024-02-23
OcrdMets.get_physical_pages_for: edge case empty list for_fileIds

### LINKED ISSUE #1192: OcrdMets.get_physical_pages regression: assertion on non-empty result
This first surfaced via [ocrd-import](https://github.com/slub/ocrd_controller/issues/37), which for each file determines if adding it would create a clash for that pageId. In doing so, it usually expects an empty result for commands like `ocrd workspace find -i some-new-id -k local_filename -k pageId`.

Demonstrably, this now fails at the following assertion:
```
  File "/data/ocr-d/ocrd_all/venv38/bin/ocrd", line 8, in <module>
    sys.exit(cli())
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/core.py", line 1157, in __call__
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/core.py", line 1688, in invoke
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/core.py", line 1688, in invoke
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/core.py", line 1434, in invoke
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/core.py", line 783, in invoke
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/decorators.py", line 92, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/click/core.py", line 783, in invoke
  File "/data/ocr-d/ocrd_all/core/src/ocrd/cli/workspace.py", line 484, in workspace_find
    pages = workspace.mets.get_physical_pages(for_fileIds=fileIds)
  File "/data/ocr-d/ocrd_all/core/src/ocrd_models/ocrd_mets.py", line 696, in get_physical_pages
    assert for_fileIds # at this point we know for_fileIds is set, assert to convince pyright
AssertionError
```

So it seems, the assertion is not warranted, and pyright was in fact right not to be convinced :)

### LINKED ISSUE #1195: Invalid structMap produced
This might be a bug in ocrd-cis actually, so beware.

We encountered a number of problems elsewhere due to an invalid physical structMap. Here, I managed to reproduce with the latest ocrd:all/maximum Docker image, with the following steps:

0. Starting with the workspace here: https://ub-backup.bib.uni-mannheim.de/~stweil/quiver-benchmark/workflows/workspaces/reichsanzeiger_random_selected_pages_ocr/data/reichsanzeiger_random/
1. I removed all filegroups except OCR-D-IMG and OCR-D-GT-SEG-LINE, using `ocrd workspace remove-group -rf`. → After this, the structMap is **OK**!
2. Then I ran `ocrd-cis-ocropy-binarize -I OCR-D-IMG -O OCR-D-BIN` → After this, the structMap is **INVALID**

Invalid structMap (multiple divs with same ID) after step 2, shortened to one physical page for emphasis:

```xml
  <mets:structMap TYPE="PHYSICAL">
    <mets:div TYPE="physSequence">
      <mets:div TYPE="page" ID="P_1879_45_0344">
        <mets:fptr FILEID="OCR-D-IMG_1879_45_0344"/>
        <mets:fptr FILEID="OCR-D-GT-SEG-LINE_1879_45_0344"/>
      </mets:div>

      <!-- ... >

      <mets:div TYPE="page" ID="P_1879_45_0344">
        <mets:fptr FILEID="OCR-D-BIN_1879_45_0344.IMG-BIN"/>
      </mets:div>
  
       <!-- ... >

    </mets:div>
  </mets:structMap>
```

(I'll upload the full data in the comments)

This causes all kind of breakage all over the place.

What I didn't check yet: if this only breaks with ocrd_cis, maybe @bertsky can share his debugging efforts here. I first had the impression that this breaks with `add` too, but as I had tried to reproduce a problem encountered by @stweil in OCR-D/quiver-benchmarks#22 it could have always been in ocrd_cis (specific workflow uses this as first step) and I could have easily confused something.

bertsky:
Fixes #1192
Fixes #1195

### DIFF
--- a/src/ocrd_models/ocrd_mets.py
+++ b/src/ocrd_models/ocrd_mets.py
@@ -4,7 +4,7 @@ API to METS
 from datetime import datetime
 import re
 from lxml import etree as ET
-from typing import Dict, Optional
+from typing import Callable, Dict, Iterator, List, Optional, Union
 
 from ocrd_utils import (
     getLogger,
@@ -61,7 +61,7 @@ class OcrdMets(OcrdXmlDocument):
     _fptr_cache : Dict[str, Dict[str, ET._Element]]
 
     @staticmethod
-    def empty_mets(now=None, cache_flag=False):
+    def empty_mets(now : Optional[str] = None, cache_flag : bool = False):
         """
         Create an empty METS file from bundled template.
         """
@@ -72,7 +72,7 @@ class OcrdMets(OcrdXmlDocument):
         tpl = tpl.replace('{{ NOW }}', '%s' % now)
         return OcrdMets(content=tpl.encode('utf-8'), cache_flag=cache_flag)
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
         """
         """
         super(OcrdMets, self).__init__(**kwargs)
@@ -91,14 +91,14 @@ class OcrdMets(OcrdXmlDocument):
             self._initialize_caches()
             self._refresh_caches()
 
-    def __str__(self):
+    def __str__(self) -> str:
         """
         String representation
         """
         return 'OcrdMets[cached=%s,fileGrps=%s,files=%s]' % (
         self._cache_flag, self.file_groups, list(self.find_files()))
 
-    def _fill_caches(self):
+    def _fill_caches(self) -> None:
         """
         Fills the caches with fileGrps and FileIDs
         """
@@ -148,13 +148,13 @@ class OcrdMets(OcrdXmlDocument):
         # log.info("Len of page_cache: %s" % len(self._page_cache[METS_PAGE_DIV_ATTRIBUTE.ID]))
         # log.info("Len of fptr_cache: %s" % len(self._fptr_cache))
 
-    def _initialize_caches(self):
+    def _initialize_caches(self) -> None:
         self._file_cache = {}
         # NOTE we can only guarantee uniqueness for @ID and @ORDER
         self._page_cache = {k : {} for k in METS_PAGE_DIV_ATTRIBUTE}
         self._fptr_cache = {}
 
-    def _refresh_caches(self):
+    def _refresh_caches(self) -> None:
         if self._cache_flag:
             self._initialize_caches()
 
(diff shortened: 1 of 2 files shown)


src/ocrd_models/ocrd_mets.py
```diff
@@ -883,8 +885,11 @@ def physical_pages_labels(self):
         return {div.get('ID'): (div.get('ORDER', None), div.get('ORDERLABEL', None), div.get('LABEL', None))
                 for div in divs}
 
-    def merge(self, other_mets, force=False, fileGrp_mapping=None, fileId_mapping=None, pageId_mapping=None,
-              after_add_cb=None, **kwargs):
+    def merge(self, other_mets, force : bool = False,
```
kba, line 888:
```suggestion
    def merge(self, other_mets : 'OcrdMets', force : bool = False, 
```

You can reference classes not yet defined by quoting them.

src/ocrd_models/ocrd_mets.py
```diff
@@ -781,7 +781,7 @@ def set_physical_page_for_file(self, pageId : str, ocrd_file : OcrdFile,
                 self._page_cache[METS_PAGE_DIV_ATTRIBUTE.ID][pageId] = el_pagediv
                 # Create a new entry in the fptr cache and 
                 # assign an empty dictionary to hold the fileids
-                self._fptr_cache[pageId] = {}
+                self._fptr_cache.setdefault(pageId, {})
```
kba, line 784: eek, thx for spotting

kba (approved):
will release right away.

bertsky:
Ouch! I still see the problem after installing this (cannot reproduce my own fix anymore)...

bertsky:
> Problem is really in ocrd_all – somehow, editable and non-editable installation of core got mixed...

Reason is simply that ocrd/core installs via normal site-packages, whereas ocrd/all on top of that installs core editable (under /build/core) – so in these images we have _both_ versions next to each other.

bertsky:
> Reason is simply that ocrd/core installs via normal site-packages, whereas ocrd/all on top of that installs core editable (under /build/core) – so in these images we have _both_ versions next to each other.

https://github.com/OCR-D/ocrd_all/issues/416

PR #1200, opened 2024-03-12
METS server: don't cache the session to avoid broken connections

kba:


MehmedGIT (approved):

MehmedGIT:
I have just got another idea. Maybe it's a good try obtaining the session a few times before complete fail - say 3? Or configurable with an env variable and setting the default to 3.

PR #1205, opened 2024-04-02
update actions

rettinghaus:
This PR adds Python 3.12 to the test runs and updates the used GH actions.

PR #1206, opened 2024-04-10
update actions

rettinghaus:
This PR adds Python 3.12 to the test runs and updates the used GH actions. 
Supersedes #1205


.github/workflows/unit-test.yml
```diff
@@ -23,18 +23,19 @@ jobs:
           - '3.9'
           - '3.10'
           - '3.11'
+          - '3.12'
         os:
           - ubuntu-22.04
           - ubuntu-20.04
-          # - macos-latest
+          - macos-latest
```
stweil, line 30: Did you test the action with macos-latest?

stweil (approved):
The updates look good and are required to fix the CI annotations. Thank you!

kba (approved):
Thanks, we should use the latest versions of the actions.

stweil:
Citing https://peps.python.org/pep-0632/:

> The distutils module [[1]](https://peps.python.org/pep-0632/#id10) has for a long time recommended using the setuptools package [[2]](https://peps.python.org/pep-0632/#id11) instead. Setuptools has recently integrated a complete copy of distutils and is no longer dependent on the standard library [[3]](https://peps.python.org/pep-0632/#id12). Pip has been silently replacing distutils with setuptools when installing packages for a long time already, and the distutils documentation has stated that it is being phased out since 2014 (or earlier). It is time to remove it from the standard library.

core still uses distutils. Therefore tests with Python 3.12 now fail.

kba:
> Citing https://peps.python.org/pep-0632/:
> 
> > The distutils module [[1]](https://peps.python.org/pep-0632/#id10) has for a long time recommended using the setuptools package [[2]](https://peps.python.org/pep-0632/#id11) instead. Setuptools has recently integrated a complete copy of distutils and is no longer dependent on the standard library [[3]](https://peps.python.org/pep-0632/#id12). Pip has been silently replacing distutils with setuptools when installing packages for a long time already, and the distutils documentation has stated that it is being phased out since 2014 (or earlier). It is time to remove it from the standard library.
> 
> core still uses distutils. Therefore tests with Python 3.12 now fail.

I have a branch which removes distutils. The `find_executable` replacement is straightforward, I still need to test the `copy_tree` replacement.

stweil:
@rettinghaus, why did you remove Python 3.12 in your latest commit? I think it is good to have a CI test with that version.

stweil:
@kba, I suggest to merge this pull request (without the last commit which removed Python 3.12 from the test matrix).

rettinghaus:
@stweil rebased and dropped last commit.

stweil (approved):

PR #1207, opened 2024-04-14
Remove old code for unsupported Python 3.7

stweil:
Since 27 June 2023, Python 3.8 is the oldest supported version of Python.


kba (approved):
Yes, time to let 3.7 be.

PR #1208, opened 2024-04-15
Fix some typos and help texts

stweil:
Most typos were found by `codespell` and `typos`.

### DIFF
--- a/src/ocrd_models/ocrd_mets.py
+++ b/src/ocrd_models/ocrd_mets.py
@@ -267,8 +267,8 @@ class OcrdMets(OcrdXmlDocument):
             local_filename (string) : ``@xlink:href`` local/cached filename of ``mets:Flocat`` of ``mets:file``
             mimetype (string) : ``@MIMETYPE`` of ``mets:file``
             local (boolean) : Whether to restrict results to local files in the filesystem
-            include_fileGrp (list[str]) : Whitelist of allowd file groups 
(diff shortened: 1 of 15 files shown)


src/ocrd_models/ocrd_mets.py
```diff
@@ -267,8 +267,8 @@ def find_files(
             local_filename (string) : ``@xlink:href`` local/cached filename of ``mets:Flocat`` of ``mets:file``
             mimetype (string) : ``@MIMETYPE`` of ``mets:file``
             local (boolean) : Whether to restrict results to local files in the filesystem
-            include_fileGrp (list[str]) : Whitelist of allowd file groups 
-            exclude_fileGrp (list[str]) : Blacklist of disallowd file groups 
+            include_fileGrp (list[str]) : List of allowed file groups
```
kba, line 270: Is this because of white/black being problematic as terms? Because if it is just that `whitelist`/`blacklist` are not genuine dictionary-listed English terms, I'd prefer to keep them.

kba (changes requested):

src/ocrd_models/ocrd_mets.py
stweil: Initially the lines were changed because of a typo, but while fixing it I decided to avoid the unnecessary color attributes as well because they are problematic for some people.

stweil: Spell checkers like `codespell` and `typos` do not complain about colored lists.

PR #1210, opened 2024-04-16
use importlib_resources for <= 3.9, not only <= 3.8, OCR-D/ocrd_froc#10

kba:


kba:
Already merged (commit 2c24cce7)

PR #1217, opened 2024-04-25
bashlib input-files: apply download_file on each input_file

### LINKED ISSUE #1216: bashlib input_files: ensure download_file (as in all Pythonic processors)
All of our processors written in Python use a `Workspace.download_file(input_file)` in their processing loop. This ensures the file is available locally, even if it was still a URL (saving it under a reproducable temporary path).

Unfortunately, our bashlib processors have no chance to get that behaviour: the `ocrd workspace find --download`  would inevitably persist the downloaded file, which is perhaps not entirely wrong, but different from the Python processors. Regardless, it's not what we do in ocrd_olena, ocrd_pagetopdf, ocrd_fileformat, ocrd_im6convert etc.

Hence, if the input fileGrp is entirely remote, we only get messages like this:
```
ERROR ocrd.ocrd-olena-binarize - input file ID=FILE_0024_DEFAULT (pageId=PHYS_0024 MIME=image/jpg) is not on disk
```

The result would be a successful run without actual output fileGrp:
```
Exception: Invalid state: expected output file group 'OCR-D-BIN' not in METS (despite processor success)
```

Now, the solution I propose is simple: have `ocrd bashlib input-files` (which does have access to `Workspace.download_file(input_file)`) do the job!

bertsky:
fixes #1216

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -3,6 +3,10 @@ checks:
 
 build:
   image: default-bionic
+  environment:
+    python:
+      version: 3.8.2
+      virtualenv: true
   nodes:
     analysis:
       dependencies:
--- a/src/ocrd/cli/bashlib.py
+++ b/src/ocrd/cli/bashlib.py
@@ -114,6 +114,10 @@ def bashlib_input_files(**kwargs):
                           input_file_grp=kwargs['input_file_grp'],
                           output_file_grp=kwargs['output_file_grp'])
     for input_files in processor.zip_input_files(mimetype=None, on_error='abort'):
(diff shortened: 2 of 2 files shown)


bertsky:
I also tried to fix the Scrutinizer setup, since it [complained](https://scrutinizer-ci.com/g/OCR-D/core/inspections/be898927-b361-4469-a2b0-bc3403923c77) that Python 3.7 is no longer supported.

(Indeed, the `image: default-bionic` seems too old, but `default-jammy` too new. So I chose the pyenv path, which seems to work.)

kba (approved):

Issue #1212, opened 2024-04-16
Replace deprecated / removed Python module distutils


stweil:
Citing https://peps.python.org/pep-0632/:

> The distutils module [[1]](https://peps.python.org/pep-0632/#id10) has for a long time recommended using the setuptools package [[2]](https://peps.python.org/pep-0632/#id11) instead. Setuptools has recently integrated a complete copy of distutils and is no longer dependent on the standard library [[3]](https://peps.python.org/pep-0632/#id12). Pip has been silently replacing distutils with setuptools when installing packages for a long time already, and the distutils documentation has stated that it is being phased out since 2014 (or earlier). It is time to remove it from the standard library.

core still uses distutils. Therefore tests with Python 3.12 now fail, and older Python versions can trigger a deprecation warning (cf. issue #850).

stweil:
See also https://github.com/OCR-D/spec/blob/master/decisions.md#supported-python-versions: "We will start to support 3.12 in Q2 2023 (:warning: won't have distutils anymore)".

stweil:
Other processors / components which also still use distutils:
- ocrd_anybaseocr
- ocrd_olena/repo/olena (?)
- opencv-python/multibuild
- opencv-python/opencv

mikegerber:
This also seems to break native Windows support of all things. They have Python 3.12.

(Never used, just tested if dinglehopper would work with the Microsoft Store's Python, because users)

mikegerber:
Replacing `from distutils.spawn import find_executable as which` with `from shutil import which` in `os.py` seems to fix my immediate problem. However I'm not sure if they have the same semantics.

Other instances of distutils usage:
- More `find_executable` (3 in total)
- `from distutils.dir_util import copy_tree` (Also possibly replacable by shutil?) (1 in total)

mikegerber:
> Replacing `from distutils.spawn import find_executable as which` with `from shutil import which` in `os.py` seems to fix my immediate problem. However I'm not sure if they have the same semantics.

`'from setuptools._distutils.spawn import find_executable` (+ Adding `setuptools` to requirements, if necessary) may be the safer way for now.

>     * `from distutils.dir_util import copy_tree` (Also possibly replacable by shutil?) (1 in total)
 
`from setuptools._distutils.dir_util import copy_tree`

stweil:
See also https://github.com/OCR-D/ocrd_anybaseocr/pull/105.

stweil:
[...]

PR #1219, opened 2024-05-03
Drop distutils

kba:


stweil:
This will fix issue #1212.

PR #1222, opened 2024-05-07
METS Server: improve logging

bertsky:
This fixes some cosmetic issues for the METS Server's log output:
- we don't need the access_log – it's quite verbose and we can always just enable the workspace debug logging instead
- we should not allow Uvicorn to initialise its own log config – but have it use our preconfigured logging facility for consistency
- the loggers should all have the same name, and to be configurable, the URL should not be part of the top-level name
- removing the UDS socket on shutdown is the normal case – no need for a warning there


MehmedGIT (approved):
Looks good.

PR #1227, opened 2024-05-24
OcrdFile: fix removing url

### LINKED ISSUE #1226: OcrdFile: cannot remove .url
If I set `file.url = None`, then OcrdFile will ignore it.

bertsky:
fixes #1226


kba (approved):
Thanks, lxml gotcha.

PR #1236, opened 2024-06-06
Workspace clean

bertsky:


tests/cli/test_workspace.py
```diff
@@ -291,6 +291,23 @@ def test_prune_files(self):
             ws2 = self.resolver.workspace_from_url(join(tempdir, 'ws', 'mets.xml'))
             self.assertEqual(len(ws2.mets.find_all_files()), 29)
 
+    def test_clean(self):
+        with TemporaryDirectory() as tempdir:
+            copytree(assets.path_to('SBB0000F29300010000/data'), join(tempdir, 'ws'))
+
+            ws1 = self.resolver.workspace_from_url(join(tempdir, 'ws', 'mets.xml'))
+            self.assertEqual(len(ws1.mets.find_all_files(local_only=True)), 29)
```
kba, line 299:
```suggestion
            assert len(ws1.mets.find_all_files(local_only=True)) == 29
```

just fyi, pytest supports normal `assert` statements, which is a bit more readable than the unittests/junit like methods.

kba (approved):
Nice, I like the analogy to `git clean` and `clean` complementing `prune`. Perhaps we could also implement `-d` to toggle the `rmdir` behavior? OTOH, I always use `-d` with `git clean` so it might not be necessary

PR #1238, opened 2024-06-07
parameter decorator: resolve path names as resources

### LINKED ISSUE #930: parameter decorator: handle preset files as resources
When we drafted the resource handling, one of the use cases was installing preset files, i.e. preconfigured (if incomplete) json parameter sets that have a discernable name and can be used as value for `--parameter|-p` but without the directory path.

That latter part is still missing. So we still need to modify … https://github.com/OCR-D/core/blob/0713e0e1c2fd11242957c3a51c974a1048dba997/ocrd/ocrd/decorators/parameter_option.py#L7-L14 … such that known resource filenames are tried first.

bertsky:
Fixes #930

Replaces #969


kba (approved):
LGTM, thanks for adapting your previous efforts and providing/adapting tests too.

PR #1243, opened 2024-06-26
ocrd process via METS server option

### LINKED ISSUE #1241: ocrd process: add METS Server CLI option
We have `-U path` in the processor CLIs now, but not for `ocrd process` yet.

bertsky:
fixes #1241

### DIFF
--- a/src/ocrd/cli/process.py
+++ b/src/ocrd/cli/process.py
@@ -19,14 +19,15 @@ from ..decorators import ocrd_loglevel
 @click.command('process')
 @ocrd_loglevel
 @click.option('-m', '--mets', help="METS to process", default=DEFAULT_METS_BASENAME)
+@click.option('-U', '--mets-server-url', help="TCP host URI or UDS path of METS server")
 @click.option('-g', '--page-id', help="ID(s) of the pages to process")
 @click.option('--overwrite', is_flag=True, default=False, help="Remove output pages/images if they already exist")
 @click.argument('tasks', nargs=-1, required=True)
-def process_cli(log_level, mets, page_id, tasks, overwrite):
+def process_cli(log_level, mets, mets_server_url, page_id, tasks, overwrite):
     """
     Process a series of tasks
     """
     initLogging()
     log = getLogger('ocrd.cli.process')
-    run_tasks(mets, log_level, page_id, tasks, overwrite)
+    run_tasks(mets, log_level, page_id, tasks, overwrite=overwrite, mets_server_url=mets_server_url)
     log.info("Finished")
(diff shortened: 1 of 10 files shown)


bertsky:
I now have some realistic tests as well. 

But before I push these: Please help me understand a detail of the current implementation @kba: In …
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/task_sequence.py#L120
…the workspace instantiated from the `--mets` argument does not contain a `dst_dir` kwarg, which means that it will merely be a clone in a temporary working directory which then via `run_cli` makes its way into the individual processor CLIs …
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/processor/helpers.py#L198-L199

So how could this ever have worked?

bertsky:
cf. da7e960 i.e. need for `resolve_mets_arguments` is what I'm talking about.

kba:
> I now have some realistic tests as well.
> 
> But before I push these: Please help me understand a detail of the current implementation @kba: In …
> 
> https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/task_sequence.py#L120
> 
> …the workspace instantiated from the `--mets` argument does not contain a `dst_dir` kwarg, which means that it will merely be a clone in a temporary working directory which then via `run_cli` makes its way into the individual processor CLIs …
> 
> https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/processor/helpers.py#L198-L199
> 
> So how could this ever have worked?

TBH, I just never encounter the use case of passing a HTTP URL to a METS file directly to any of the CLI that accept `--mets` nowadays, I always clone before and pass a local METS file as `--mets`, in which case, `dst_dir` is derived from `Path(mets_url).parent`.

> cf. https://github.com/OCR-D/core/commit/da7e96076f192ba506a70208501ddeeaa167bf63 i.e. need for resolve_mets_arguments is what I'm talking about.

```diff
-    workspace = resolver.workspace_from_url(mets, mets_server_url=mets_server_url)
+    workdir, mets, basename, _ = resolver.resolve_mets_arguments(None, mets, None)
+    workspace = resolver.workspace_from_url(mets, workdir, mets_basename=basename,
+                                            mets_server_url=mets_server_url)
```

What is the intended effect here? Because `resolve_mets_arguments(None, mets)` will raise a ValueError if `mets` is a remote (HTTP) URL. If it is a local file, it will also return `Path(mets).parent`, so I don't see how it is better than just calling `workspace_from_url` directly.

But it might very well be that my assumptions are wrong - under what circumstances are the temporary directories a problem?

kba:
> Because that function needs a `dst_dir` argument or will only yield a partial clone.

I don't have a good explanation  right now, I tend to run the tools from the workspace directory so I may not have noticed it because of that. Will investigate further, but since the PR fixes that issue at least where you noticed it and it adds necessary functionality to `ocrd process`, let's merge.

PR #1244, opened 2024-07-03
Remove unused socket file on METS Server startup

joschrew:
The METS Server does not start if its desired socket file already exists. When experimenting with ocrd-network the socket file occasionally is not deleted on mets-server shutdown in some edge-cases e.g. killing the processing-server manually. Workaround is to just delete the socket file manually which is cumbersome.
This change tests if the socket file exists and deletes it in case it is not used by any socket.

### DIFF
--- a/src/ocrd/mets_server.py
+++ b/src/ocrd/mets_server.py
@@ -555,5 +555,3 @@ def is_socket_in_use(socket_path):
             return False
         client.close()
         return True
-    else:
-        return False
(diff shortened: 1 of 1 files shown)


kba (approved):

src/ocrd/mets_server.py
```diff
@@ -541,3 +544,16 @@ async def add_file(
...
+    if Path(socket_path).exists():
+        client = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        try:
+            client.connect(socket_path)
+        except OSError:
+            return False
+        client.close()
+        return True
+    else:
+        return False
```
kba, line 559:
```suggestion
```

Redundant, function returns `None` by default.

Issue #1237, opened 2024-06-06
resmgr download directories


bertsky:
If I do

    ocrd resmgr download -a -n ~/.local/share/ocrd-resources/ocrd-origami-segment/bbz ocrd-origami-segment bbz2

then I end up with

```
  File "/ocrd_all/core/build/__editable__.ocrd-2.65.0-py3-none-any/ocrd/cli/resmgr.py", line 155, in download
    fpath = resmgr.download(
  File "/ocrd_all/core/build/__editable__.ocrd-2.65.0-py3-none-any/ocrd/resource_manager.py", line 314, in download
    self._copy_impl(url, fpath, progress_cb)
  File "/ocrd_all/core/build/__editable__.ocrd-2.65.0-py3-none-any/ocrd/resource_manager.py", line 257, in _copy_impl
    with open(child_dst, 'wb') as f_out, open(child, 'rb') as f_in:
IsADirectoryError: [Errno 21] Is a directory: '~/.local/share/ocrd-resources/ocrd-origami-segment/bbz/v3'
```

See [this older comment](https://github.com/OCR-D/core/issues/691#issuecomment-1038152665)

bertsky:
And [this older review](https://github.com/bertsky/core/pull/3#pullrequestreview-887874649) says the culprit is here:

https://github.com/OCR-D/core/blob/01955620274e461a8470b37404ef1e9996ca3221/src/ocrd/resource_manager.py#L249-L265

This is not recursive (enough): it cannot copy with subdirectories.

PR #1248, opened 2024-07-04
fix: resmgr download directories #1237

MehmedGIT:
A fix for #1237. The resource manager should now properly copy the directories recursively.


MehmedGIT:
IMO, we should also move the copy file and copy directory methods to `ocrd_utils/os.py`

bertsky (approved):
Perfect – thanks a lot!

kba (approved):

Issue #1218, opened 2024-04-26
resmgr list-available * includes non-processors (again)


bertsky:
This seems to be a regression: the dynamic resource discovery now attempts to query non-processors like `ocrd-import` and `ocrd-make` and `ocrd-cis-data`, which produces ugly error messages.

PR #1249, opened 2024-07-04
fix: resmgr list-available * includes non-processors (again) #1218

MehmedGIT:
A partial fix for #1218.

There are still some annoying Tensorflow prints. Unfortunately, these cannot be disabled from the core or I am unaware how.
```sh
(venv38-core) mm@MM-Notebook:~/repos/core$ ocrd resmgr list-available
2024-07-04 12:17:53.701854: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F AVX512_VNNI FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2024-07-04 12:17:53.778978: I tensorflow/core/util/port.cc:104] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2024-07-04 12:17:53.781393: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /home/mm/venv38-core/lib/python3.8/site-packages/cv2/../../lib64:
2024-07-04 12:17:53.781404: I tensorflow/compiler/xla/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
2024-07-04 12:17:54.178713: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer.so.7'; dlerror: libnvinfer.so.7: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /home/mm/venv38-core/lib/python3.8/site-packages/cv2/../../lib64:
2024-07-04 12:17:54.178761: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer_plugin.so.7'; dlerror: libnvinfer_plugin.so.7: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /home/mm/venv38-core/lib/python3.8/site-packages/cv2/../../lib64:
2024-07-04 12:17:54.178766: W tensorflow/compiler/tf2tensorrt/utils/py_utils.cc:38] TF-TRT Warning: Cannot dlopen some TensorRT libraries. If you would like to use Nvidia GPU with TensorRT, please make sure the missing libraries mentioned above are installed properly.
2024-07-04 12:17:59.115332: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F AVX512_VNNI FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2024-07-04 12:17:59.191225: I tensorflow/core/util/port.cc:104] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2024-07-04 12:17:59.193509: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /home/mm/venv38-all/lib/python3.8/site-packages/cv2/../../lib64:/home/mm/venv38-core/lib/python3.8/site-packages/cv2/../../lib64:
2024-07-04 12:17:59.193522: I tensorflow/compiler/xla/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
2024-07-04 12:17:59.624514: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer.so.7'; dlerror: libnvinfer.so.7: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /home/mm/venv38-all/lib/python3.8/site-packages/cv2/../../lib64:/home/mm/venv38-core/lib/python3.8/site-packages/cv2/../../lib64:
2024-07-04 12:17:59.624560: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer_plugin.so.7'; dlerror: libnvinfer_plugin.so.7: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /home/mm/venv38-all/lib/python3.8/site-packages/cv2/../../lib64:/home/mm/venv38-core/lib/python3.8/site-packages/cv2/../../lib64:
2024-07-04 12:17:59.624565: W tensorflow/compiler/tf2tensorrt/utils/py_utils.cc:38] TF-TRT Warning: Cannot dlopen some TensorRT libraries. If you would like to use Nvidia GPU with TensorRT, please make sure the missing libraries mentioned above are installed properly.
WARNING:tensorflow:From /home/mm/venv38-all/lib/python3.8/site-packages/tensorflow/python/compat/v2_compat.py:107: disable_resource_variables (from tensorflow.python.ops.variable_scope) is deprecated and will be removed in a future version.
Instructions for updating:
non-resource variables are not supported in the long term
...
```

### DIFF
--- a/src/ocrd/resource_manager.py
+++ b/src/ocrd/resource_manager.py
@@ -102,7 +102,7 @@ class OcrdResourceManager:
             for exec_dir in environ['PATH'].split(':'):
                 for exec_path in Path(exec_dir).glob(f'{executable}'):
                     self.log.debug(f"Inspecting '{exec_path} --dump-json' for resources")
-                    ocrd_tool = get_ocrd_tool_json(exec_path, True)
+                    ocrd_tool = get_ocrd_tool_json(exec_path)
                     for resdict in ocrd_tool.get('resources', ()):
                         if exec_path.name not in database:
                             database[exec_path.name] = []
--- a/src/ocrd/resource_manager.py
+++ b/src/ocrd/resource_manager.py
@@ -103,8 +103,7 @@ class OcrdResourceManager:
             for exec_dir in environ['PATH'].split(':'):
                 for exec_path in Path(exec_dir).glob(f'{executable}'):
                     if not exec_path.name.startswith('ocrd-'):
-                        self.log.debug(f"Not in 'ocrd-*' format, skipping '{exec_path}'")
-                        continue
+                        self.log.warning(f"OCR-D processor executable '{exec_path}' has no 'ocrd-' prefix")
                     if exec_path.name in skip_executables:
                         self.log.debug(f"No dump json available, skipping '{exec_path}'")
                         continue
--- a/src/ocrd/resource_manager.py
+++ b/src/ocrd/resource_manager.py
@@ -102,7 +102,7 @@ class OcrdResourceManager:
             for exec_dir in environ['PATH'].split(':'):
                 for exec_path in Path(exec_dir).glob(f'{executable}'):
                     self.log.debug(f"Inspecting '{exec_path} --dump-json' for resources")
-                    ocrd_tool = get_ocrd_tool_json(exec_path)
+                    ocrd_tool = get_ocrd_tool_json(exec_path, True)
                     for resdict in ocrd_tool.get('resources', ()):
                         if exec_path.name not in database:
                             database[exec_path.name] = []
--- a/src/ocrd/resource_manager.py
+++ b/src/ocrd/resource_manager.py
@@ -105,7 +105,7 @@ class OcrdResourceManager:
                     if not exec_path.name.startswith('ocrd-'):
                         self.log.warning(f"OCR-D processor executable '{exec_path}' has no 'ocrd-' prefix")
                     if exec_path.name in skip_executables:
-                        self.log.debug(f"No dump json available, skipping '{exec_path}'")
+                        self.log.debug(f"Not an OCR-D processor CLI, skipping '{exec_path}'")
                         continue
                     self.log.debug(f"Inspecting '{exec_path} --dump-json' for resources")
                     ocrd_tool = get_ocrd_tool_json(exec_path)
--- a/src/ocrd/resource_manager.py
+++ b/src/ocrd/resource_manager.py
@@ -99,8 +99,15 @@ class OcrdResourceManager:
         if not executable:
             return database.items()
         if dynamic:
+            skip_executables = ["ocrd-cis-data", "ocrd-import", "ocrd-make"]
             for exec_dir in environ['PATH'].split(':'):
                 for exec_path in Path(exec_dir).glob(f'{executable}'):
+                    if not exec_path.name.startswith('ocrd-'):
+                        self.log.debug(f"Not in 'ocrd-*' format, skipping '{exec_path}'")
+                        continue
+                    if exec_path.name in skip_executables:
+                        self.log.debug(f"No dump json available, skipping '{exec_path}'")
+                        continue
                     self.log.debug(f"Inspecting '{exec_path} --dump-json' for resources")
                     ocrd_tool = get_ocrd_tool_json(exec_path)
                     for resdict in ocrd_tool.get('resources', ()):
--- a/src/ocrd_utils/os.py
+++ b/src/ocrd_utils/os.py
@@ -75,38 +75,27 @@ def unzip_file_to_dir(path_to_zip, output_directory):
     z.close()
 
 @lru_cache()
-def get_ocrd_tool_json(executable, skip_error_print: bool = False):
+def get_ocrd_tool_json(executable):
     """
     Get the ``ocrd-tool`` description of ``executable``.
     """
     ocrd_tool = {}
     executable_name = Path(executable).name
-    # TODO: Potentially these executables should be moved as a constant somewhere else
-    skip_executables = ["ocrd-cis-data", "ocrd-import", "ocrd-make"]
-    if executable_name not in skip_executables:
-        try:
-            ocrd_tool = loads(run([executable, '--dump-json'], stdout=PIPE).stdout)
-        except (JSONDecodeError, OSError) as e:
-            if not skip_error_print:
-                getLogger('ocrd.utils.get_ocrd_tool_json').error(f'{executable} --dump-json produced invalid JSON: {e}')
-        except Exception as e:
-            if not skip_error_print:
-                getLogger('ocrd.utils.get_ocrd_tool_json').error(f'{executable} --dump-json failed: {e}')
+    try:
+        ocrd_tool = loads(run([executable, '--dump-json'], stdout=PIPE).stdout)
+    except (JSONDecodeError, OSError) as e:
+        getLogger('ocrd.utils.get_ocrd_tool_json').error(f'{executable} --dump-json produced invalid JSON: {e}')
     if 'resource_locations' not in ocrd_tool:
         ocrd_tool['resource_locations'] = ['data', 'cwd', 'system', 'module']
     return ocrd_tool
 
 @lru_cache()
-def get_moduledir(executable, skip_error_print: bool = False):
+def get_moduledir(executable):
     moduledir = None
     try:
         moduledir = run([executable, '--dump-module-dir'], encoding='utf-8', stdout=PIPE).stdout.rstrip('\n')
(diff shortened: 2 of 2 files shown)


MehmedGIT:
This PR may require some adaptations to #1250.

src/ocrd_utils/os.py
```diff
@@ -75,27 +75,38 @@ def unzip_file_to_dir(path_to_zip, output_directory):
...
     """
+    ocrd_tool = {}
     executable_name = Path(executable).name
-    try:
-        ocrd_tool = loads(run([executable, '--dump-json'], stdout=PIPE).stdout)
-    except (JSONDecodeError, OSError) as e:
-        getLogger('ocrd.utils.get_ocrd_tool_json').error(f'{executable} --dump-json produced invalid JSON: {e}')
-        ocrd_tool = {}
+    # TODO: Potentially these executables should be moved as a constant somewhere else
+    skip_executables = ["ocrd-cis-data", "ocrd-import", "ocrd-make"]
```
bertsky, line 85: Why is this necessary? Why here?

AFAICS the cause of these false positives is simply the `ocrd-*` default glob …
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/cli/resmgr.py#L48-L54
… combined with general search in PATH …
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/resource_manager.py#L100-L104

So it would be more fitting to suppress these in resmgr.

src/ocrd_utils/os.py
```diff
@@ -75,27 +75,38 @@ def unzip_file_to_dir(path_to_zip, output_directory):
...
+    if executable_name not in skip_executables:
+        try:
+            ocrd_tool = loads(run([executable, '--dump-json'], stdout=PIPE).stdout)
+        except (JSONDecodeError, OSError) as e:
+            if not skip_error_print:
+                getLogger('ocrd.utils.get_ocrd_tool_json').error(f'{executable} --dump-json produced invalid JSON: {e}')
+        except Exception as e:
+            if not skip_error_print:
+                getLogger('ocrd.utils.get_ocrd_tool_json').error(f'{executable} --dump-json failed: {e}')
```
bertsky, line 94: IMO it's not a good pattern to suppress these warnings. Either the tool exists – then we _should_ see the error message, or it does not – then we should not even _attempt_ to run it.

The problem in #1218 is that in …
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/resource_manager.py#L42
… the default preinstalled resource database contains 
- outdated processors that we do not use anymore 
- entries that might not in fact be installed currently

So the right thing IMO would be to filter … 
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/resource_manager.py#L85-L87
… by those that can be found in PATH.

bertsky (commented):
clashes with #1250 

also I don't think this is the right approach – see below

src/ocrd_utils/os.py
MehmedGIT: > entries that might not in fact be installed currently

Why should the resmgr be limited by the fact that a processor is locally installed or not? The models should be downloadable regardless. The resmgr does the ocrd tool json dump by running the processor - the main limitation. Hopefully, that will be resolved with #1250. 

For now, I can still filter these though.

src/ocrd_utils/os.py
MehmedGIT: Right, I changed that.

src/ocrd_utils/os.py
bertsky: You're right – this should be independent. (I think I confused this with the other issue, which covers `list_installed` rather than `list_available`.)

Ok, but perhaps instead of running `--dump-json` and `--dump-module-dir` directly (and having to deal with different cases of error), we should first look whether the executable exists and only _then_ attempt something with a potential error message?

MehmedGIT: > Ok, but perhaps instead of running --dump-json and --dump-module-dir directly (and having to deal with different cases of error), we should first look whether the executable exists

The resource manager inside `list_available` executes the `--dump-json` only for executables found in `PATH`. I think an additional filter is required to make sure the executable name follows the `ocrd-*` pattern. The main issue then is when paths have `ocrd-*` inside as false positives such as `ocrd-make`. These should be either explicitly banned in a list or the `--dump-json` should be executed regardless and the error message should be suppressed (you are not in favor of that option based on your comments above).

bertsky: Ah, indeed, `executable` already is from PATH.

Filtering: yes, we [already agreed false positives should be ruled out in resmgr itself](https://github.com/OCR-D/core/pull/1249/files#r1665531809).

But then – at least for `list-available` I don't see any need to suppress warnings in ocrd_utils.

src/ocrd/resource_manager.py
```diff
@@ -99,10 +99,17 @@ def list_available(self, executable=None, dynamic=True, name=None, database=None
         if not executable:
             return database.items()
         if dynamic:
+            skip_executables = ["ocrd-cis-data", "ocrd-import", "ocrd-make"]
             for exec_dir in environ['PATH'].split(':'):
                 for exec_path in Path(exec_dir).glob(f'{executable}'):
+                    if not exec_path.name.startswith('ocrd-'):
+                        self.log.debug(f"Not in 'ocrd-*' format, skipping '{exec_path}'")
+                        continue
```
bertsky, line 107: If the user explicitly requested an executable not starting with `ocrd-`, then IMO we should not skip that (but let it fail).

src/ocrd/resource_manager.py
```diff
@@ -99,10 +99,17 @@ def list_available(self, executable=None, dynamic=True, name=None, database=None
...
+                    if exec_path.name in skip_executables:
+                        self.log.debug(f"No dump json available, skipping '{exec_path}'")
```
bertsky, line 109:
```suggestion
                        self.log.debug(f"Not an OCR-D processor CLI, skipping '{exec_path}'")
```

src/ocrd/resource_manager.py
bertsky: Or perhaps issue a warning?
```suggestion
                    if not exec_path.name.startswith('ocrd-'):
                        self.log.warning(f"OCR-D processor executable '{exec_path}' has no 'ocrd-' prefix")
```

bertsky (approved):

MehmedGIT:
No longer clashes with #1250.

src/ocrd/resource_manager.py
```diff
@@ -104,6 +104,7 @@ def list_available(self, executable=None, dynamic=True, name=None, database=None
                 for exec_path in Path(exec_dir).glob(f'{executable}'):
                     if not exec_path.name.startswith('ocrd-'):
                         self.log.warning(f"OCR-D processor executable '{exec_path}' has no 'ocrd-' prefix")
+                        continue
```
bertsky, line 107: I thought we agreed this should **NOT** be skipped?

bertsky: https://github.com/OCR-D/core/pull/1249#discussion_r1665696308

MehmedGIT: Sorry, you're right, reverting.

PR #1246, opened 2024-07-03
fix: Resource Manager: Unlink archives after extracting #1245

MehmedGIT:
~A fix for https://github.com/OCR-D/core/issues/1245 and some minor refactoring related to strings.~

- Some minor refactoring over the resource manager.
- A fix for #1201
- If overwrite is not set, the existing entries will be skipped with a warning instead of raising an exception

PR #1258, opened 2024-07-16
tests: disable resmgr --overwrite tests, #1246

kba:


MehmedGIT (approved):
The Scrutinizer still gets stuck on a Yes/No question.

kba:
> The Scrutinizer still gets stuck on a Yes/No question.

Yeah, I tried to fix that with `DEBIAN_FRONTEND=noninteractive` but this won't help with messages about restarting the system. Let's just disable scrutinizer, the unit tests there don't add anything and the pylint stuff, though pretty, we can run ourselves.

PR #1266, opened 2024-08-01
workspace find: print correct output fields

### LINKED ISSUE #1202: workspace find --download prints None for each file
Currently, running `ocrd workspace find --download` will download files, but print `None` for every file it downloaded (instead of the new `local_filename`).

The cause is that `ret_entry` does not get updated after download_file:

https://github.com/OCR-D/core/blob/c5b5580ba5c517e6ee151a5b3a1d8fe1b3ba0d88/src/ocrd/cli/workspace.py#L468-L488

bertsky:
fixes #1202 and a regression that printed a removal notice instead of the requested output fields when undoing downloads


kba (approved):
Thanks

Issue #563, opened 2020-08-13
Resolver.workspace_from_url: usage of clobber_mets not consistent with the docstring


hnesk:
clobber_mets = False does not raise an exception for existing mets.xml ('raise'), but fails silently ('skip')
https://github.com/OCR-D/core/blob/4e11b8e70a2426f3d25466f462db13bf31215919/ocrd/ocrd/resolver.py#L114
see:
https://github.com/OCR-D/core/blob/4e11b8e70a2426f3d25466f462db13bf31215919/ocrd/ocrd/resolver.py#L154

I don't know what would be the right thing to do: Fix the docstring or fix the implementation?

bertsky:
> I don't know what would be the right thing to do: Fix the docstring or fix the implementation?

I'd prefer the latter option.

hnesk:
Oh, I just found this PR https://github.com/OCR-D/core/pull/425#pullrequestreview-346994175

bertsky:
@kba this needs to be revisited – including the discussion after #425.

We still have:
https://github.com/OCR-D/core/blob/987da9bb59cf8ddb5bc9f706d01db7652d092178/src/ocrd/resolver.py#L221

At the very least, the docstring needs to be fixed.

PR #1268, opened 2024-08-06
Raise FileExistsError in Resolver.workspace_from_url if clobber_mets is False

kba:
As laid out in #563, the `clobber_mets` behavior was different from what the docstring said. Instead of raising and Exception if the METS file already existed and `clobber_mets` being `False`, the METS file was silently skipped.

The easiest way would have been to just change the docstring. But the current behavior is problematic because silently skipping downloading/copying the METS XML will lead to hard-to-debug errors when users subsequently work on the wrong METS XML file.

### DIFF
--- a/src/ocrd/resolver.py
+++ b/src/ocrd/resolver.py
@@ -221,7 +221,7 @@ class Resolver():
 
         log.debug("workspace_from_url\nmets_basename='%s'\nmets_url='%s'\nsrc_baseurl='%s'\ndst_dir='%s'",
             mets_basename, mets_url, src_baseurl, dst_dir)
-        self.download_to_directory(dst_dir, mets_url, basename=mets_basename, if_exists='overwrite' if clobber_mets else 'skip')
+        self.download_to_directory(dst_dir, mets_url, basename=mets_basename, if_exists='overwrite' if clobber_mets else 'raise')
 
         workspace = Workspace(self, dst_dir, mets_basename=mets_basename, baseurl=src_baseurl, mets_server_url=mets_server_url)
 
--- a/src/ocrd/resolver.py
+++ b/src/ocrd/resolver.py
@@ -95,12 +95,15 @@ class Resolver():
                 log.debug("Stop early, src_path and dst_path are the same: '%s' (url: '%s')" % (src_path, url))
                 return str(ret)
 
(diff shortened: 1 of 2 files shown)


bertsky:
> The easiest way would have been to just change the docstring. But the current behavior is problematic because silently skipping downloading/copying the METS XML will lead to hard-to-debug errors when users subsequently work on the wrong METS XML file.

I tend to agree, but the discussion in #425 initially outlined a different scenario:

> > I do not understand the use-cases for clobber_mets=False, and I know of no example of it
> The original idea was that you pass a METS URL which is only downloaded the first time it is to be used. Subsequent processors should not overwrite the METS every time they process a workspace.

However, you then followed up with the following explanation, which already speaks to the current line of thought:

> The METS might not be downloaded yet. That's why we introduced the idea of a workspace which was initially just a crutch to make it easier for processors and we were careful not to use it as a first-class concept e.g. in our specs. Instead of handling the intricacies of downloading full/partial representations of the work, they now just had to act on a directory and files. Since then however, workspaces have become the standard way to interact with METS/Images/PAGE and most if not all the facilities to handle non-local data outside ocrd workspace has become obsolete.

So IIUC the initial scenario of _always_ using `workspace_from_url` to get a handle on METS and its contents already went out of the window, and we now have to properly differentiate between _creating_ and _using_ a workspace. Thus the `workspace_from_url` function now entirely falls into the former category, and `clobber_mets=False` (still the default) without a failure/exception does not make sense anymore.

bertsky (approved):
LGTM.

kba:
> So IIUC the initial scenario of _always_ using `workspace_from_url` to get a handle on METS and its contents already went out of the window, and we now have to properly differentiate between _creating_ and _using_ a workspace. Thus the `workspace_from_url` function now entirely falls into the former category, and `clobber_mets=False` (still the default) without a failure/exception does not make sense anymore

Yeah, it is a bit messy due to poor initial planning and yes, the "clone once then work on the local workspace" approach is the default now, the "on demand" approach is largely obsolete. But I'd keep `clobber_mets=False` to avoid introducing a new source of mysterious behavior. I cannot really imagine a scenario where overwriting the METS during clone would be preferable to removing the workspace manually or cloning into a different directory but the flag/kwarg is there if people do need that.

Issue #1265, opened 2024-07-31
ocrd network client processing processor: is non-blocking


bertsky:
When we first discussed the possible client CLI, we kind of agreed to make the typical behaviour of long-running commands blocking (as in the non-network analogue), encapsulating all internal polling or callback mechanics.

But for now the `processor` command is non-blocking – it [just prints the job ID](https://github.com/OCR-D/core/blob/81c0b4c01eb1db54999c5db8feae120df2cc1a7b/src/ocrd_network/cli/client.py#L78-L83).

Do we really want that? So far we have no `status` command, and the user would still need to write their own polling loop.

MehmedGIT:
> But for now the processor command is non-blocking – it [just prints the job ID](https://github.com/OCR-D/core/blob/81c0b4c01eb1db54999c5db8feae120df2cc1a7b/src/ocrd_network/cli/client.py#L78-L83).

That is ideal for non-interactive clients, however, for interactive clients we could also watch the log file of that specific job till the job fails or succeeds. The same goes for the workflow endpoint. Currently, it is possible to poll workflow status from the processing server. Check an example output:

```
{"ocrd-olena-binarize":{"SUCCESS":25},
"ocrd-anybaseocr-crop":{"SUCCESS":24,"FAILED":1},
"ocrd-cis-ocropy-denoise":{"SUCCESS":24,"CANCELLED":1},
"ocrd-cis-ocropy-deskew":{"SUCCESS":24,"CANCELLED":1},
"ocrd-tesserocr-segment-region":{"SUCCESS":24,"CANCELLED":1},
"ocrd-segment-repair":{"SUCCESS":24,"CANCELLED":1},
"ocrd-cis-ocropy-clip":{"SUCCESS":24,"CANCELLED":1},
"ocrd-cis-ocropy-segment":{"SUCCESS":24,"CANCELLED":1},
"ocrd-cis-ocropy-dewarp":{"SUCCESS":24,"CANCELLED":1},
"ocrd-tesserocr-recognize":{"SUCCESS":24,"CANCELLED":1},
"failed-processor-tasks":{"ocrd-anybaseocr-crop":[{"job_id":"64c73535-bbb9-4595-bfc8-7b84b3dd94e7","page_id":"PHYS_0005"}]}}
```
[...]

PR #1269, opened 2024-08-06
Extend the network client

MehmedGIT:
A brief beginning of addressing the points in https://github.com/OCR-D/core/issues/1265:
- Fix `network-integration-test` failing due to a bad env variable
- Ocrd network client:
  - can submit workflow jobs via the CLI
  - can check processing and workflow job statuses
  - can check the list of deployed processing workers/processor servers
  - can check the ocrd tool json of a deployed processor
  - can check the log file of a processing job
- Two new env variables to control ocrd network client behavior: `OCRD_NETWORK_CLIENT_POLLING_SLEEP` and `OCRD_NETWORK_CLIENT_POLLING_TIME`
- Integration tests for the ocrd network client
- Reuse client logic in the processing server integration tests to avoid code duplication


MehmedGIT:
Connections to the client callback URL would be more complicated than in the example @joschrew shared. For the CI/CD pipeline, we need an appropriate host defining and port mapping just like in the other services in the docker-compose file.

src/ocrd_network/cli/client.py
```diff
@@ -1,11 +1,43 @@
...
+        STOP_WAITING_CALLBACK = True
+
+
+class ClientCallbackServer(HTTPServer):
+    """
+    A simple http-server that listens for callbacks from the Processing Server/Worker.
+    """
+    def __init__(self):
+        super().__init__(server_address=("0.0.0.0", 0), RequestHandlerClass=ClientCallbackHandler)
+        self.callback_url = f"http://172.17.0.1:{self.server_address[1]}"
```
kba, line 38: What is the purpose of `self.callback_url` of the callback server? And why does it use the docker host IP but with the same port as the callback server?

Is it so that the (remote) processing server can "find" the local callback server on the network - which would not work with `0.0.0.0`/`localhost`?

In any case we should make this configurable or deduce it from the environment, otherwise it will not work if Docker is set up differently.

Also, I'm wondering how well this scales, the limiting factor is the number of free ports?

kba (commented):
So far so good, appreciate also the test.

We should really find a different mechanism than a global variable. I have proposed a way to at least scope such a variable to the server instance. Why do we need that mechanism anyway? IIUC `callback_server.handle_request` will handle exactly one request with no timeout, which is exactly what we want, is it not?

But I'm also wondering whether having an alternative based on polling for blocking processing might still be worth it. Because I imagine that a lot of firewalls won't allow the callback to a temporary server run on the user's machine. Polling is less efficient obviously but it does not require juggling IP addresses, ports and servers.

Makefile
```diff
@@ -286,7 +286,7 @@ network-module-test: assets
 INTEGRATION_TEST_IN_DOCKER = docker exec core_test
 network-integration-test:
 	$(DOCKER_COMPOSE) --file tests/network/docker-compose.yml up -d
-	-$(INTEGRATION_TEST_IN_DOCKER) pytest -k 'test_integration_' -v --ignore-glob="$(TESTDIR)/network/*ocrd_all*.py"
+	-$(INTEGRATION_TEST_IN_DOCKER) pytest -k 'test_integration_' -v --ignore-glob="tests/network/*ocrd_all*.py"
```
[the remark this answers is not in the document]
MehmedGIT: `$(TESTDIR)` is problematic only in the docker environment, i.e., integration tests. Works as expected for `network-module-test` and `test`. Would you still like that changed?

src/ocrd_network/cli/client.py
```diff
@@ -1,11 +1,43 @@
...
+        self.send_response(200)
+        self.send_header("Content-Type", "text/plain")
+        self.end_headers()
+        self.wfile.write("finished".encode("utf-8"))
+        len = int(self.headers.get("Content-Length", 0))
+        data = self.rfile.read(len).decode("utf-8")
+        # TODO: how should the callback-content be handled/printed
+        print(f"Processor finished: {data}")
+        global STOP_WAITING_CALLBACK
+        STOP_WAITING_CALLBACK = True
```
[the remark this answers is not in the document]
MehmedGIT: I just implemented it the same way till there were running tests, then I would have changed it. It is good to know that the passed CallbackHandler class has a back reference to the server class. That should work.

src/ocrd_network/cli/client.py
MehmedGIT: > What is the purpose of self.callback_url of the callback server? 

It is for setting the `user_callback` field in the processing request, check [here](https://github.com/OCR-D/core/blob/1c8b9db05431d9a52e4763b9efcf4f123e3d65c8/src/ocrd_network/processing_worker.py#L237) and [here](https://github.com/OCR-D/core/blob/master/src/ocrd_network/models/job.py#L20)

> And why does it use the docker host IP but with the same port as the callback server? Is it so that the (remote) processing server can "find" the local callback server on the network - which would not work with `0.0.0.0`/`localhost`?

Using the localhost also fails by reaching timeout. I tried with a docker host IP although I knew it would probably fail here. It does. It works with the setup of @joschrew because 1) he runs the client in a native environment and 2) the processing worker runs in a docker environment without defining a separate network (not exactly sure for 2)).

Things are different when the client runs in a docker environment as in the integration test. The default docker host is no longer working since there is a defined separate network in the docker-compose file named `ocrd_network_test` in the `.env`. That network's host address is different than the default docker host address. The host for the processing server is defined as a string in the docker-compose file. The address resolution is then handled correctly by docker. The processing server URL at run-time becomes: `http://ps-docker-host.ocrd_network_test:8000`. Similarly, I assume that it is needed to define a string host for the client (`http://client-docker-host.ocrd_network_test:0`) so the processing workers can reach it within the same docker network.

> Also, I'm wondering how well this scales, the limiting factor is the number of free ports?

That should be the case, yes. But since we have a separate docker network there should still be more free ports than the localhost.

MehmedGIT: It will be still more complicated to handle the cases when the client runs natively since the callback address changes. The user should not be bothered by configuring that correctly. It should work by default. There is currently the `internal_callback_url` in the processing server config file. I think we would need one for the `client_callback_url` as well for switching between the docker and native environments. The polling still remains the much simpler solution.

MehmedGIT: I have implemented the polling client. Processing and Workflow tests have passed now.

Makefile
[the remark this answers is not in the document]
kba: As long as it works, no need. If I run into this again, I will fix in a separate PR.

src/ocrd_network/client.py
```diff
@@ -1,37 +1,39 @@
...
-        self.log.info(f"Sending processing request to: {req_url}")
-        self.log.debug(req_json)
-        response = requests_post(url=req_url, headers=req_headers, json=req_json)
-        return response.json()
+        self.polling_timeout = timeout
+        self.polling_wait = wait
+        self.polling_tries = int(timeout/wait)
+
+    def poll_job_status_till_timeout_fail_or_success(self, job_id: str) -> str:
+        return poll_job_status_till_timeout_fail_or_success(
```
kba, line 26: This is confusing, why are the imported function and this method the same name? On the surface this looks like infinite recursion.

src/ocrd_network/client_utils.py
```diff
@@ -0,0 +1,81 @@
+from requests import get as request_get, post as request_post
+from time import sleep
+from .constants import JobState, NETWORK_PROTOCOLS
+
+
+def _poll_endpoint_status(ps_server_host: str, job_id: str, job_type: str, tries: int, wait: int):
+    if job_type not in ["workflow", "processor"]:
+        raise ValueError("Unknown job type, expected 'workflow' or 'processor'")
```
kba, line 8:
```suggestion
        raise ValueError(f"Unknown job type '{job_type}', expected 'workflow' or 'processor'")
```

src/ocrd_utils/config.py
```diff
@@ -145,6 +145,16 @@ def _ocrd_download_timeout_parser(val):
         description="Default address of Processing Server to connect to (for `ocrd network client processing`).",
         default=(True, ''))
 
+config.add("OCRD_NETWORK_CLIENT_POLLING_TIMEOUT",
```
kba, line 148: Don't forget to add these to `src/ocrd/cli/__init__.py` for `--help` documentation.

kba (approved):

src/ocrd_network/cli/client.py
kba: Yeah, polling is much less error-prone, thanks for changing.

src/ocrd_network/client.py
MehmedGIT: I see why this is confusing and I would change that. The idea was to avoid duplicate code by providing a separate wrapper. It is not an infinite recursion though - the methods have different signatures.

src/ocrd_utils/config.py
MehmedGIT: Fixed with [0136db0](https://github.com/OCR-D/core/pull/1269/commits/0136db04e39269f77dfab3b71e46e43fa189e839).

src/ocrd_network/client.py
MehmedGIT: Resolved: [4115937](https://github.com/OCR-D/core/pull/1269/commits/41159371337542761a3fe9212d77224cec665df4).

joschrew:
I started to review, or rather to test the client but I couldn't finish so far. I will continue tomorrow. Here is what is not working for me up until now:
- OCRD_NETWORK_SERVER_ADDR_PROCESSING in not working as a default for `--address`. I think simply `default` is missing as argument for the click-decorator
- `ocrd network client processing processor ocrd-cis-ocropy-binarize --address "http://localhost:8000" -I DEFAULT -O BIN-2 -m /data/vd18test/mets.xml` gives an error: `TypeError: send_processing_job_request() got an unexpected keyword argument 'block'`. The parameter `block_till_job_end` could be renamed to block to resolve this.
- this is not related to this pr (but I came across this while testing) but in ocrd_network/server_utils.py in line 215 and 218 it must be `HTTP_400_BAD_REQUEST` instead of `HTTP_404_BAD_REQUEST` (which does not exist).
- running a processing job (`send_processing_job_request`) with the client does not work for me currently. The problem seems to be `parameters` in the inputs, at least it works if I remove this parameter from `job_input`. While debugging I think I realized a list instead of a dict is provided. Here I had to stop.

MehmedGIT:
> OCRD_NETWORK_SERVER_ADDR_PROCESSING in not working as a default for --address. I think simply default is missing as argument for the click-decorator

I'm not a big fan of copy-pasting the same default everywhere. Instead, I resolved it with [4de1e83](https://github.com/OCR-D/core/pull/1269/commits/4de1e83776bd280599ad7d8e5a508c3adcf90927)

> The parameter block_till_job_end could be renamed to block to resolve this.

Resolved [8e7ba26](https://github.com/OCR-D/core/pull/1269/commits/8e7ba26ec414106e0715f185e2508d2feb2604d5).

> this is not related to this pr (but I came across this while testing) but in ocrd_network/server_utils.py in line 215 and 218 it must be HTTP_400_BAD_REQUEST instead of HTTP_404_BAD_REQUEST (which does not exist).

You have caught something bigger than just that simple error. Resolved with [69808b6](https://github.com/OCR-D/core/pull/1269/commits/69808b6b0e368199fb6bab14c1ba6979d8cd33b4) and [d1af85b](https://github.com/OCR-D/core/pull/1269/commits/d1af85b931496be879d26ea6a8309f6cb8561236). The HTTP exception was being caught with the general Exception catch which was duplicating the error output, and thus obfuscating it.

> running a processing job (send_processing_job_request) with the client does not work for me currently. The problem seems to be parameters in the inputs, at least it works if I remove this parameter from job_input. While debugging I think I realized a list instead of a dict is provided. Here I had to stop.

That was not very pleasant to debug. The `ocrd.decorators.parameter_option` is returning `['{}']` and not `{}` by default. Which led to the `parameters={'{': '}'}` instead of `parameters={}`.
```python
21:57:40.222 WARNING ocrd_network.processing_server - Job input: processor_name='ocrd-cis-ocropy-binarize' path_to_mets='/home/mm/repos/ocrd_network_tests/ws29/data/mets.xml' workspace_id=None description='OCR-D Network client request' input_file_grps=['DEFAULT'] output_file_grps=['BIN-TEST'] page_id=None parameters={'{': '}'} result_queue_name=None callback_url=None agent_type=<AgentType.PROCESSING_WORKER: 'worker'> job_id='d1e075e3-6cb0-40c7-b854-d2d77322cfbf' depends_on=None
21:57:40.222 ERROR ocrd_network.processing_server - Failed to validate processing job input against the tool json of processor: ocrd-cis-ocropy-binarize
["[] Additional properties are not allowed ('{' was unexpected)"]
```

Dirty fix: [50f73c5](https://github.com/OCR-D/core/pull/1269/commits/50f73c58064bf9bed5479c370f356cefeda92210). Not sure how to better handle that.

joschrew (approved):
After the recent changes the client is working for me.

src/ocrd_network/client.py
```diff
@@ -1,28 +1,62 @@
...
+    post_ps_workflow_request,
+    verify_server_protocol
+)
 
 
 class Client:
-    def __init__(self, server_addr_processing: str = config.OCRD_NETWORK_SERVER_ADDR_PROCESSING):
+    def __init__(
+        self,
+        server_addr_processing: str,
```
kba, line 19:
```suggestion
        server_addr_processing: Optional[str],
```

But why not assign the default directly for `server_addr_processing` but do so for `timeout` and `wait`?

src/ocrd_network/server_utils.py
```diff
@@ -210,12 +212,12 @@ def validate_job_input(logger: Logger, processor_name: str, ocrd_tool: dict, job
         raise_http_exception(logger, status.HTTP_404_NOT_FOUND, message)
     try:
         report = ParameterValidator(ocrd_tool).validate(dict(job_input.parameters))
-        if not report.is_valid:
-            message = f"Failed to validate processing job input against the tool json of processor: {processor_name}\n"
-            raise_http_exception(logger, status.HTTP_400_BAD_REQUEST, message + report.errors)
     except Exception as error:
```
kba, line 216: When would this exception be raised? Validators should not raise errors but return a report. If you don't have a specific use case, it might be better to just not do `try/except` so that potential errors are actually raised and fixed.

src/ocrd_network/client.py
MehmedGIT: Because I did not want to place defaults for each click option `--address` for all cli methods. Since the address is passed directly to the constructor without checking if set or not, it overwrites the `server_addr_processing` with `None`. Since there are no `timeout` and `wait` flags in the client cli, these are set by default.

src/ocrd_network/server_utils.py
MehmedGIT: I wanted to be extra secure against unexpected errors. The workers or the server should never crash due to some error. The exceptions are still logged.

kba: Understood, but errors in

```
report = ParameterValidator(ocrd_tool).validate(dict(job_input.parameters))
```

would mean that something is broken in our implementation or in jsonschema. So this would likely not be a fluke that happens once, we would need to stop processing and fix the bug at the source. The only variable factor is the `job_input.parameters` dict and if user-provided input breaks the validator - which it must absolutely never do - we should fix that in the validator.

src/ocrd_network/cli/client.py
```diff
@@ -31,56 +61,92 @@ def processing_cli():
...
+    """
+    Check the status of a previously submitted processing job.
+    """
+    client = Client(server_addr_processing=address)
+    job_status = client.check_job_status(processing_job_id)
+    assert job_status
+    print(f"Processing job status: {job_status}")
+
+
 @processing_cli.command('processor')
```
kba, line 93: Any reason why this is not called `run`, like for workflows?

kba:
@MehmedGIT I've created the changelog for this PR, anything essential missing?

src/ocrd_network/cli/client.py
bertsky: I agree – that's more consistent. (And "run ocrd-cis-ocropy-segment" sounds better than "process ocrd-cis-ocropy-segment".)

That proposal might have [originated with me](https://github.com/OCR-D/core/issues/1032#issuecomment-1511482113). If so, please feel free to change.

kba: Thanks for the background info. `run` would be better, but I would not want to break existing scripts right in this PR, but we could do it for 3.0 stable (with a `:fire:` entry in the CHANGELOG).

MehmedGIT: [6608539](https://github.com/OCR-D/core/pull/1269/commits/66085392bc35334530b520b859f8760dc3518b80). 

> `run` would be better, but I would not want to break existing scripts right in this PR

In the processing server the endpoint has already been refactored to `/processor/run/{processor_name}` in the past. So not much to be changed. Only scripts using the `ocrd network client` cli would break. Me and @joschrew have none, our scripts are submitting the jobs directly to the Processing Server without the client.

src/ocrd_network/server_utils.py
MehmedGIT: > The only variable factor is the `job_input.parameters` dict and if user-provided input breaks the validator - which it must absolutely never do - we should fix that in the validator.

A failing validator would also probably fail most of the processing/workflow jobs. Input from the user breaking down the entire infrastructure is conceptually wrong to me. Even if all the processing jobs fail, the Processing Server should still respond to other requests such as checking logs of old jobs.  

There is also currently no graceful shutdown for the processing server. I.e., once the server dies, anything inside the internal processing cache of the server (not the RabbitMQ) will be lost.

MehmedGIT:
IMO, good to be merged.

PR #1271, opened 2024-08-16
ocrd_utils: separate make_xml_id, PcGtsType.id: apply make_xml_id

bertsky:


src/ocrd_utils/str.py
```diff
@@ -95,12 +96,16 @@ def make_file_id(ocrd_file, output_file_grp):
             ret = output_file_grp + '_' + ocrd_file.pageId
         else:
             ret = output_file_grp + '_' + ocrd_file.ID
+    return make_xml_id(ret)
+
+def make_xml_id(idstr: str) -> str:
+    ret = idstr
```
kba, line 102:
```suggestion
    """
    Turn ``idstr`` into a valid ``xml:id`` literal by replacing ``:`` with ``_``, removing everything non-alphanumeric, ``.`` and ``-`` and prepending `id_` if ``idstr`` starts with a number.
    """
    ret = idstr
```

kba (approved):

PR #1275, opened 2024-09-24
#1240 backports for v2

bertsky:
As discussed this is basically a cherry-picked version of #1240 with just the fixes (and some minor additions to get tests running smoothly again).

Changelog must also be cherry-picked from, but that's completely manual and can be done after review.

Most importantly, this solves the problems with Pillow 10.

### DIFF
--- a/src/ocrd/cli/ocrd_tool.py
+++ b/src/ocrd/cli/ocrd_tool.py
@@ -29,6 +29,8 @@ class OcrdToolCtx():
         self.filename = filename
         with codecs.open(filename, encoding='utf-8') as f:
             self.content = f.read()
+            # perhaps the validator should _always_ run (for default expansion)
+            # so validate command only for the report?
             self.json = loads(self.content)
 
 pass_ocrd_tool = click.make_pass_decorator(OcrdToolCtx)
--- a/src/ocrd/decorators/__init__.py
+++ b/src/ocrd/decorators/__init__.py
@@ -36,6 +36,7 @@ def ocrd_cli_wrap_processor(
     profile_file=None,
     version=False,
     overwrite=False,
+    resolve_resource=None,
     show_resource=None,
     list_resources=False,
     # ocrd_network params start #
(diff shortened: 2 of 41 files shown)


src/ocrd/cli/ocrd_tool.py
```diff
@@ -29,6 +29,8 @@ def __init__(self, filename):
         self.filename = filename
         with codecs.open(filename, encoding='utf-8') as f:
             self.content = f.read()
+            # perhaps the validator should _always_ run (for default expansion)
+            # so validate command only for the report?
```
kba, line 33: Yes, I see no reason not to run the default expansion always.

src/ocrd/decorators/__init__.py
```diff
@@ -148,8 +152,13 @@ def exit():
             s = io.StringIO()
             pstats.Stats(pr, stream=s).sort_stats("cumulative").print_stats()
             print(s.getvalue())
-        atexit.register(exit)
-    run_processor(processorClass, mets_url=mets, workspace=workspace, **kwargs)
+        atexit.register(goexit)
+    if log_filename:
+        log_ctx = redirect_stderr_and_stdout_to_file(log_filename)
```
kba, line 157: BTW this is one of the things I got rid of in the fork for our k8s deployment.

kba (approved):
LGTM

src/ocrd/decorators/__init__.py
kba: No, but something we should discuss because that does not play well with k8s.

src/ocrd/cli/ocrd_tool.py
kba: Of course, I just randomly answered a question in the comments here ;)

src/ocrd/decorators/__init__.py
bertsky: Oh, I see.

PR #1276, opened 2024-09-27
Identify provides the wrong ResolutionUnit for pyramid Tiff Files

mexthecat:
Added space after %U in ImageMagick identify format prameter.

We work with pyramid Tiffs in our project. Since several images are combined in one file, identify provides all resolutions and resolution units one after the other. However, there is no space between the first and second image and the resolution unit is therefore read incorrectly. 

```
ret = run(['identify', '-format', r'%[resolution.x] %[resolution.y] %U', img.filename], check=False, stderr=PIPE, stdout=PIPE)
...
tokens = ret.stdout.decode('utf-8').split(' ', 3)
```

`['118.11', '118.11', 'PixelsPerCentimeter118.11', '118.11 PixelsPerCentimeter118.11 118.11 PixelsPerCentimeter118.11 118.11 PixelsPerCentimeter118.11 118.11 PixelsPerCentimeter']`

A simple space after the %U should solve the problem

```
ret = run(['identify', '-format', r'%[resolution.x] %[resolution.y] %U ', img.filename], check=False, stderr=PIPE, stdout=PIPE)
...
tokens = ret.stdout.decode('utf-8').split(' ', 3)
```

`['118.11', '118.11', 'PixelsPerCentimeter', '118.11 118.11 PixelsPerCentimeter 118.11 118.11 PixelsPerCentimeter 118.11 118.11 PixelsPerCentimeter 118.11 118.11 PixelsPerCentimeter ']`

### DIFF
--- a/tests/model/test_exif.py
+++ b/tests/model/test_exif.py
@@ -27,7 +27,10 @@ from ocrd_models import OcrdExif
      1457, 2084, 1, 1, 1, 'inches', 'RGB', None),
     # tolerate multi-frame TIFF:
     ('gutachten/data/IMG/IMG_1.tif',
-     2088, 2634, 300, 300, 300, 'inches', 'RGB', 'raw')
+     2088, 2634, 300, 300, 300, 'inches', 'RGB', 'raw'),
+    # multi-frame TIFF with metric pixel density (is actually YCBCR not RGB but Pillow thinks otherwise...)
+    ('indian-ferns/data/OCR-D-IMG/0004.tif',
+     2626, 3620, 28, 28, 28, 'cm', 'RGB', 'jpeg'),
 ])
 def test_ocrd_exif(path, width, height, xResolution, yResolution, resolution, resolutionUnit, photometricInterpretation, compression):
     """Check EXIF attributes for different input formats
--- a/tests/model/test_exif.py
+++ b/tests/model/test_exif.py
@@ -24,7 +24,10 @@ from ocrd_models import OcrdExif
     ('leptonica_samples/data/OCR-D-IMG/OCR-D-IMG_1555_007.jpg',
      944, 1472, 1, 1, 1, 'inches', 'RGB', None),
     ('kant_aufklaerung_1784-jp2/data/OCR-D-IMG/INPUT_0020.jp2',
-     1457, 2084, 1, 1, 1, 'inches', 'RGB', None)
+     1457, 2084, 1, 1, 1, 'inches', 'RGB', None),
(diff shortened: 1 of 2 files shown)


bertsky:
Thanks a lot for the fix! 

Do you happen to have an example image handy that we can use in the repo as a test case?

mexthecat:
Yes - have some images but can not upload them here -> tif not allowed.

bertsky:
Could you try to zip them and attach that?

(Also, is the image unproblematic for publication, i.e. no licensing or other legal problems?)

mexthecat:
from https://archive.org/details/albumofindianfer00baynrich/albumofindianfer00baynrich/
[0004.zip](https://github.com/user-attachments/files/17162894/0004.zip)

bertsky:
Turns out we already had an multi-frame TIFF example in the test assets – see [44deb80](https://github.com/OCR-D/core/pull/1276/commits/44deb80434dbcf40289f7ce451e416f5f021a54d). 

@kba what do you think, should we add 0004.tif as well? (It has `cm` instead of `inches` and should yield `photometricInterpretation="YCBCR"` but instead gives `"RGB"` because Pillow does not catch it and ImageMagick never gets asked about it.)

tests/model/test_exif.py
```diff
@@ -27,7 +27,10 @@
      1457, 2084, 1, 1, 1, 'inches', 'RGB', None),
     # tolerate multi-frame TIFF:
     ('gutachten/data/IMG/IMG_1.tif',
-     2088, 2634, 300, 300, 300, 'inches', 'RGB', 'raw')
+     2088, 2634, 300, 300, 300, 'inches', 'RGB', 'raw'),
+    # multi-frame TIFF with metric pixel density (is actually YCBCR not RGB but Pillow thinks otherwise...)
```
bertsky, line 31: Actually, Pillow just does not say anything. We could ask ImageMagick about it, though.

(However, `%[EXIF:PhotometricInterpretation]` does not yield anything... Not sure how to get that info.)

So IMO this test should state `YCbCr` and fail until we fix `OcrdExif.run_identify` to include the results from the CLI, or we find a way to make Pillow give us the info.

BTW, using Pillow's [TiffTags](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#tiff) we can actually get the right value:
```python
# val = img.tag_v2.named()['PhotometricInterpretation']
val = img.tag_v2.get(262)
interpretation = dict((v, k) for k, v in TiffTags.TAGS_V2.get(262).enum.items())
val = interpretation[val]
```

PR #1277, opened 2024-10-01
introduce `-p/--print-status` to ocrd network client blocking commands

MehmedGIT:
Provide flexibility to ocrd network client blocking commands by letting users control the printing of outputs.

### DIFF
--- a/src/ocrd_network/server_utils.py
+++ b/src/ocrd_network/server_utils.py
@@ -264,8 +264,8 @@ def kill_mets_server_zombies(minutes_ago=60) -> List[int]:
             continue
         cmdline = cmdline_file.read_text().replace('\x00', ' ')
         if re.match(cmdline_pat, cmdline):
-            pid = procdir.name
+            pid = int(procdir.name)
             ret.append(pid)
             print(f'METS Server with PID {pid} was created {ctime_ago} minutes ago, more than {minutes_ago}, so killing (cmdline="{cmdline})', file=sys.stderr)
-            os.kill(int(pid), signal.SIGTERM)
+            os.kill(pid, signal.SIGTERM)
     return ret
--- a/src/ocrd_network/server_utils.py
+++ b/src/ocrd_network/server_utils.py
@@ -4,7 +4,7 @@ import signal
 from pathlib import Path
 from json import dumps, loads
 from urllib.parse import urljoin
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 from time import time
 
 from fastapi import HTTPException, status, UploadFile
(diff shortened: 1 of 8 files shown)


src/ocrd_utils/config.py
```diff
@@ -160,13 +160,18 @@ def _ocrd_download_timeout_parser(val):
...
 config.add("OCRD_NETWORK_CLIENT_POLLING_TIMEOUT",
            description="Timeout for a blocking ocrd network client (in seconds).",
            parser=int,
            default=(True, 3600))
 
+config.add("OCRD_NETWORK_CLIENT_POLLING_PRINT",
+           description="Whether the blocking client commands should print status output each iteration.",
+           parser=bool,
+           default=(True, False))
+
```
kba, line 174: I appreciate the functionality, obviously. But since you have added `print_output` kwargs, why not just have this as a parameter to `ocrd network client workflow run --print-state`? It's fine either way but I'm wondering where else polling happens and whether enabling that config env var will cause a log of logging in different places.

kba (approved):

src/ocrd_utils/config.py
MehmedGIT: That's also possible, yes. Just for usability reasons. Polling happens only on workflow jobs and processing jobs.

> I'm wondering where else polling happens and whether enabling that config env var will cause a log of logging in different places.

No, it will not have such side effects in other places, e.g., processing server integration tests. By default, the printing is always False, and the environment variable is passed only for the network client.

MehmedGIT: [342ef3a](https://github.com/OCR-D/core/pull/1277/commits/342ef3a78f3620ff3e63200b2a9bc4c11639c581)

kba: Thanks for the explanation and the CLI flag

kba (approved):

src/ocrd_utils/config.py
```diff
@@ -160,13 +160,18 @@ def _ocrd_download_timeout_parser(val):
...
+           default=(True, 10))
 
 config.add("OCRD_NETWORK_CLIENT_POLLING_TIMEOUT",
            description="Timeout for a blocking ocrd network client (in seconds).",
            parser=int,
            default=(True, 3600))
 
+config.add("OCRD_NETWORK_CLIENT_POLLING_PRINT",
+           description="Whether the blocking client commands should print status output each iteration.",
+           parser=bool,
```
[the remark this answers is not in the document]
MehmedGIT: Good, but I have removed that environment variable when implementing the flag.

src/ocrd_network/server_utils.py
```diff
@@ -241,3 +247,25 @@ def validate_first_task_input_file_groups_existence(logger: Logger, mets_path: s
...
+        ctime_ago = int((now - procdir.stat().st_ctime) / 60)
+        if ctime_ago < minutes_ago:
+            continue
+        cmdline = cmdline_file.read_text().replace('\x00', ' ')
+        if re.match(cmdline_pat, cmdline):
+            pid = procdir.name
+            ret.append(pid)
+            print(f'METS Server with PID {pid} was created {ctime_ago} minutes ago, more than {minutes_ago}, so killing (cmdline="{cmdline})', file=sys.stderr)
+            os.kill(int(pid), signal.SIGTERM)
+    return ret
```
MehmedGIT, line 271: This returns `List[str]` not `List[int]` as expected.

kba: Eek, you're right. Will fix.

PR #1278, opened 2024-10-01
post_ps_workflow_request: pagewise configurable

kba:
Make the page-wise feature of the processing server toggleable when running workflows with `ocrd network client workflow run`.

Probably a good idea to merge this with #1277 since it touches the same files.

### DIFF
--- a/src/ocrd_network/cli/client.py
+++ b/src/ocrd_network/cli/client.py
@@ -180,7 +180,7 @@ def check_workflow_job_status(address: Optional[str], workflow_job_id: str):
                                 'the "OCRD_NETWORK_SERVER_ADDR_PROCESSING" env variable is used by default')
 @click.option('-m', '--path-to-mets', required=True)
 @click.option('-w', '--path-to-workflow', required=True)
-@click.option('-p/-P', '--page-wise/--no-page-wise', is_flag=True, default=False, help="Whether to generate per-page jobs")
+@click.option('--page-wise/--no-page-wise', is_flag=True, default=False, help="Whether to generate per-page jobs")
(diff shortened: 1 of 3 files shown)


src/ocrd_network/cli/client.py
```diff
@@ -179,6 +180,7 @@ def check_workflow_job_status(address: Optional[str], workflow_job_id: str):
                                 'the "OCRD_NETWORK_SERVER_ADDR_PROCESSING" env variable is used by default')
 @click.option('-m', '--path-to-mets', required=True)
 @click.option('-w', '--path-to-workflow', required=True)
+@click.option('-p/-P', '--page-wise/--no-page-wise', is_flag=True, default=False, help="Whether to generate per-page jobs")
 @click.option('-b', '--block', default=False, is_flag=True,
               help='If set, the client will block till job timeout, fail or success.')
 @click.option('-p', '--print-state', default=False, is_flag=True,
```
MehmedGIT, line 186: I have not paid attention to the shortcut `-p/-P` and introduced it again for `--print-state`. Just replace that, the rest is good.

MehmedGIT (approved):

src/ocrd_network/cli/client.py
kba: Oh right, yeah, we probably don't even need the short versions. I think selecting `--page-wise` vs `--no-page-wise` should be a very conscious decision worth typing a long option.

MehmedGIT: Resolved with [854403d](https://github.com/OCR-D/core/pull/1278/commits/854403de6ea880c31b82463bba3850c07565327d)

PR #1280, opened 2024-10-01
Dockerfile.cuda-torch: do NOT rm /build/core

bertsky:
since we installed core in editable mode!

### DIFF
--- a/Dockerfile.cuda-torch
+++ b/Dockerfile.cuda-torch
@@ -9,7 +9,5 @@ RUN make deps-torch
 
 WORKDIR /data
 
-RUN rm -fr /build
-
 CMD ["/usr/local/bin/ocrd", "--help"]
 


bertsky:
Warrants a post-release IMO. Or, if possible, a quick follow-up with v2.69.1.

bertsky:
> Warrants a post-release IMO. Or, if possible, a quick follow-up with v2.69.1.

Or is that even necessary? We could just merge to master – the CD should pick this up and release on Dockerhub, overwriting the current latest and :v2.69.0, right?

kba:
> > Warrants a post-release IMO. Or, if possible, a quick follow-up with v2.69.1.
> 
> Or is that even necessary? We could just merge to master – the CD should pick this up and release on Dockerhub, overwriting the current latest and :v2.69.0, right?

Yep, that should be the case. Not the way it should be done but fine in this case :)

kba (approved):
LGTM

PR #1281, opened 2024-10-01
client_utils: use a retrying http client

kba:
I added those to my local fork for robustness to handle network connectivity issues. Not sure how sensible that is and hard to investigate how effective this is, so just a draft for now. If we decide it does make sense to have a retrying (and possibly also debouncing retries for more failure resilience) http client, we could generalize this.

(copied from 3b040b7a by @bertsky)

### DIFF
--- a/src/ocrd_network/client_utils.py
+++ b/src/ocrd_network/client_utils.py
@@ -1,5 +1,9 @@
-from requests import get as request_get, post as request_post
+from requests import Session, get as request_get
 from time import sleep
+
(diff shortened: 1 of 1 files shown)


MehmedGIT (approved):
I do not see any downsides of a retrying HTTP client that would keep us away from it. 
The only small disadvantage may be filled logs on the Processing Server side when 500 errors occur, since the server returns these for unexpected errors. The other error codes are reasonable.

PR #1283, opened 2024-10-02
add endpoint DELETE /kill_mets_server_zombies to kill -SIGTERM METS servers with ctime > 60mins ago

kba:
Also, use `os.kill` directly in `OcrdMetsServer.kill_process` (no idea why I did that via `subprocess.run` 🤷)

### DIFF
--- a/src/ocrd/mets_server.py
+++ b/src/ocrd/mets_server.py
@@ -1,8 +1,10 @@
 """
 # METS server functionality
 """
+import os
 import re
 from os import _exit, chmod
+import signal
 from typing import Dict, Optional, Union, List, Tuple
 from time import sleep
 from pathlib import Path
@@ -428,8 +430,7 @@ class OcrdMetsServer:
 
     @staticmethod
     def kill_process(mets_server_pid: int):
-        subprocess_run(args=["kill", "-s", "SIGINT", f"{mets_server_pid}"], shell=False, universal_newlines=True)
-        return
+        return os.kill(mets_server_pid, signal.SIGTERM)
 
     def shutdown(self):
         if self.is_uds:
--- a/src/ocrd/mets_server.py
+++ b/src/ocrd/mets_server.py
@@ -430,7 +430,12 @@ class OcrdMetsServer:
 
     @staticmethod
     def kill_process(mets_server_pid: int):
-        return os.kill(mets_server_pid, signal.SIGTERM)
+        os.kill(mets_server_pid, signal.SIGINT)
+        sleep(3)
+        try:
+            os.kill(mets_server_pid, signal.SIGKILL)
+        except ProcessLookupError as e:
+            pass
 
     def shutdown(self):
         if self.is_uds:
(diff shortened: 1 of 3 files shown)


src/ocrd/mets_server.py
```diff
@@ -428,8 +430,7 @@ def create_process(mets_server_url: str, ws_dir_path: str, log_file: str) -> int
 
     @staticmethod
     def kill_process(mets_server_pid: int):
-        subprocess_run(args=["kill", "-s", "SIGINT", f"{mets_server_pid}"], shell=False, universal_newlines=True)
-        return
+        return os.kill(mets_server_pid, signal.SIGTERM)
```
MehmedGIT, line 433: I would still prefer to handle that softly first with `signal.SIGINT` sleep for some time then send `signal.SIGTERM` to give enough time for the Mets Server process to clean stuff after itself (if required).

kba: Sure, I'm open to that. Ideally we should test this and see whether SIGINT is enough (I know it was in my testing but since I need to work around this issue today, I want to be certain that process `mets_server_pid` is gone after `kill.process` (and the workaround zombie killer endpoint). `SIGINT` and `SIGTERM` should behave mostly the same anyway, `SIGKILL` would be the harsher solution.

MehmedGIT: Yes, just the idea is to stay flexible and let the process catch the signal and do some internal cleaning. Sorry, I meant the harsher`SIGKILL` instead of `SIGTERM`. Just in case the process still gets stuck during the clean up after catching the SIGINT.

src/ocrd_network/server_utils.py
```diff
@@ -241,3 +247,25 @@ def validate_first_task_input_file_groups_existence(logger: Logger, mets_path: s
         if group not in available_groups:
             message = f"Input file group '{group}' of the first processor not found: {input_file_grps}"
             raise_http_exception(logger, status.HTTP_422_UNPROCESSABLE_ENTITY, message)
+
+
+def kill_mets_server_zombies(minutes_ago=60) -> list[int]:
```
MehmedGIT, line 252: Should return `List[int]`

kba: Thx [4862d72](https://github.com/OCR-D/core/pull/1283/commits/4862d72fe6f7149ff4ce97d56ac870837bafddc5)

src/ocrd/mets_server.py
MehmedGIT, line 433:
```suggestion
        os.kill(mets_server_pid, signal.SIGINT)
        sleep(3)
        os.kill(mets_server_pid, signal.SIGKILL)
```

MehmedGIT (changes requested):
That's what I meant, at least just inside the `OcrdMetsServer.kill_process()`. I don't mind what exactly happens inside the workaround endpoint.

src/ocrd/mets_server.py
kba: This will result in `ProcessLookupError` in the `SIGKILL` line if the `SIGINT` succeeded.

MehmedGIT: The SIGINT may suffice after all. It is just to cover the rare cases which may be ignored for now.

MehmedGIT (approved):

src/ocrd/mets_server.py
MehmedGIT: Yeap, that is what I meant.

PR #1284, opened 2024-10-02
Fix mets server: clean leftover zombie processes

MehmedGIT:
This PR fixes:
- No more zombie processes for Mets Servers when they terminate!
- Appropriately deleting the Unix Domain Socket in more complicated configurations (e.g., UDS over TCP multiplexing)
- More extensive logging (to separate files as well) for:
  + `ocrd_network.server_cache.locked_pages`
  + `ocrd_network.server_cache.processing_requests`
  + `ocrd.models.ocrd_mets.server.{self.url}`
  + `ocrd_network.tcp_to_uds_mets_proxy`
  
Please make sure you adapt your logging configuration file appropriately to avoid extreme amounts of logs. Maybe an adaption is needed to the default logging configuration file - open for discussions.

### DIFF
--- a/src/ocrd/mets_server.py
+++ b/src/ocrd/mets_server.py
@@ -447,7 +447,7 @@ class OcrdMetsServer:
                 Path(self.url).unlink()
 
     def startup(self):
-        self.log.info(f"Configuring up the Mets Server")
+        self.log.info(f"Configuring the Mets Server")
 
         workspace = self.workspace
 
--- a/src/ocrd/mets_server.py
+++ b/src/ocrd/mets_server.py
@@ -157,13 +157,13 @@ class ClientSideOcrdMets:
         Request writing the changes to the file system
         """
         if not self.multiplexing_mode:
-            self.session.request("PUT", url=self.url)
+            return self.session.request("PUT", url=self.url).text
         else:
-            self.session.request(
+            return self.session.request(
                 "POST",
                 self.url,
                 json=MpxReq.save(self.ws_dir_path)
-            )
+            ).json()["text"]
 
     def stop(self):
         """
@@ -171,14 +171,13 @@ class ClientSideOcrdMets:
         """
         try:
             if not self.multiplexing_mode:
-                self.session.request("DELETE", self.url)
-                return
+                return self.session.request("DELETE", self.url).text
             else:
-                self.session.request(
+                return self.session.request(
                     "POST",
                     self.url,
                     json=MpxReq.stop(self.ws_dir_path)
-                )
+                ).json()["text"]
         except ConnectionError:
             # Expected because we exit the process without returning
             pass
@@ -348,12 +347,12 @@ class MpxReq:
     @staticmethod
     def save(ws_dir_path: str) -> Dict:
         return MpxReq.__args_wrapper(
-            ws_dir_path, method_type="PUT", response_type="empty", request_url="", request_data={})
+            ws_dir_path, method_type="PUT", response_type="text", request_url="", request_data={})
 
     @staticmethod
     def stop(ws_dir_path: str) -> Dict:
         return MpxReq.__args_wrapper(
-            ws_dir_path, method_type="DELETE", response_type="empty", request_url="", request_data={})
+            ws_dir_path, method_type="DELETE", response_type="text", request_url="", request_data={})
 
     @staticmethod
     def reload(ws_dir_path: str) -> Dict:
--- a/src/ocrd/mets_server.py
+++ b/src/ocrd/mets_server.py
@@ -437,11 +437,8 @@ class OcrdMetsServer:
         except ProcessLookupError as e:
             pass
 
-    def shutdown(self):
-        if self.is_uds:
-            if Path(self.url).exists():
-                self.log.warning(f"Due to a server shutdown, removing the existing UDS socket file: {self.url}")
-                Path(self.url).unlink()
+    @staticmethod
+    def shutdown():
         # os._exit because uvicorn catches SystemExit raised by sys.exit
         _exit(0)
 
@@ -472,7 +469,8 @@ class OcrdMetsServer:
             """
             Write current changes to the file system
             """
-            return workspace.save_mets()
+            workspace.save_mets()
+            return Response(status_code=200, content="The Mets Server is writing changes to disk.")
 
         @app.delete(path='/')
         async def stop():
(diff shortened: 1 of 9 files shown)


MehmedGIT:
@joschrew, would be great if you could test your docker setup with this PR to see if there are any issues before this is merged.

src/ocrd/mets_server.py
```diff
@@ -438,15 +438,18 @@ def kill_process(mets_server_pid: int):
             pass
 
     def shutdown(self):
+        # os._exit because uvicorn catches SystemExit raised by sys.exit
+        # _exit(0)
```
kba, line 442:
```suggestion
```

Graceful shutdown working nicely AFAICT, so the this can go for good.

src/ocrd/mets_server.py
```diff
@@ -438,15 +438,18 @@ def kill_process(mets_server_pid: int):
...
         if self.is_uds:
             if Path(self.url).exists():
                 self.log.warning(f"Due to a server shutdown, removing the existing UDS socket file: {self.url}")
                 Path(self.url).unlink()
-        # os._exit because uvicorn catches SystemExit raised by sys.exit
-        _exit(0)
 
     def startup(self):
-        self.log.info("Starting up METS server")
+        self.log.info(f"Configuring up the Mets Server")
```
kba, line 452:
```suggestion
        self.log.info(f"Configuring the Mets Server")
```

Though I found `Starting up` clearer.

src/ocrd/mets_server.py
```diff
@@ -556,17 +580,22 @@ async def add_file(
...
         if self.is_uds:
             # Create socket and change to world-readable and -writable to avoid permission errors
             self.log.debug(f"chmod 0o677 {self.url}")
             server = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+            # TODO: Not required after #1284, consider removing
+            """
             if Path(self.url).exists() and not is_socket_in_use(self.url):
                 # remove leftover unused socket which blocks startup
                 Path(self.url).unlink()
+            """
```
kba, line 598:
```suggestion
```

Then let's remove it, it was a workaround anyway.

src/ocrd/mets_server.py
```diff
@@ -578,10 +607,10 @@ async def add_file(
         uvicorn_kwargs['log_config'] = None
         uvicorn_kwargs['access_log'] = False
 
-        self.log.debug("Starting uvicorn")
+        self.log.info("Starting the uvicorn Mets Server")
         uvicorn.run(app, **uvicorn_kwargs)
 
-
+# TODO: Not required after #1284, consider removing
```
kba, line 613: Cannot propose outside the diff context, but if we don't need it any more, feel free to rremove `def is_socket_in_use...`.

src/ocrd_network/runtime_data/deployer.py
```diff
@@ -153,26 +155,41 @@ def start_uds_mets_server(self, ws_dir_path: str) -> Path:
...
-        stop_mets_server(mets_server_url=mets_server_url)
+        mets_server_pid = self.mets_servers[str(self.mets_servers_paths[str(Path(path_to_mets).parent)])]
+        self.log.info(f"Terminating mets server with pid: {mets_server_pid}")
+        p = psutil.Process(mets_server_pid)
+        stop_mets_server(self.log, mets_server_url=mets_server_url, ws_dir_path=str(Path(path_to_mets).parent))
+        if p.is_running():
+            p.wait()
+            self.log.info(f"Terminated mets server with pid: {mets_server_pid}")
+        else:
+            self.log.info(f"Mets server has already terminated with pid: {mets_server_pid}")
```
kba, line 194:
```suggestion
            self.log.info(f"Mets server with pid: {mets_server_pid} has already terminated.")
```

kba (changes requested):
Excellent, this will make managing the PS much easier.

`_exit` was a terrible idea, obviously, my bad.

Besides some minor documentation issues, LGTM.

However, I am unsure about the additional logging. Not so much the new logs (which are appreciated) but the upgrading from `log.debug` to `log.info`. Since at least in my use case (kubernetes), everything, including all separate log files, is redirected to a single STDOUT stream, this will make it hard to spot relevant logging in all the noise. At least we should use different loggers for the very frequent loggings (i.e. have a `self.log/ocrd_network.processing_server` for regular stuff and `self.verbose_log/ocrd_network.processing_server_verbose`), so it is easier to configure in the `ocrd_logging.conf` and to filter by in the log viewer?

src/ocrd/mets_server.py
MehmedGIT: Yes, but then there is another log right before starting the uvicorn server at the end of the method. I thought a differentiation between the two may be good.

src/ocrd/mets_server.py
MehmedGIT: Yes, but @joschrew should potentially first confirm that with his docker setup. I did not want to remove that since he added the workaround and I could not test with his setup.

kba (approved):
> However, I am unsure about the additional logging. Not so much the new logs (which are appreciated) but the upgrading from `log.debug` to `log.info`. Since at least in my use case (kubernetes), everything, including all separate log files, is redirected to a single STDOUT stream, this will make it hard to spot relevant logging in all the noise. At least we should use different loggers for the very frequent loggings (i.e. have a `self.log/ocrd_network.processing_server` for regular stuff and `self.verbose_log/ocrd_network.processing_server_verbose`), so it is easier to configure in the `ocrd_logging.conf` and to filter by in the log viewer?

The logging is already distributed over different loggers, so that part is not an issue.

And the `debug->info` I can also live with, the intention is to not force users to set global log level to `debug` to see these. In any case, it can be adapted in the logging.conf if desired.

I'll be deploying this now and it is good to merge once @joschrew okays it.

MehmedGIT:
> There the socket-file is deleted in Deployer.start_uds_mets_server when starting the mets_server for a workspace and the socket-file is existing. This is necessary for me to work because when shutting down the docker containers the socket file is not deleted. It seems to me, that the shutdown Method of OcrdMetsServer is not called, I couldn't find any log entry, which should have been issued in this method. But that's ok for everything to work for me.

Strangely, it seems the DELETE request is never sent to the mets server in the docker environment. The socket file is volume mapped on the host OS, and deleted with a shutdown method call by the Mets server process that resides inside a container. So the socket file would not be removed even if the shutdown method was invoked correctly. It seems that not much can be done for the docker environment other than the fallback [solution](https://github.com/OCR-D/core/blob/network_client_block_prints/src/ocrd_network/runtime_data/deployer.py#L149-L153) you talked about.

src/ocrd/mets_server.py
MehmedGIT: Resolved with [5055309](https://github.com/OCR-D/core/pull/1284/commits/50553093180ac6b641273c08d559bbadf5e9b1d2)

PR #1285, opened 2024-10-09
Rabbitmq heartbeat env

MehmedGIT:
A new env variable for the RabbitMQ to eliminate a hard-coded heartbeat value.

### DIFF
--- a/tests/network/config.py
+++ b/tests/network/config.py
@@ -89,11 +89,21 @@ test_config.add(
 
 test_config.add(
     name="OCRD_NETWORK_RABBITMQ_CLIENT_CONNECT_ATTEMPTS",
-    description="""
-    Number of attempts for a RabbitMQ client to connect before failing
-    """,
+    description="Number of attempts for a RabbitMQ client to connect before failing",
     parser=int,
     default=(True, 3)
+)
+
+test_config.add(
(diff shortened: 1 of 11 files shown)


tests/network/config.py
```diff
@@ -89,11 +89,21 @@
...
+)
+
+test_config.add(
+    name="OCRD_NETWORK_RABBITMQ_HEARTBEAT",
     description="""
-    Number of attempts for a RabbitMQ client to connect before failing
+    Controls AMQP heartbeat timeout negotiation during connection tuning. An integer value always overrides the value 
+    proposed by broker. Use 0 to deactivate heartbeats and None to always accept the broker's proposal. If a callable 
+    is given, it will be called with the connection instance and the heartbeat timeout proposed by broker as its 
+    arguments. The callback should return a non-negative integer that will be used to override the broker's proposal.
```
kba, line 103:
```suggestion
    Controls AMQP heartbeat timeout (in seconds) negotiation during connection tuning. An integer value always overrides the value 
    proposed by broker. Use 0 to deactivate heartbeat.
```

The rest is not applicable here because it it's an env var that only accepts `int`.

And make clear it's measured in seconds.

kba (changes requested):
Forgot to finish the review yesterday. LGTM, just the updated the comment.

kba (approved):

PR #1288, opened 2024-10-09
ocrd_utils.initLogging: also add handler to root logger

bertsky:
As a follow-up to https://github.com/OCR-D/core/commit/ccb416b13e7f91781568fda8e60ad8182bfea88c, but more general:

`initLogging` should _always_ affect the root logger `''`, too. (Not just when disabling logging.)

Reasons:
- to be consistent with file config (which by default _also_ configures the unqualified `root` logger for our default handlers
- to prevent imported libraries from initializing logging on their own (like `import tensorflow`) – the place to check whether logging has been set up yet is always `getLogger('').hasHandlers()`, also in Python's logging.basicConfig BTW

I know we deliberately changed this to _exclude_ the root logger earlier. But I cannot find the reasoning behind that anymore. Maybe "we are a library, not an application"? Well, I agree actually, but then just don't call `initLogging`. (Perhaps at that time we got distracted by the fact that we still **did** call `initLogging` unconditionally when importing `ocrd_utils.logging`.)

BTW, IMO most places (core and modules code) can get their loggers from Python stdlib's `logging.getLogger` instead of our `ocrd_utils.logging.getLogger`, because we don't add anything to that nowadays.

### DIFF
--- a/src/ocrd_utils/ocrd_logging.conf
+++ b/src/ocrd_utils/ocrd_logging.conf
@@ -56,22 +56,22 @@ handlers=consoleHandler,fileHandler
 # ocrd loggers
 [logger_ocrd]
 level=INFO
-handlers=consoleHandler,fileHandler
+handlers=
 qualname=ocrd
-propagate=0
 
 [logger_ocrd_network]
 level=INFO
-handlers=consoleHandler,processingServerHandler
+#handlers=consoleHandler,processingServerHandler
+handlers=processingServerHandler
 qualname=ocrd_network
-propagate=0
+#propagate=0
 
 #
 # logger tensorflow
 #
 [logger_ocrd_tensorflow]
 level=ERROR
-handlers=consoleHandler
+handlers=
 qualname=tensorflow
 
 #
@@ -79,7 +79,7 @@ qualname=tensorflow
 #
 [logger_ocrd_shapely_geos]
 level=ERROR
-handlers=consoleHandler
+handlers=
 qualname=shapely.geos
 
 
@@ -88,7 +88,7 @@ qualname=shapely.geos
 #
 [logger_ocrd_PIL]
 level=INFO
-handlers=consoleHandler
+handlers=
 qualname=PIL
 
 #
@@ -96,34 +96,32 @@ qualname=PIL
 #
 [logger_paramiko]
 level=INFO
-handlers=consoleHandler
+handlers=
 qualname=paramiko
-propagate=0
 
 [logger_paramiko_transport]
 level=INFO
-handlers=consoleHandler
+handlers=
 qualname=paramiko.transport
-propagate=0
 
 #
 # uvicorn loggers
 #
 [logger_uvicorn]
 level=INFO
-handlers=consoleHandler
+handlers=
 qualname=uvicorn
 [logger_uvicorn_access]
 level=WARN
-handlers=consoleHandler
+handlers=
 qualname=uvicorn.access
 [logger_uvicorn_error]
 level=INFO
-handlers=consoleHandler
+handlers=
 qualname=uvicorn.error
 [logger_multipart]
 level=INFO
-handlers=consoleHandler
+handlers=
 qualname=multipart
 
 
--- a/src/ocrd_utils/ocrd_logging.conf
+++ b/src/ocrd_utils/ocrd_logging.conf
@@ -34,7 +34,7 @@ keys=defaultFormatter,detailedFormatter
 # default logger "root" using consoleHandler
 #
 [logger_root]
-level=INFO
+level=WARNING
 handlers=consoleHandler,fileHandler
 
 
--- a/src/ocrd_utils/logging.py
+++ b/src/ocrd_utils/logging.py
@@ -181,11 +181,8 @@ def initLogging(builtin_only=False, force_reinit=False, silent=not config.OCRD_L
         ocrd_handler = logging.StreamHandler(stream=sys.stderr)
         ocrd_handler.setFormatter(logging.Formatter(fmt=LOG_FORMAT, datefmt=LOG_TIMEFMT))
         ocrd_handler.setLevel(logging.DEBUG)
-        for logger_name in ROOT_OCRD_LOGGERS:
-            logger = logging.getLogger(logger_name)
-            logger.addHandler(ocrd_handler)
-            if logger_name:
-                logger.propagate = False # avoid duplication (from root handler)
+        root_logger = logging.getLogger('')
+        root_logger.addHandler(ocrd_handler)
         for logger_name, logger_level in LOGGING_DEFAULTS.items():
             logging.getLogger(logger_name).setLevel(logger_level)
     _initialized_flag = True
--- a/src/ocrd_utils/logging.py
+++ b/src/ocrd_utils/logging.py
@@ -48,13 +48,6 @@ __all__ = [
     'setOverrideLogLevel',
 ]
 
-# These are the loggers we add handlers to
-ROOT_OCRD_LOGGERS = [
-    '',
-    'ocrd',
-    'ocrd_network'
-]
-
 LOGGING_DEFAULTS = {
     'ocrd': logging.INFO,
     'ocrd_network': logging.INFO,
@@ -198,24 +191,16 @@ def disableLogging(silent=not config.OCRD_LOGGING_DEBUG):
     if _initialized_flag and not silent:
         print("[LOGGING] Disabling logging", file=sys.stderr)
     _initialized_flag = False
-    # logging.basicConfig(level=logging.CRITICAL)
-    # logging.disable(logging.ERROR)
-    # remove all handlers for the ocrd logger
-    for logger_name in ROOT_OCRD_LOGGERS:
-        for handler in logging.getLogger(logger_name).handlers[:]:
-            logging.getLogger(logger_name).removeHandler(handler)
-    for logger_name in LOGGING_DEFAULTS:
-        logging.getLogger(logger_name).setLevel(logging.NOTSET)
+    # remove all handlers we might have added (via initLogging on builtin or file config)
+    for logger_name in logging.root.manager.loggerDict:
+        if not silent:
+            print(f'[LOGGING] Resetting {logger_name} log level and handlers')
+        logger = logging.getLogger(logger_name)
+        logger.setLevel(logging.NOTSET)
+        for handler in logger.handlers[:]:
+            logger.removeHandler(handler)
+    for handler in logging.root.handlers[:]:
+        logging.root.removeHandler(handler)
     # Python default log level is WARNING
     logging.root.setLevel(logging.WARNING)
 
-# Initializing stream handlers at module level
-# would cause message output in all runtime contexts,
-# including those which are already run for std output
-# (--dump-json, --version, ocrd-tool, bashlib etc).
-# So this needs to be an opt-in from the CLIs/decorators:
-#initLogging()
-# Also, we even have to block log output for libraries
-# (like matplotlib/tensorflow) which set up logging
-# themselves already:
-disableLogging()
--- a/src/ocrd_utils/logging.py
+++ b/src/ocrd_utils/logging.py
@@ -163,18 +163,6 @@ def initLogging(builtin_only=False, force_reinit=False, silent=not config.OCRD_L
     global _initialized_flag
     if _initialized_flag and not force_reinit:
         return
-    # disableLogging()
-
-    # https://docs.python.org/3/library/logging.html#logging.disable
-    # If logging.disable(logging.NOTSET) is called, it effectively removes this
-    # overriding level, so that logging output again depends on the effective
-    # levels of individual loggers.
-    logging.disable(logging.NOTSET)
-
-    # remove all handlers for the ocrd root loggers
-    for logger_name in ROOT_OCRD_LOGGERS:
-        for handler in logging.getLogger(logger_name).handlers[:]:
-            logging.getLogger(logger_name).removeHandler(handler)
 
     config_file = None
     if not builtin_only:
--- a/src/ocrd_utils/logging.py
+++ b/src/ocrd_utils/logging.py
@@ -50,6 +50,7 @@ __all__ = [
 
 # These are the loggers we add handlers to
 ROOT_OCRD_LOGGERS = [
+    '',
     'ocrd',
     'ocrd_network'
 ]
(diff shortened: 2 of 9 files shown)


MehmedGIT:
> Maybe "we are a library, not an application"? Well, I agree actually, but then just don't call initLogging.

But if `initLogging` is not invoked then the logging `is` (maybe it's a `was` now, not sure) not working at all when core code is imported in another project say Operandi. Ideally, the behavior of the root logger, `ocrd` and `ocrd_network` should not be affected by the default logging config file if the user does not do it specifically with a manual logging config file. 

This said, the `ocrd_network` logging acts weirdly and I am not able to find out how to fix it. I did some experiments and observations that I am going to share in next Monday's meeting.

kba:
> I know we deliberately changed this to _exclude_ the root logger earlier. But I cannot find the reasoning behind that anymore. Maybe "we are a library, not an application"? Well, I agree actually, but then just don't call `initLogging`. (Perhaps at that time we got distracted by the fact that we still call `initLogging` unconditionally when importing `ocrd_utils.logging`.)

Yes, the idea was to behave unintrusively as a library but as you point out, in practice we don't

> 
> BTW, IMO most places (core and modules code) can get their loggers from Python stdlib's `logging.getLogger` instead of our `ocrd_utils.logging.getLogger`, because we don't add anything to that nowadays.

Agreed.


> This said, the `ocrd_network` logging acts weirdly and I am not able to find out how to fix it. I did some experiments and observations that I am going to share in next Monday's meeting.

🙏

Let's discuss on Monday, I'm all for pushing for a clean logging because that is high on the list of most painful aspects of the OCR-D software at the moment.

bertsky:
@MehmedGIT 
> But if `initLogging` is not invoked then the logging `is` (maybe it's a `was` now, not sure) not working at all when core code is imported in another project say Operandi. 

What do you mean _not working at all_? Operandi can have its own `logging.basicConfig` or `logging.config.fileConfig` or whatever. All OCR-D components simply provide their messages on their respective loggers, the application decided what to do with them (creating handlers and setting levels).

If the application is a standalone OCR-D CLI (→`ocrd_utils.initLogging`) though, then no other components (think Tensorflow) should interfere with logging setup (i.e. handlers+levels). Hence this PR.

> Ideally, the behavior of the root logger, `ocrd` and `ocrd_network` should not be affected by the default logging config file if the user does not do it specifically with a manual logging config file. Otherwise, it bugs programmatically set file handlers and log levels (in `ocrd_network` at least).

Sry, I don't understand that at all.

bertsky:
> > (Perhaps at that time we got distracted by the fact that we still call `initLogging` unconditionally when importing `ocrd_utils.logging`.)
> 
> Yes, the idea was to behave unintrusively as a library but as you point out, in practice we don't

Actually, I was trying to point out that we do.

We do **not** call `initLogging` at module-import time anymore. Which is correct IMO. But then we **can** add a handler to the root logger when we **do** call it.

MehmedGIT:
> What do you mean not working at all?

Not logging anything, and if `initLogging` is not called, the initialization phase of the Operandi Server just freezes.

> Operandi can have its own logging.basicConfig or logging.config.fileConfig or whatever.

Yes, but there is some kind of interference because of the `ocrd_logging.conf` coming from the core. I have to manually provide `ocrd_logging.conf` to avoid some side effects but still not completely. 

I am talking about the experience I had in the past. Maybe these issues are no longer there. I am not sure, because I have my [hack](https://github.com/subugoe/operandi/blob/main/src/utils/operandi_utils/logging.py#L38-L53) to reconfigure the loggers at run-time after the `initLogging` call to prevent any undesired side effects.

> Sry, I don't understand that at all.

I was talking about the `ocrd_logging.conf` file inside `ocrd_utils`.

bertsky:
> Not logging anything, and if `initLogging` is not called, the initialization phase of the Operandi Server just freezes.

I will try to unwrap: Operandi [on startup](https://github.com/subugoe/operandi/blob/6a71a04a26bda6dc8f1ef64d273058f8bba5cd89/src/server/operandi_server/server.py#L77) calls [ocrd_utils.initLogging](https://github.com/subugoe/operandi/blob/6a71a04a26bda6dc8f1ef64d273058f8bba5cd89/src/utils/operandi_utils/utils.py#L14-L30). So you are saying that

1. without ocrd_utils.initLogging, nothing gets logged at all.  
   I would assume that Python will automatically do a `logging.basicConfig` – unless perhaps some rogue library you are using already inits logging (like our Tensorflow example). So you should _normally_ at least get warnings. Have you tried instrumenting to see whether `getLogger('').handlers` shows anything?
2. without ocrd_utils.initLogging, the server (i.e. probably `on_startup_event`) blocks.  
   That's odd. What does the stack trace show in that state?

Also, have you tried calling `logging.basicConfig` or `logging.config.fileConfig` instead of the ocrd_utils version?

> Yes, but there is some kind of interference because of the `ocrd_logging.conf` coming from the core. I have to manually provide `ocrd_logging.conf` to avoid some side effects but still not completely.

I don't understand how our package data log config can interfere at all, unless `ocrd_utils.initLogging` is called.

It sounds like what you describe has been observed when core would still do `initLogging` at import time (which was clearly an antipattern and it does not anymore).

> I am talking about the experience I had in the past. Maybe these issues are no longer there.

I would hope so. Could you please try it out – perhaps even with and without this change here?

> I am not sure, because I have my [hack](https://github.com/subugoe/operandi/blob/main/src/utils/operandi_utils/logging.py#L38-L53) to reconfigure the loggers at run-time after the `initLogging` call to prevent any undesired side effects.

Ok, that's quite involved (I don't fully understand the InterceptHandler). 

_Side effects_ being handlers and levels you don't want? (So it could also be the same issue as above?)

> Ideally, the behavior of the root logger, `ocrd` and `ocrd_network` should not be affected by the default logging config file if the user does not do it specifically with a manual logging config file. Otherwise, it bugs programmatically set file handlers and log levels (in `ocrd_network` at least).
> (I was talking about the `ocrd_logging.conf` file inside `ocrd_utils`.)

So you were merely saying that an optional file config should not interfere with programmatic setup?

I agree – and that's what `initLogging` in its current form seems to achieve. (Just that the programmatic setup is missing the root logger, hence this PR.)

MehmedGIT:
> I don't understand how our package data log config can interfere at all, unless `ocrd_utils.initLogging` is called.

Because it has to be called when using core methods, and there is no workaround to avoid it (may no longer be valid).

> It sounds like what you describe has been observed when core would still do initLogging at import time (which was clearly an antipattern and it does not anymore).

That may be the case, yes. I will have to verify again.

> Side effects being handlers and levels you don't want? (So it could also be the same issue as above?)

Yes, absolutely. Take for instance just core - the `ocrd_network.ProcessingWorker`. From observations - the log level in the constructor seems to be `ERROR` or `WARNING` (not `INFO`) since I can only see messages with that log level forwarded to the logging file of the specific worker. However, the log level of `ocrd_network.ProcessingWorker.publish_result_to_all` method is `INFO`. Because these are normally forwarded to the log file. So when the default `ocrd_logging.conf` is used, although it sets `ocrd_network` logger level to `INFO`, it is not `INFO` everywhere I would expect it to be. The worst part is that I am not sure why. 

The only place where everything is logged as expected to a file is the `ocrd_network.process_helpers.invoke_processor()` because of a new context manager, check [here](https://github.com/OCR-D/core/blob/master/src/ocrd_network/process_helpers.py#L31-L33).

> So you were merely saying that an optional file config should not interfere with programmatic setup?

Yes. It is not that optional in the end since core provides a default for you and the only way to avoid that is to provide a manual logging configuration file.

bertsky:
@kba @MehmedGIT in the last 6 commits, I rewrote the entire thing. I also made sure we have no side effects between tests anymore (which haunted me for quite a while). 

It makes much more sense to me this way – and it works under all circumstances. Please let me know what you think. We can also defer this to v3 if deemed too radical.

MehmedGIT:
Seems some adaptations are needed inside `ocrd_network` module as well. [Here](https://github.com/OCR-D/core/blob/master/src/ocrd_network/process_helpers.py#L31-L33) specifically. The processing job logs have many exceptions such as shown below. Just removing the `initLogging` call does not help. Any ideas?
```python
--- Logging error ---
  File "/usr/local/lib/python3.8/logging/__init__.py", line 1089, in emit
    self.flush()
  File "/usr/local/lib/python3.8/logging/__init__.py", line 1069, in flush
    self.stream.flush()
OSError: [Errno 5] Input/output error
Call stack:
  File "/home/mm/venv38-all/bin/ocrd-cis-ocropy-denoise", line 8, in <module>
    sys.exit(ocrd_cis_ocropy_denoise())
  File "/home/mm/venv38-all/lib/python3.8/site-packages/click/core.py", line 1157, in __call__
  File "/home/mm/venv38-all/lib/python3.8/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/mm/venv38-all/lib/python3.8/site-packages/click/core.py", line 1434, in invoke
  File "/home/mm/venv38-all/lib/python3.8/site-packages/click/core.py", line 783, in invoke
  File "/home/mm/venv38-all/lib/python3.8/site-packages/ocrd_cis/ocropy/cli.py", line 28, in ocrd_cis_ocropy_denoise
    return ocrd_cli_wrap_processor(OcropyDenoise, *args, **kwargs)
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd/decorators/__init__.py", line 71, in ocrd_cli_wrap_processor
    check_and_run_network_agent(processorClass, subcommand, address, database, queue)
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd/decorators/__init__.py", line 196, in check_and_run_network_agent
    processing_worker.start_consuming()
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd_network/processing_worker.py", line 134, in start_consuming
    self.rmq_consumer.start_consuming()
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd_network/rabbitmq_utils/consumer.py", line 45, in start_consuming
    self._channel.start_consuming()
  File "/home/mm/venv38-all/lib/python3.8/site-packages/pika/adapters/blocking_connection.py", line 1883, in start_consuming
    self._process_data_events(time_limit=None)
  File "/home/mm/venv38-all/lib/python3.8/site-packages/pika/adapters/blocking_connection.py", line 2044, in _process_data_events
    self.connection.process_data_events(time_limit=time_limit)
  File "/home/mm/venv38-all/lib/python3.8/site-packages/pika/adapters/blocking_connection.py", line 851, in process_data_events
    self._dispatch_channel_events()
  File "/home/mm/venv38-all/lib/python3.8/site-packages/pika/adapters/blocking_connection.py", line 567, in _dispatch_channel_events
    impl_channel._get_cookie()._dispatch_events()
  File "/home/mm/venv38-all/lib/python3.8/site-packages/pika/adapters/blocking_connection.py", line 1510, in _dispatch_events
    consumer_info.on_message_callback(self, evt.method,
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd_network/processing_worker.py", line 110, in on_consumed_message
    self.process_message(processing_message=processing_message)
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd_network/processing_worker.py", line 187, in process_message
    invoke_processor(
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd_network/process_helpers.py", line 35, in invoke_processor
    run_processor(
  File "/home/mm/repos/core/build/__editable__.ocrd-2.69.0-py3-none-any/ocrd/processor/helpers.py", line 84, in run_processor
    log.debug("Running processor %s", processorClass)
Message: 'Running processor %s'
Arguments: (<class 'ocrd_cis.ocropy.denoise.OcropyDenoise'>,)
```

bertsky:
@MehmedGIT 
> Seems some adaptations are needed inside `ocrd_network` module as well. [Here](https://github.com/OCR-D/core/blob/master/src/ocrd_network/process_helpers.py#L31-L33) specifically. The processing job logs have many exceptions such as shown below. Just removing the `initLogging` call does not help. Any ideas?

I may have broke `force_reinit` (as there is no cleanup during `initLogging` anymore). Perhaps I should call `disableLogging` if both flags (`initialized_flag` and `force_reinit`) are true...

bertsky:
Can you try with [192895a](https://github.com/OCR-D/core/pull/1288/commits/192895afd5c38314e9a7de59b866b8bfedf1f523) again?

bertsky:
> The mets server's log file is not logging after the initialization though.

What do you expect to see? Our logger is not used for anything beyond `DEBUG` level. And we do _not_ set `log_config` of Uvicorn (and even disable its `access_log`), so I would expect nothing.

bertsky:
See [this recipe](https://github.com/fastapi/fastapi/issues/1508) for how we could include more messages from Uvicorn.

MehmedGIT:
> > The mets server's log file is not logging after the initialization though.
> 
> What do you expect to see? Our logger is not used for anything beyond `DEBUG` level. And we do _not_ set `log_config` of Uvicorn (and even disable its `access_log`), so I would expect nothing.

Ah, okay. I am expecting to see the `INFO` level logs of the [mets_server](https://github.com/OCR-D/core/blob/master/src/ocrd/mets_server.py).

bertsky:
> I am expecting to see the `INFO` level logs of the [mets_server](https://github.com/OCR-D/core/blob/master/src/ocrd/mets_server.py).

Are you perhaps still using a METS Server installation before [your extensive logging change](https://github.com/OCR-D/core/commit/e0ff4ebd3ea200a73b200375e75a4886eb1941fc)? (That would explain why you get `Starting up` instead of `Configuring`...)

kba (approved):

src/ocrd_utils/ocrd_logging.conf
```diff
@@ -56,30 +56,30 @@ handlers=consoleHandler,fileHandler
...
-propagate=0
+#propagate=0
 
 #
 # logger tensorflow
 #
 [logger_ocrd_tensorflow]
 level=ERROR
-handlers=consoleHandler
+handlers=
```
kba, line 74: What are the semantics of setting handlers to an empty value? It's not ovious from https://docs.python.org/3/library/logging.config.html for me - I assume it's incremental, i.e. does not explicitly add a handler but does not remove existing handlers either?

bertsky: Depends on how `fileConfig` gets run. The default is `disable_existing_loggers=True`, i.e. before parsing, all loggers (and their handlers) are removed, as done in our `disableLogging`. See [documentation](https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig).

Still, what was surprising to me is that one _must_ add a `handlers` statement, it cannot be omitted – even if one just wants the messages to get forwarded up in the hierarchy without extra handlers.

The config file is documented [here](https://docs.python.org/3/library/logging.config.html#logging-config-fileformat) BTW.

PR #1289, opened 2024-10-10
CLI decorator: only import ocrd_network when needed

bertsky:
Avoids about 17 MB RSS of unnecessary networking imports when _not_ using the networking subcommands on the CLI.


kba (approved):
`import` below module top-level is normally frowned upon but in this case it makes sense, no need to pull all of that in if it is not used. LGTM

MehmedGIT (approved):

PR #1290, opened 2024-10-10
create PyPI CD

bertsky:


kba (approved):
We'll see when I do v2.70.0 now 🤞

bertsky:
> We'll see when I do v2.70.0 now 🤞

yes, but once merged, we can also trigger manually – it should then fail during upload, saying the version is already present.

Unfortunately, with GHA it's not possible to "backport" this workflow. So in workflow_dispatch you can select any branch or tag, but if that workflow was not present in that version, you simply get informed that it's not available. So obviously the manual trigger is not much worth...

PR #1292, opened 2024-10-14
Fix broken ocrd_network processing worker logs

MehmedGIT:
The processing worker logger acted weirdly for some time. Pay attention to the log levels produced by `self.log.__dict__`

```python
2024-10-14 14:57:54,874.874 WARNING ocrd_network.processing_worker - ---------- __init__() CONFIGURED LOGGER: {'filters': [], 'name': 'ocrd_network.processing_worker', 'level': 0, 'parent': <Logger ocrd_network (WARNING)>, 'propagate': True, 'handlers': [<FileHandler /tmp/ocrd_network_logs/processing_workers/worker.843428.ocrd-cis-ocropy-binarize.log (NOTSET)>], 'disabled': False, '_cache': {10: False, 20: False, 40: True}, 'manager': <logging.Manager object at 0x7fa652ba9b50>}
2024-10-14 14:57:54,879.879 WARNING ocrd_network.processing_worker - ---------- start_consuming() CONFIGURED LOGGER: {'filters': [], 'name': 'ocrd_network.processing_worker', 'level': 0, 'parent': <Logger ocrd_network (WARNING)>, 'propagate': True, 'handlers': [<FileHandler /tmp/ocrd_network_logs/processing_workers/worker.843428.ocrd-cis-ocropy-binarize.log (NOTSET)>], 'disabled': False, '_cache': {10: False, 20: False, 40: True, 30: True}, 'manager': <logging.Manager object at 0x7fa652ba9b50>}
2024-10-14 14:58:12,140.140 WARNING ocrd_network.processing_worker - ---------- on_consumed_message() CONFIGURED LOGGER: {'filters': [], 'name': 'ocrd_network.processing_worker', 'level': 0, 'parent': <Logger ocrd_network (WARNING)>, 'propagate': True, 'handlers': [<FileHandler /tmp/ocrd_network_logs/processing_workers/worker.843428.ocrd-cis-ocropy-binarize.log (NOTSET)>], 'disabled': False, '_cache': {10: False, 20: False, 40: True, 30: True}, 'manager': <logging.Manager object at 0x7fa652ba9b50>}
2024-10-14 14:58:12,141.141 WARNING ocrd_network.processing_worker - ---------- process_message() CONFIGURED LOGGER: {'filters': [], 'name': 'ocrd_network.processing_worker', 'level': 0, 'parent': <Logger ocrd_network (WARNING)>, 'propagate': True, 'handlers': [<FileHandler /tmp/ocrd_network_logs/processing_workers/worker.843428.ocrd-cis-ocropy-binarize.log (NOTSET)>], 'disabled': False, '_cache': {10: False, 20: False, 40: True, 30: True}, 'manager': <logging.Manager object at 0x7fa652ba9b50>}
2024-10-14 14:58:15,279.279 WARNING ocrd_network.processing_worker - ---------- publish_result_to_all() CONFIGURED LOGGER: {'filters': [], 'name': 'ocrd_network.processing_worker', 'level': 0, 'parent': <Logger ocrd_network (INFO)>, 'propagate': True, 'handlers': [<FileHandler /tmp/ocrd_network_logs/processing_workers/worker.843428.ocrd-cis-ocropy-binarize.log (NOTSET)>], 'disabled': False, '_cache': {20: True, 40: True}, 'manager': <logging.Manager object at 0x7fa652ba9b50>}
2024-10-14 14:58:15,279.279 INFO ocrd_network.processing_worker - Result message: {'job_id': 'b3efe133-7c7d-4316-abc1-d99c35c5f62b', 'state': 'SUCCESS', 'workspace_id': '', 'path_to_mets': '/home/mm/repos/ocrd_network_tests/ws29/data/mets.xml'}
2024-10-14 14:58:15,280.280 INFO ocrd_network.processing_worker - Publishing result to internal callback url (Processing Server): None
2024-10-14 14:58:15,280.280 INFO ocrd_network.processing_worker - Posting result message to callback_url "http://localhost:8080/result_callback"
2024-10-14 14:58:15,317.317 INFO ocrd_network.processing_worker - Response from callback_url "<Response [200]>"
2024-10-14 14:58:15,318.318 INFO ocrd_network.processing_worker - Successfully processed RabbitMQ message
2024-10-14 14:58:15,320.320 WARNING ocrd_network.processing_worker - ---------- on_consumed_message() CONFIGURED LOGGER: {'filters': [], 'name': 'ocrd_network.processing_worker', 'level': 0, 'parent': <Logger ocrd_network (INFO)>, 'propagate': True, 'handlers': [<FileHandler /tmp/ocrd_network_logs/processing_workers/worker.843428.ocrd-cis-ocropy-binarize.log (NOTSET)>], 'disabled': False, '_cache': {20: True, 40: True, 30: True, 10: False}, 'manager': <logging.Manager object at 0x7fa652ba9b50>}
```

To reproduce it is enough to print the logger object at the beginning of each method:
```python
self.log.warning(f"{self.log.__dict__}")
```

Seems when jiggling with the logging in the past, the `initLogging()` call was removed from the processing worker. The logger level is still set to `INFO` at some point, but an implicit call to `initLogging()` sets the logger level earlier.

### DIFF
--- a/src/ocrd_network/processing_server.py
+++ b/src/ocrd_network/processing_server.py
@@ -79,7 +79,6 @@ class ProcessingServer(FastAPI):
     """
 
     def __init__(self, config_path: str, host: str, port: int) -> None:
-        initLogging()
         self.title = "OCR-D Processing Server"
         super().__init__(
             title=self.title,
@@ -87,6 +86,7 @@ class ProcessingServer(FastAPI):
             on_shutdown=[self.on_shutdown],
             description="OCR-D Processing Server"
         )
+        initLogging()
         self.log = getLogger("ocrd_network.processing_server")
         log_file = get_processing_server_logging_file_path(pid=getpid())
         configure_file_handler_with_formatter(self.log, log_file=log_file, mode="a")
@@ -156,7 +156,7 @@ class ProcessingServer(FastAPI):
             queue_names = self.deployer.find_matching_network_agents(
                 worker_only=True, str_names_only=True, unique_only=True
             )
-            self.log.debug(f"Creating message queues on RabbitMQ instance url: {self.rabbitmq_url}")
+            self.log.info(f"Creating message queues on RabbitMQ instance url: {self.rabbitmq_url}")
             create_message_queues(logger=self.log, rmq_publisher=self.rmq_publisher, queue_names=queue_names)
 
             self.deployer.deploy_network_agents(mongodb_url=self.mongodb_url, rabbitmq_url=self.rabbitmq_url)
@@ -168,6 +168,7 @@ class ProcessingServer(FastAPI):
         uvicorn_run(self, host=self.hostname, port=int(self.port))
 
     async def on_startup(self):
+        self.log.info(f"Initializing the Database on: {self.mongodb_url}")
         await initiate_database(db_url=self.mongodb_url)
 
     async def on_shutdown(self) -> None:
--- a/src/ocrd_network/processing_worker.py
+++ b/src/ocrd_network/processing_worker.py
@@ -9,12 +9,12 @@ is a single OCR-D Processor instance.
 """
 
 from datetime import datetime
-from os import getpid
+from os import getpid, getppid
 from pika import BasicProperties
 from pika.adapters.blocking_connection import BlockingChannel
 from pika.spec import Basic
 
-from ocrd_utils import getLogger
+from ocrd_utils import getLogger, initLogging
 from .constants import JobState
 from .database import sync_initiate_database, sync_db_get_workspace, sync_db_update_processing_job, verify_database_uri
 from .logging_utils import (
@@ -35,14 +35,16 @@ from .utils import calculate_execution_time, post_to_callback_url
 
 class ProcessingWorker:
     def __init__(self, rabbitmq_addr, mongodb_addr, processor_name, ocrd_tool: dict, processor_class=None) -> None:
+        initLogging()
         self.log = getLogger(f'ocrd_network.processing_worker')
         log_file = get_processing_worker_logging_file_path(processor_name=processor_name, pid=getpid())
         configure_file_handler_with_formatter(self.log, log_file=log_file, mode="a")
 
         try:
             verify_database_uri(mongodb_addr)
-            self.log.debug(f'Verified MongoDB URL: {mongodb_addr}')
+            self.log.info(f'Verified MongoDB URL: {mongodb_addr}')
             self.rmq_data = verify_and_parse_mq_uri(rabbitmq_addr)
+            self.log.info(f'Verified RabbitMQ URL: {rabbitmq_addr}')
         except ValueError as error:
             msg = f"Failed to parse data, error: {error}"
             self.log.exception(msg)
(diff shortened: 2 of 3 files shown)


src/ocrd_network/processing_server.py
```diff
@@ -79,14 +79,14 @@ class ProcessingServer(FastAPI):
...
     def __init__(self, config_path: str, host: str, port: int) -> None:
-        initLogging()
         self.title = "OCR-D Processing Server"
         super().__init__(
             title=self.title,
             on_startup=[self.on_startup],
             on_shutdown=[self.on_shutdown],
             description="OCR-D Processing Server"
         )
+        initLogging()
```
bertsky, line 89: Oh, any particular reason for calling our init before the superclass? (I cannot find anything in `fastapi.Application` or its superclass `starlette.Starlette` that would effectivate `logging.basicConfig`, and thus interfere with our setup, but maybe I am missing something...)

Anyways, just curious.

bertsky (approved):
Wow – thanks for digging!

Still strange why the correct log level would be set somewhere between `process_message` and `publish_result_to_all`...

src/ocrd_network/processing_server.py
MehmedGIT: > Oh, any particular reason for calling our init before the superclass?

Do you mean calling the init after the superclass since that is the change? I moved that to ensure that the superclass does not currently and would not potentially affect the logging outcome in the future (with new releases).

MehmedGIT: Worth mentioning is that this PR does not completely fix the logging of the Processing Server. The log file still contains fewer logs than the ones appearing on the console.

bertsky: > > Oh, any particular reason for calling our init before the superclass?
> 
> Do you mean calling the init after the superclass since that is the change? 

Yes, I did (just forgot the word `not`), sry

> I moved that to ensure that the superclass does not currently and would not potentially affect the logging outcome in the future (with new releases).

Hmm. But is that (FastAPI interfering with our logging setup) not exactly what this change risks bringing about? It seems you assume that our `initLogging` can safely undo whatever other logging setup has been done at that point.

bertsky: > Worth mentioning is that this PR does not completely fix the logging of the Processing Server. The log file still contains fewer logs than the ones appearing on the console.

Could you try tracing all the `.handlers` objects of the relevant loggers?

MehmedGIT: > Hmm. But is that (FastAPI interfering with our logging setup) not exactly what this change risks bringing about?

Now I see what you mean. 

> It seems you assume that our initLogging can safely undo whatever other logging setup has been done at that point.

I assume that we can stay flexible by still being able to overwrite FastAPI loggers with the `ocrd_logging.conf` as well. But that of course could then break FastAPI loggers if not implemented properly. 

I am not sure which is preferable/safe in our case: pre- or post- superclass `initLogging` call.

MehmedGIT: > Could you try tracing all the .handlers objects of the relevant loggers?

Yes, I will try that.

bertsky: > I assume that we can stay flexible by still being able to overwrite FastAPI loggers with the `ocrd_logging.conf` as well. But that of course could then break FastAPI loggers if not implemented properly.
> I am not sure which is preferable/safe in our case: pre- or post- superclass `initLogging` call.

Yes. Or conversely, if they get to init after us, our preferences could be lost. So we do have to assume _something_.

Since they (fastapi and starlette) currently do not initialize logging themselves, not even inadvertently (e.g. by calling `logging.debug` instead of `logger.debug` anywhere), as any well behaved library should be expected to, I guess your change _is_ actually preferable: they get to set default levels for their loggers, then we get to override whatever we want plus setup our handlers and formatters.

Issue #1223, opened 2024-05-14
warnings in processor output garble dump-json


bertsky:
We have…
```
ocrd process \
"cis-ocropy-binarize -I DEFAULT -O OCR-D-BIN"
"anybaseocr-crop -I OCR-D-BIN -O OCR-D-CROP"
"skimage-binarize -I OCR-D-CROP -O OCR-D-BIN2 -P method li"
"skimage-denoise -I OCR-D-BIN2 -O OCR-D-BIN-DENOISE -P level-of-operation page"
"tesserocr-deskew -I OCR-D-BIN-DENOISE -O OCR-D-BIN-DENOISE-DESKEW -P operation_level page"
"cis-ocropy-segment -I OCR-D-BIN-DENOISE-DESKEW -O OCR-D-SEG -P level-of-operation page"
"cis-ocropy-dewarp -I OCR-D-SEG -O OCR-D-SEG-LINE-RESEG-DEWARP"
"calamari-recognize -I OCR-D-SEG-LINE-RESEG-DEWARP -O OCR-D-OCR -P checkpoint_dir qurator-gt4histocr-1.0"
The TensorFlow library was compiled to use AVX instructions, but these aren't available on your machine.
09:00:38.996 ERROR ocrd.utils.get_ocrd_tool_json - ocrd-calamari-recognize --dump-json produced invalid JSON: Expecting value: line 1 column 1 (char 0)
Traceback (most recent call last):
File "/usr/local/bin/ocrd", line 8, in <module>
sys.exit(cli())
File "/usr/local/lib/python3.8/site-packages/click/[core.py](http://core.py/)", line 1157, in call
return self.main(*args, **kwargs)
File "/usr/local/lib/python3.8/site-packages/click/[core.py](http://core.py/)", line 1078, in main
rv = self.invoke(ctx)
File "/usr/local/lib/python3.8/site-packages/click/[core.py](http://core.py/)", line 1688, in invoke
return _process_result(sub_ctx.command.invoke(sub_ctx))
File "/usr/local/lib/python3.8/site-packages/click/[core.py](http://core.py/)", line 1434, in invoke
return ctx.invoke(self.callback, **ctx.params)
File "/usr/local/lib/python3.8/site-packages/click/[core.py](http://core.py/)", line 783, in invoke
return __callback(*args, **kwargs)
File "/usr/local/lib/python3.8/site-packages/ocrd/cli/[process.py](http://process.py/)", line 31, in process_cli
[...]

PR #1295, opened 2024-11-20
run initLogging unconditionally

kba:
As discussed in the Tech Call today, we do still need to call `initLogging` to prevent other modules from adding handlers that might interfere with STDOUT.

And the `tests/cli/test_log.py` needs to call `disableLogging` before the first `initLogging` because otherwise pytest will highjack the logging :/

### DIFF
--- a/src/ocrd_utils/logging.py
+++ b/src/ocrd_utils/logging.py
@@ -205,3 +205,8 @@ def disableLogging(silent=not config.OCRD_LOGGING_DEBUG):
     # Python default log level is WARNING
     logging.root.setLevel(logging.WARNING)
 
+# Initialize logging now to prevent any other libraries (tensorflow,
+# matplotlib ...) from adding their own root handler, which might mangle the
+# STDOUT (our logging will log to STDERR so as not to interfere with
+# --dump-json etc.)
+initLogging()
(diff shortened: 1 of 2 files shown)


src/ocrd_utils/logging.py
```diff
@@ -205,3 +205,8 @@ def disableLogging(silent=not config.OCRD_LOGGING_DEBUG):
     # Python default log level is WARNING
     logging.root.setLevel(logging.WARNING)
 
+# Initialize logging now to prevent any other libraries (tensorflow,
+# matplotlib ...) from adding their own root handler, which might mangle the
+# STDOUT (our logging will log to STDERR so as not to interfere with
+# --dump-json etc.)
+initLogging()
```
bertsky, line 212: No, that's not what I meant. See https://github.com/OCR-D/core/issues/1223#issuecomment-2470587136

bertsky (changes requested):

src/ocrd_utils/logging.py
bertsky: Alternatively, we could still do `initLogging` on the module level, though not in `ocrd_utils.logging` but rather `ocrd.decorators`, which arguably will only be imported by applications that do need log handlers.

PR #1302, opened 2025-02-03
Fix processor executor shutdown

bertsky:
fixes issues with v3 processor CIs (during pytest shutdown)

### DIFF
--- a/src/ocrd/processor/base.py
+++ b/src/ocrd/processor/base.py
@@ -130,7 +130,8 @@ class DummyExecutor:
     def __init__(self, initializer=None, initargs=(), **kwargs):
         initializer(*initargs)
     def shutdown(self, **kwargs):
-        pass
+        # allow gc to catch processor instance (unless cached)
+        _page_worker_set_ctxt(None, None)
     def submit(self, fn, *args, **kwargs) -> DummyFuture:
         return DummyFuture(fn, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


bertsky:
Note: we can (update and merge) v3 for ocrd_calamari, ocrd_kraken, ocrd_cis, ocrd_tesserocr as soon as this is merged and released.

src/ocrd/processor/base.py
```diff
@@ -505,7 +506,7 @@ def process_workspace(self, workspace: Workspace) -> None:
                 # set up multitasking
                 max_workers = max(0, config.OCRD_MAX_PARALLEL_PAGES)
                 if self.max_workers > 0 and self.max_workers < config.OCRD_MAX_PARALLEL_PAGES:
-                    self._base_logger.info("limiting number of threads from %d to %d", max_workers, self.max_workers)
+                    self._base_logger.info("limiting number of workers from %d to %d", max_workers, self.max_workers)
```
kba, line 509: Should we also do `s,threads,workers,` in the description of `OCRD_MAX_PARALLEL_PAGES` config option?

kba (approved):

src/ocrd/processor/base.py
bertsky: indeed!
see 217059db

bertsky:
ah, too late!

PR #1303, opened 2025-02-05
Fix ocrd_network client run argument order

bertsky:


bertsky:
@kba should I make a v2 backport of [0d277ac](https://github.com/OCR-D/core/pull/1303/commits/0d277ac18cdfaa5e296fe7bc63e635be568d2841)?

I should probably create an issue for this, but perhaps someone can get me on track here: Shouldn't we also allow the user to pass the `ocrd_utils.config` options to the server, somehow? (In v3, you'll remember, we introduced fine-grained controls like OCRD_MISSING_INPUT, OCRD_MISSING_OUTPUT, OCRD_EXISTING_OUTPUT, that should in principle be configurable on a case-by-case basis. Without any such delegation, one would have to live with the defaults set on the server side, so there's not much use for them in the network scenario...)

MehmedGIT (approved):

bertsky:
> I should probably create an issue for this, but perhaps someone can get me on track here: Shouldn't we also allow the user to pass the `ocrd_utils.config` options to the server, somehow? (In v3, you'll remember, we introduced fine-grained controls like OCRD_MISSING_INPUT, OCRD_MISSING_OUTPUT, OCRD_EXISTING_OUTPUT, that should in principle be configurable on a case-by-case basis. Without any such delegation, one would have to live with the defaults set on the server side, so there's not much use for them in the network scenario...)

Another problem is with how parameter resources get resolved now. In the CLI setting, we allowed an ambiguity between JSON literals and file names, so the latter would have to be resolved to absolute paths (for the various resource locations), which in turn obviously dependend on the processor class (because of the custom `module` location):

https://github.com/OCR-D/core/blob/4d33491ab4e515a60e240cbab2b1510b4e58fa02/src/ocrd/decorators/__init__.py#L95-L103

But in the network setting, where parameters can be added to the processing or workflow request, no such resolution (from resource name to resource path) takes place –
- On the server side, there is only a parameter validation step  
  https://github.com/OCR-D/core/blob/4d33491ab4e515a60e240cbab2b1510b4e58fa02/src/ocrd_network/server_utils.py#L220
- On the client side, there is only the JSON string vs file disambiguation, but no prior resolution.  
  https://github.com/OCR-D/core/blob/4d33491ab4e515a60e240cbab2b1510b4e58fa02/src/ocrd_network/cli/client.py#L142

Did we abandon the entire mechanism, or am I missing something?

kba (approved):

kba:
> @kba should I make a v2 backport of [0d277ac](https://github.com/OCR-D/core/pull/1303/commits/0d277ac18cdfaa5e296fe7bc63e635be568d2841)?

It depends - does the previous behavior trigger a bug? AFAICT we're only calling the method with kwargs or am I missing something?

PR #1304, opened 2025-02-13
fixup 39f76258: avoid explicit .stop() for QueueListener

bertsky:
Needed for proper shutdown of page-parallel processors, esp. in pytest context (think CI).

Sorry to drag this on, but there were still fuzzy deadlocks in 3.0.2 that I did not see earlier.

It appears that is is sufficient to just let go of the QueueListener, without explicitly stopping it.

Tested with ocrd_cis (CLI), ocrd_kraken (Pytest), ocrd_tesserocr (both), ocrd_calamari (Pytest).

### DIFF
--- a/src/ocrd/processor/base.py
+++ b/src/ocrd/processor/base.py
@@ -534,15 +534,20 @@ class Processor():
                     # forward messages from log queue (in subprocesses) to all root handlers
                     log_listener = logging.handlers.QueueListener(log_queue, *logging.root.handlers, respect_handler_level=True)
                     log_listener.start()
+                tasks = None
                 try:
                     self._base_logger.debug("started executor %s with %d workers", str(executor), max_workers or 1)
                     tasks = self.process_workspace_submit_tasks(executor, max_seconds)
                     stats = self.process_workspace_handle_tasks(tasks)
                 finally:
                     executor.shutdown(kill_workers=True, wait=False)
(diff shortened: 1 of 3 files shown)


src/ocrd/processor/base.py
```diff
@@ -534,15 +534,20 @@ def process_workspace(self, workspace: Workspace) -> None:
...
                     tasks = self.process_workspace_submit_tasks(executor, max_seconds)
                     stats = self.process_workspace_handle_tasks(tasks)
                 finally:
                     executor.shutdown(kill_workers=True, wait=False)
+                    self._base_logger.debug("stopped executor %s after %d tasks", str(executor), len(tasks) if tasks else -1)
                     if max_workers > 1:
-                        log_listener.stop()
-                        del log_listener
+                        # can cause deadlock:
+                        #log_listener.stop()
```
kba, line 547: Are there ramifications when running processing workers/cached instances? I.e. are these garbage-collected or will this lead to a memory leak or exhaustion of resources over time if the `log_listener` is not explicitly closed?

bertsky: Good question!

I would say no, because the log_listener and executor are not part of the processor instance (i.e. not referenced). They are local bindings of `Processor.process_workspace`, so they should indeed be GCed after that (or `run_processor` or whatever entry point).

But I have not even tested the combination of processor instance caching and page-parallel processing, yet. Will do!

kba (approved):
Many thanks!

Issue #1294, opened 2024-11-12
add resmgr server with fastapi/uvicorn


bertsky:
> Elaborating a bit on option 2: of course, the (generated) docker-compose.yml for each module could also provide an additional server entry point – a simple REST API wrapper for resmgr CLI. Its (generated) volume and variable config would have to match the respective Processing Worker (or Processor Server) to be used. But the local resmgr would not need to "know" anything beyond what it can see in its thin container – a local ocrd-all-tool.json and ocrd-all-module-dir.json precomputed for the processors of that module at build time, plus the filesystem in that container and mounted volumes.
>
> In addition, to get the same central resmgr user experience (for all processor executables at the same time), one would still need 
> - either a single server (with resmgr-like endpoints or even providing some `/discovery/processor/resources`) which delegates to the individual `resmgr` servers,
> - or an intelligent `resmgr` client doing the same.
>
> Regardless, crucially, this central component needs to _know_ about all the deployed resmgr services – essentially holding a mapping from processor executables to module resmgr server host-port pairs. This could be generated along with the docker-compose.yml (in a new format like `ocrd-all-module-dir.json`), or the latter even gets parsed directly.

_Originally posted by @bertsky in https://github.com/OCR-D/ocrd_all/issues/69#issuecomment-2470095507_

bertsky:
Implementation could borrow heavily from ocrd.mets_server and ocrd.cli.workspace.workspace_serve_cli, with endpoints 1:1 providing ocrd.cli.resmgr commands.

Could you please take this on @joschrew?

bertsky:
@joschrew please read my [comprehensive analysis on the resmgr problem](https://github.com/OCR-D/ocrd_all/issues/69#issuecomment-2406157599) for context. We are _not_ anywhere near any working solution at the moment.

This is _not_ about whether or how a processor can resolve (or locate) requested resources (i.e. models) at runtime.

It is about the functionality of the `ocrd resmgr` CLI made available within the thin container regime, i.e. (for all processors)
- listing installed models,
- listing additional models available for download, and
- downloading models.

I just gave the METS Server example because it contains a simple FastAPI + Uvicorn structure that you can borrow from. (Of course, the same can be found in the Processing Server, but there it is spread across multiple modules.)

joschrew:
I basically just can repeat myself, I already tried to understand what was written in the linked issue you mentioned.

> This is not about whether or how a processor can resolve (or locate) requested resources (i.e. models) at runtime.

[...]

Issue #1251, opened 2024-07-04
resmgr list-installed only knows about 3 processors with preconfigured resources


bertsky:
But surely, such a user database was at least created by that call. And if you did not run any `list-available` prior to that, then that database would be just a mirror of the distributed ocrd/resource_list.yml (hence only those 3 processors).

So we just uncovered another serious bug: initialisation does not search the PATH for ocrd-* executables, only `list-available` does. But without these database entries, `list-installed` never even attempts to look for other executables!

bertsky:
explanation: ResourceManager inits its database from the predistributed `ocrd/resource_list.yml`:
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/resource_manager.py#L42-L47

new database entries only get made by either
1. `list-available` (with some executable glob pattern)
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/resource_manager.py#L100-L109
2. `list-installed` (when explicitly naming the executable)
https://github.com/OCR-D/core/blob/79c61e303c87f229d5c96aedc0da31ef82b0f5d3/src/ocrd/resource_manager.py#L168

so not even a `download` adds the respective entry!

expectation: `list-installed *` or just `list-installed` (without a name) should look for all executables in PATH, regardless of existing database entries.

perhaps, considering #1250, we could make an exception if some `ocrd-all-tool.json` is installed: in that case, one should _not_ waste time searching PATH, but can just pick the precomputed list.

MehmedGIT:
I have removed the `~/.config/ocrd/resources.yml`, then installed the core again from the current master branch. This is the result:
```sh
(venv38-core) mm@MM-Notebook:~/repos/core$ ocrd resmgr list-installed
12:38:19.387 INFO ocrd.resource_manager - ocrd-cis-ocropy-recognize resource '3gs.csv.gz' (/home/mm/venv38-all/lib/python3.8/site-packages/ocrd_cis/data/3gs.csv.gz) not a known resource, creating stub in /home/mm/.config/ocrd/resources.yml'
12:38:20.374 INFO ocrd.resource_manager - ocrd-cis-ocropy-recognize resource 'config.json' (/home/mm/venv38-all/lib/python3.8/site-packages/ocrd_cis/data/config.json) not a known resource, creating stub in /home/mm/.config/ocrd/resources.yml'
[...]

PR #1309, opened 2025-02-18
Implementation of the resource manager server (issue #1294)

MehmedGIT:
A PR draft for addressing #1294. Still in progress.

What has been added so far:
- Support `ocrd network resmgr-server --address host:port` for triggering Resource Manager Server (RMS) in the background
- For each host mentioned in the PS config file, the deployer will deploy a resource manager server on port 45555 of that host
- Basic `list-available` functionality for the RMS
- Basic `list-installed` functionality for the RMS
- A partial fix for #1251 (check the detailed comment from @bertsky [here](https://github.com/OCR-D/core/pull/1309#issuecomment-2717959747))
- Refactoring of core resource manager


src/ocrd_network/resource_manager_server.py
```diff
@@ -0,0 +1,107 @@
...
+        self.add_api_routes()
+
+        base_router = APIRouter()
+        base_router.add_api_route(
+            path="/",
+            endpoint=self.home_page,
+            methods=["GET"],
+            status_code=status.HTTP_200_OK,
+            summary="Get information about the processing server"
+        )
```
kba, line 40: Redundant to `add_api_routes()`, no?

src/ocrd_network/resource_manager_server.py
```diff
@@ -0,0 +1,107 @@
...
+
+    async def home_page(self):
+        message = f"The home page of the {self.title}"
+        json_message = {
+            "message": message,
+            "time": datetime.now().strftime("%Y-%m-%d %H:%M")
+        }
+        return json_message
+
+    async def list_available_models(self, executable: Any = None, dynamic: bool = True, name: Any = None):
```
kba, line 92:
```suggestion
    async def list_available_resources(self, executable: Any = None, dynamic: bool = True, name: Any = None):
```

Since it's not just about models. Same for the other methods.

kba (commented):
So far so good, though I still need to understand how exactly this works

bertsky:
@MehmedGIT indeed, this calls for more aggressive refactoring. IMO the high-level `ocrd.cli.resmgr.download` (checking executable is installed, checking resource is available, discerning local paths from URLs, resolving target `location`, providing resulting name/path) should become independent of the CLI, i.e. part of `ocrd.resource_manager` (but not necessarily OcrdResourceManager), and get passed the logger and `progress_cb` as kwargs.

Perhaps like this:
```python
def download(
    logger: Optional(logging.Logger),
    progress_cb: Callable[[int], Any],
    resmgr: OcrdResourceManager, 
    any_url: Optional[str], 
    no_dynamic: bool, 
    resource_type: str,
    path_in_archive: str,
    allow_uninstalled: bool,
    overwrite: bool,
    location: str,
    executable: str,
    name: str
):
    # current implementation of cli.resmgr.download
```
(or perhaps as kwargs with their own defaults, perhaps returning the fpath and usable resource name as tuple instead of just logging it)

The ResourceManagerServer could then call this to its own liking.

MehmedGIT:
@bertsky, agree. For now, I will mimick the `ocrd.cli.resmgr.download` method and append the logger print statements to a list, which will be returned. Instead of `sys.exit(1)` I will raise HTTPException and will get rid of the progress bar but still keep the database till I decide how to refactor that away properly.

bertsky:
Indeed, the new/refactored `download` would also need a general exception (just `ValueError`?) instead of exit.

Perhaps it's a good time to remove the entire loathed user database facility from resmgr, too? (cf. https://github.com/OCR-D/core/issues/1251)

MehmedGIT:
> Indeed, the new/refactored download would also need a general exception (just ValueError?) instead of exit.

HTTPException instead of ValueError; we do not want to crash the resource manager server on a wrong request, right?

> Perhaps it's a good time to remove the entire loathed user database facility from resmgr, too? (cf. https://github.com/OCR-D/core/issues/1251)

I would suggest reconsidering the wildcard (`'*'`) options as well and requiring the processor and model name. Otherwise, the return of a response may take more than 30 minutes if a wildcard is used for a processor name. At least, it does with the current implementation.

MehmedGIT:
If @kba has no objections (since he was going to create a PR to fix the resmgr), I will do some basic refactoring without breaking the CLI behaviour. 

> Perhaps the PS discovery can re-implement a similar behavior for convenience in the future

I think for listing available/installed the wildcard is not an issue for now (but maybe will become after refactoring the database away).

MehmedGIT:
https://github.com/OCR-D/core/pull/1309/commits/bb0b0cd2f8d25e7dd6b9a7b82029361aaf780a5b should fix the missing entries for the `'*'` glob. `ocrd resmgr download '*'` should now yield the expected behaviour, of course, after running first `ocrd resmgr list-available` and potentially `ocrd resmgr list-installed` to update the `resources.yml`. 

The issue was that not all entries found were registered properly in the database by invoking `self.save_user_list()`. Although that method is called inside `self.add_to_user_database()`, the latter was called inside an if case, and not everytime a new resource was found! Triggering the `self.save_user_list()` in the end of `list_installed()` and `list_available()` does the job.

src/ocrd/cli/resmgr.py
```diff
@@ -119,16 +118,33 @@ def download(any_url, no_dynamic, resource_type, path_in_archive, allow_uninstal
...
+                      f"refusing to install to invalid location")
+            sys.exit(1)
+        if location == 'module':
+            basedir = get_moduledir(this_executable)
+            if not basedir:
+                basedir = resmgr.location_to_resource_dir('data')
+        else:
+            # TODO @mehmedGIT: The basedir always defaults to CWD when location is not passed from the CLI,
+            #  this may be an undesired default behaviour. Consider setting that up with a default path somewhere.
+            #  For example, RESOURCES_DIR_SYSTEM or config.XDG_DATA_HOME, for consistency and predictability.
```
kba, line 136: I would be open to changing the default to `config.XDG_DATA_HOME` but I think there was a reason for this default behavior but cannot remember it now. @bertsky do you?

src/ocrd/cli/resmgr.py
```diff
@@ -119,16 +118,33 @@ def download(any_url, no_dynamic, resource_type, path_in_archive, allow_uninstal
                 'path_in_archive': path_in_archive}]
             )]
     for this_executable, this_reslist in reslist:
+        if not location:
+            # TODO @mehmedGIT: What if there is no 'resource_locations' in this executable ocrd tool json?
+            #  Then the location will always default to None
```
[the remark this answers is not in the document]
MehmedGIT: Oh, I see. That comes from the `type=click.Choice([])`, where the first option of the list is also the default.

src/ocrd/cli/resmgr.py
MehmedGIT: We can still keep the default to `CWD` if `config.XDG_DATA_HOME` was never set. I would still prefer the default to be `RESOURCES_DIR_SYSTEM`, i.e., `/usr/local/share/ocrd-resources`, if `XDG_DATA_HOME` is None.

MehmedGIT: Indeed, as in the previous comment, I overlooked that `type=click.Choice([])` defaults to the first entry.

MehmedGIT: > Based on the above, `location` is set to the first allowable among `resource_locations`, which most often is just data. So I don't understand what you mean by `defaults to CWD`.

This is not true. If no location is provided from the CLI, it defaults to None. Maybe in the past, the first option was click.Choice([]) was considered the default, but that is no longer true unless specified manually with `default='data'`

src/ocrd/resource_manager.py
```diff
@@ -191,7 +192,7 @@ def list_installed(self, executable=None):
                     }
                 else:
                     resdict = self.add_to_user_database(this_executable, res_filename, resource_type=res_type)
-                resdict['path'] = str(res_filename)
+                # resdict['path'] = str(res_filename)
```
bertsky, line 195: Why not store the path into the returned list of dicts anymore? (So far, it seems to be used only in `print_resources`, but why not?)

MehmedGIT: @bertsky, since the `self.save_user_list()` method writes to the `resources.yml`, the `path` occurred there as a key, which led to the failure of the `OcrdResourceListValidator` and consequential loads. Unfortunately, simply doing the saving before assigning a new `path` key also does not help. I did not want to make a deep copy of the entire database for the extra `path` key output. Neither wanted to modify the `ocrd_tool.schema.yml` by adding an extra `path` field. I will not get rid of the `path` completely; I just need to figure out how to achieve the same behaviour optimally. Maybe that will become clear after I refactor the database itself.

bertsky: Oh, I see. Yes, I also would not like to see `path` in the schema (as this syntax is meant to be shared between ocrd-tool.json developer descriptions and resource.yml installations).

But in this function the final `save_user_list()`  seems redundant, because every `add_to_user_database()` will already invoke that (for each executable, before adding `path`).

I don't remember why we originally decided to save the file for every database update. It should be independent IMO.

Also, currently we do the saving **twice** for every processor, because `add_to_user_database()` also invokes `list_available()`, which now finally invokes `save_user_list()` as well.

MehmedGIT: > But in this function the final save_user_list() seems redundant, because every add_to_user_database() will already invoke that (for each executable, before adding path).

True, it seems redundant, but that was part of the fix for the missing resources of #1251. Still not sure why. Also, `add_to_user_database()` is not called for resources found at module level.

> I don't remember why we originally decided to save the file for every database update. It should be independent IMO.

Agree. I will optimize that when I get there. There are unnecessary saves to and loads from the yaml file for each discovered resource.

> Also, currently we do the saving twice for every processor, because add_to_user_database() also invokes list_available(), which now finally invokes save_user_list() as well.

Right. I have also spotted that in the logs. I think a simpler search method is needed instead of relying on the `list_available()` method, which also has other side effects.

bertsky: offtopic, but just a few lines above (I cannot suggest outside of the narrow diff hunk context):

instead of...
```python
elif str(res_filename.parent) == moduledir:
```
...please write...
```python
elif str(res_filename.parent).startswith(moduledir):
```
(because there are many module-provided data files that are in subdirectories, which now end up as `cwd` resources)

src/ocrd_utils/os.py
```diff
@@ -151,7 +151,7 @@ def list_all_resources(executable, moduled=None, xdg_data_home=None):
         if datadir.is_dir():
             candidates += datadir.iterdir()
     if 'system' in resource_locations:
-        systemdir = Path('/usr/local/share/ocrd-resources', executable)
+        systemdir = Path(RESOURCES_DIR_SYSTEM, executable)
         if systemdir.is_dir():
             candidates += systemdir.iterdir()
     if 'module' in resource_locations and moduled:
```
bertsky, line 157: btw, we should include `'*.egg-info/*', 'copyright.txt', 'LICENSE*', 'README.md',` in the list of anti-patterns below

MehmedGIT:
Noted and appended your comment as a reference to the top. Thanks!

src/ocrd/resource_manager.py
bertsky: > > But in this function the final save_user_list() seems redundant, because every add_to_user_database() will already invoke that (for each executable, before adding path).
> 
> True, it seems redundant, but that was part of the fix for the missing resources of #1251.

Yes, that did help – but [does not](https://github.com/OCR-D/core/pull/1309#issuecomment-2717959747) search for and add executables to the database in all cases. So I would still consider this behaviour an open matter.

> Still not sure why. Also, `add_to_user_database()` is not called for resources found at module level.

True. The more I think about it, the less I understand what might have been the original idea behind the user database. 

Clearly, @kba wanted to save the time of searching executables and resources repeatedly, hence the shortcuts. But not being able to add executables to the database seems broken, and not having module resources show up as registered/installed resources, too.

> > I don't remember why we originally decided to save the file for every database update. It should be independent IMO.
> 
> Agree. I will optimize that when I get there. There are unnecessary saves to and loads from the yaml file for each discovered resource.

Yes, once we have a clearer idea about the database lifetime, it should be easy to reduce file I/O.

> > Also, currently we do the saving twice for every processor, because add_to_user_database() also invokes list_available(), which now finally invokes save_user_list() as well.
> 
> Right. I have also spotted that in the logs. I think a simpler search method is needed instead of relying on the `list_available()` method, which also has other side effects.

...like short-cutting via `ocrd-all-tool.json` if available?

> I also do not like the database deduplication method. Preventing duplication should perform better than adding and then trying to remove duplications afterwards.

As discussed in the chat, the list (instead of dict) structure (and hence deduplication) might have been meant to allow keeping additional user-defined versions of the same resource name.

bertsky:
@kba what? wait! This was still a **draft**. And shouldn't you have merged https://github.com/OCR-D/core/pull/1315 into this first?

bertsky:
https://github.com/OCR-D/core/releases/tag/v3.2.0 is also bugus – it does _not_ include #1315

MehmedGIT:
> @kba what? wait! This was still a draft. And shouldn't you have merged https://github.com/OCR-D/core/pull/1315 into this first?

I did not react because I can also create another PR for the networt client, request forwarding over the Processing Server, etc. However, I am also fine if this PR is reverted back and I continue here.

kba:
Yeah, I messed up, this was not intentional. I'll try to fix it :(

kba:
> @kba what? wait! This was still a **draft**. And shouldn't you have merged #1315 into this first?

I did not realize #1315 was based on #1309 was the problem.

> https://github.com/OCR-D/core/releases/tag/v3.2.0 is also bugus – it does not include https://github.com/OCR-D/core/pull/1315

`master` (and the release) does include #1315, that PR is only open because it is based on #1309 and #1315 has not been merged into #1309.

I revert the merge and release a hotfix without it.

We'll need to rename the branch and open a new PR though, GitHub does not allow continuing working on a merged branch AFAIK.

I should (and will from now on) create release branches so I spot these oversights beforehand from now on.

MehmedGIT:
> We'll need to rename the branch and open a new PR though, GitHub does not allow continuing working on a merged branch AFAIK.

@kba Whatever is faster and esier on your end. I am fine with a new brach and PR to continue working on the RM Server.

bertsky:
> `master` (and the release) does include #1315, that PR is only open because it is based on #1309 and #1315 has not been merged into #1309.

Oh, I see. Sorry, did not notice.

> I revert the merge and release a hotfix without it.

That's to avoid having an incomplete version of the ResourceManagerServer in master and release, right?

If so, I'm for that.

> We'll need to rename the branch and open a new PR though, GitHub does not allow continuing working on a merged branch AFAIK.

Ok, if that's required, so be it. We should be careful not to loose sight of our discussion, inasfar as it still matters, e.g.
- https://github.com/OCR-D/core/pull/1309#issuecomment-2717959747 on how to actually get to #1251
- https://github.com/OCR-D/core/pull/1309#issuecomment-2697603396 on a refactored common base function ocrd.resource_manager.download for both ocrd.cli.resmgr.download and ocrd_network.ResourceManagerServer.download_resource (perhaps excluding the `*` behaviour though)

MehmedGIT:
Reopening the PR after the accidental merge.

PR #1311, opened 2025-03-04
update OCR-D/spec to v3.26.0

kba:


kba:
Test failures due to brownout by GitHub nudging us to update away from 20.04 runners by April 1st. Which we will do but after the next release.

PR #1314, opened 2025-03-10
fix deps-tf1 requirements DIRSTACK fix

bertsky:
sry about that!


bertsky:
@kba I believe this should fix all the cuda- Docker CDs. But I'll trigger the Docker Image CI test to see if that's indeed the case.

bertsky:
> to see if that's indeed the case.

it works – https://github.com/OCR-D/core/actions/runs/13769869312/job/38505546895

PR #1315, opened 2025-03-12
fix name vs subdir, add type checking for resource candidates

bertsky:


bertsky:
I don't know how @kba originally conceived of this, but I was surprised to find that `list_all_resources` did normally dive into directories twice, but then gave us just the basename as resource name. This resulted in Calamari models (which are directories that have subdirectories) to be split up. The only place that needs a recursive dive are module directories IMO. But then that subpath **is** part of the model name, because it is **resolved** against the module directory. I had to refactor a bit to properly discern the base path (directory of the resource location) from the subpath (resource name).

The rest is mostly the filtering of false positives via MIME types and suffixes.

note: we still have to discuss where the log.info message for unregistered resources is more appropriate:
- in `handle_resource` when there is no `size` key – https://github.com/bertsky/core/commit/0635b0fe2469c1ccf17711c7ccdc6915cc39816b
- in `add_to_user_database` when it was established that this entry is new – https://github.com/OCR-D/core/pull/1315/commits/8b66e1b97d875f894cc15d40d97b7dfa33ca7bc9

bertsky:
For the CI failure I believe we just need to update repo/assets from master...

src/ocrd/resource_manager.py
```diff
@@ -62,6 +62,12 @@ def __init__(self, userdir=None, xdg_config_home=None, xdg_data_home=None, skip_
                 self.save_user_list()
             self.load_resource_list(self.user_list)
 
+    def __repr__(self):
+        return f"user_list={str(self.user_list)} " + \
+               f"exists={self.user_list.exists()} " + \
+               f"database: {len(self.database)} executables " + \
+               f"{sum(map(len, self.database.values()))} resources"
```
MehmedGIT, line 69: This is a neat solution that may be used as a response when checking the resource manager's current state. I assume we no longer want to get rid of the database, or?

src/ocrd/resource_manager.py
```diff
@@ -196,30 +198,28 @@ def list_installed(self, executable: str = None):
...
-                    resdict = self.add_to_user_database(this_executable, res_filename, resource_type=res_type)
+                    resdict = self.add_to_user_database(this_executable, res_name, res_filename, resource_type=res_type)
                 # resdict['path'] = str(res_filename)
                 reslist.append(resdict)
             ret.append((this_executable, reslist))
         self.save_user_list()
         return ret
 
-    def add_to_user_database(self, executable, res_filename, url=None, resource_type='file'):
+    def add_to_user_database(self, executable, res_name, res_filename, url=None, resource_type='file'):
```
MehmedGIT, line 208: Why do we need the additional parameter `res_name` if we can already get it from `res_filename`? If it is for flexibility reasons, I would suggest having `res_name` with a default value of `Path(res_filename).name` unless there is a reason not to do that.

src/ocrd_utils/os.py
```diff
@@ -123,19 +127,38 @@ def list_resource_candidates(executable, fname, cwd=getcwd(), moduled=None, xdg_
         candidates.append(join(moduled, fname))
     return candidates
 
-def list_all_resources(executable, moduled=None, xdg_data_home=None):
+def list_all_resources(executable : str, ocrd_tool : Optional[Dict[str, Any]] = None, moduled : Optional[str] = None, xdg_data_home : Optional[str] = None) -> List[Tuple[str,str]]:
```
MehmedGIT, line 130: I avoid whitespaces before semicolons when `typing` to fit more parameters on a single line for the longer method signatures. Also, with `Optional[]` typing, passing `None` as a default is unnecessary; it is `None` by default.

src/ocrd_utils/os.py
```diff
@@ -144,37 +167,88 @@ def list_all_resources(executable, moduled=None, xdg_data_home=None):
...
+                    'TAGS', '.DS_Store',
+                    # C extensions
+                    '*.so',
+                    # translations
+                    '*.mo', '*.pot',
+                    '*.log', '*.orig', '*.BAK',
+                    '.git/*',
+                    # our stuff
+                    'ocrd-tool.json',
+                    'environment.pickle', 'resource_list.yml', 'lib.bash']):
```
MehmedGIT, line 204: Should we move the long list to `ocrd_utils.constants.py`?

src/ocrd_utils/os.py
```diff
@@ -186,13 +260,16 @@ def get_processor_resource_types(executable, ocrd_tool=None):
...
             return ['*/*']
         ocrd_tool = get_ocrd_tool_json(executable)
-    if not next((True for p in ocrd_tool.get('parameters', {}).values() if 'content-type' in p), False):
-        # None of the parameters for this processor are resources (or not
-        # the resource parameters are not properly declared, so output both
-        # directories and files
+    mime_types = [mime
+                  for param in ocrd_tool['parameters'].values()
+                  if param['type'] == 'string' and param.get('format', '') == 'uri' and 'content-type' in param
+                  for mime in param['content-type'].split(',')]
```
MehmedGIT, line 266: Why `ocrd_tool['parameters']` instead of the safer `ocrd_tool.get('parameters', {})`?

MehmedGIT (approved):
> note: we still have to discuss where the log.info message for unregistered resources is more appropriate:
    * in handle_resource when there is no size key – https://github.com/bertsky/core/commit/0635b0fe2469c1ccf17711c7ccdc6915cc39816b
    * in add_to_user_database when it was established that this entry is new – https://github.com/OCR-D/core/commit/8b66e1b97d875f894cc15d40d97b7dfa33ca7bc9
    
From my understanding, the database will become more consistent and avoid duplicate entries? If that is the case, I would prefer the logging to happen inside `add_to_user_database`.

src/ocrd/resource_manager.py
bertsky: For example, in the Tesseract case, the [pretrained script models](https://github.com/tesseract-ocr/tessdata_best/tree/main/script) are usually referenced as e.g. `script/Latin.traineddata`. But we could of course flatten this in our tessdata installations. (And we could also try to get rid of the `configs/` resource...)

Another example is the new ocrd_pagetopdf's `data/AletheiaSans.ttf`, but that could just be a silent default again.

All other module resources I can find are in the top level. So perhaps we should reconsider the recursive dive in `list_all_resources` at the module location – if we make that flat as well, we can switch back to the `res_filename.name` paradigm (and simple strings instead of location-name tuples)...

src/ocrd_utils/os.py
bertsky: Ok, I saw different styles so I was confused. I also like fewer whitespace, but in the past some linters gave me a hard time because of that...

The bit about `None` default I don't understand, though. IIUC a `=None` tells Python that it's a kwarg, rather than a positional one. That should be independent of type hints. And if I use `=None`, then it *must* be `Optional` or `|None` otherwise the type guard will complain.

src/ocrd_utils/os.py
bertsky: If it is going to be re-used, yes. But I doubt that...

src/ocrd_utils/os.py
bertsky: I was under the false impression that we can rely on OcrdToolValidator generating these as defaults. But what we actually get is the raw dict from the JSON, so you are right, it should be that

src/ocrd/resource_manager.py
MehmedGIT: > But we could of course flatten this in our tessdata installations. (And we could also try to get rid of the configs/ resource...)

Not sure how complicated that would be to change. 

> if we make that flat as well, we can switch back to the res_filename.name paradigm (and simple strings instead of location-name tuples)...

Not necessary to switch back, I guess; Nested resources should still be handled properly by `core` if there was no spec guiding how to store the resources.

src/ocrd_utils/os.py
MehmedGIT: > Ok, I saw different styles so I was confused. I also like fewer whitespace, but in the past some linters gave me a hard time because of that...

Weird, because internal Python libraries, when typing, use less whitespace.

> IIUC a =None tells Python that it's a kwarg, rather than a positional one. That should be independent of type hints. And if I use =None, then it must be Optional or |None otherwise the type guard will complain.

Right. `Optional[type]` was a replacement for `Union[type, None]`, but not defining the default value. My bad.

src/ocrd/resource_manager.py
bertsky: > > But we could of course flatten this in our tessdata installations. (And we could also try to get rid of the configs/ resource...)
> 
> Not sure how complicated that would be to change.

We moved away from installing under tessdata/script/ a while ago. Some users might have to update their installations, but it would not be hard to enforce (code-wise). 

And the `configs` download was just a convenience for the standalone CLI – we could just make that part of the installation procedure, without a resmgr entry.

> > if we make that flat as well, we can switch back to the res_filename.name paradigm (and simple strings instead of location-name tuples)...
> 
> Not necessary to switch back, I guess; Nested resources should still be handled properly by `core` if there was no spec guiding how to store the resources.

What is astonishing is that the [current spec](https://github.com/OCR-D/spec/blob/master/ocrd_tool.md#file-parameters) states pretty much the converse of what was implemented:
- module resources must be directly under the top-level module directory
- all other resources can be arbitrary relative paths

So if we want to stick with that, we would need to keep my location-name tuple structure, but change the iterators in `list_all_resources` to support recursion. (But it might be hard to get the level right: take Calamari for example again – you must not descend into `fraktur_19th_century/?.ckpt/`. So the actual maxlevel seems to then depend on the processor, which ocrd_utils does not know anything about.)

I am inclined to rather change the spec: flat top-level everywhere.

MehmedGIT: > I am inclined to rather change the spec: flat top-level everywhere.

Agree!

bertsky: So in accordance with https://github.com/OCR-D/spec/pull/263 I changed back the retval of `list_all_resources` to list of str, and disabled recursive moduled discovery – all locations but CWD are flat. I also added some basic tests for `list_all_resources`, which I found completely lacking.

bertsky:
> From my understanding, the database will become more consistent and avoid duplicate entries? If that is the case, I would prefer the logging to happen inside `add_to_user_database`.

It kept it there and removed it outside.

PR is ready from my side. We still have the other issues, this just tries to get some sense into the path resolution rules.

kba:
Merging this into #1309 branch which I'll reopen with the revert reverted. Again, sorry for the inconvenience.

PR #1319, opened 2025-03-28
Continuation of #1309:  Implementation of the resource manager server (issue #1294)

kba:
This branch should be back in a state it can be worked on.

I did:
- base it on #1309
- merged #1315 into it (which I haven't yet pushed to the original #1309 branch so as not to close it in case I did something wrong here)
- merged `master` into it
- reverted the revert commit, restoring the changes of #1309 and #1315 

I.e. we should be back where we were before the accidental merge, plus the @bertsky's commits from yesterday.

Only the newer commits show up here because they are already part of `master` though, but they are there and reflected in the worktree.


MehmedGIT (approved):
Everything seems to work fine and can be merged.

PR #1320, opened 2025-03-31
1294 impl rm server without reverts

kba:


kba:
No benefit in this vs. #1319, we'll have to do the double revert anyway, so best to continue work with the directly-master-mergeable #1319.

PR #1321, opened 2025-04-08
Processor: allow empty output file grp

bertsky:
Spec actually says this is legitimate. We have processors like ocrd-docstruct, or might have evaluators etc. that simply do not produce output in the form of a file group.

### DIFF
--- a/src/ocrd/decorators/__init__.py
+++ b/src/ocrd/decorators/__init__.py
@@ -107,7 +107,7 @@ def ocrd_cli_wrap_processor(
     if 'parameter_override' in kwargs:
         set_json_key_value_overrides(kwargs['parameter'], *kwargs.pop('parameter_override'))
     # Assert -I / -O
-    if 'input_file_grp' not in kwargs:
+    if not kwargs.get('input_file_grp', None):
         raise ValueError('-I/--input-file-grp is required')
     if 'output_file_grp' not in kwargs:
         raise ValueError('-O/--output-file-grp is required') # actually, it may be None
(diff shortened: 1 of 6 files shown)


kba (approved):
I ran into this same issue with https://github.com/OCR-D/ocrd_anybaseocr/pull/112.

src/ocrd/decorators/__init__.py
```diff
@@ -107,10 +107,10 @@ def resolve(name):
     if 'parameter_override' in kwargs:
         set_json_key_value_overrides(kwargs['parameter'], *kwargs.pop('parameter_override'))
     # Assert -I / -O
-    if not kwargs['input_file_grp']:
+    if 'input_file_grp' not in kwargs:
```
kba, line 110:
```suggestion
    if not kwargs.get('input_file_grp', None):
```

as discussed

kba: `ocrd-olahd-client` requires neither input nor output. But since that is a fairly esoteric processor that no one seems to be using, probably not reason enough to change that behavior in spec/core.

@MehmedGIT @j-panzer Is the OLA-HD client still in use?

MehmedGIT: @kba, no it is not used by Operandi nor the Integration script.

bertsky: Perhaps we could treat this as a special case: do require an input fileGrp, but only for informational purposes or to represent workflow dependency (final fileGrp), not for actually doing something specific with that fileGrp.

So instead of `input_file_grp_cardinality: 0` this would become 1.

It would require users specify an input fileGrp, but that should normally not be a problem – either pass the "maximal" fileGrp of the workflow, or just use `OCR-D-IMG`.

(Just to make sure we don't bend the spec because of this.)

Issue #1324, opened 2025-04-17
local_filename quirks when METS Server is present


bertsky:
In a workflow enabled by METS Server, with fully up-to-date versions, I get the following quite surprising stacktrace:
```
11:06:15.047 INFO ocrd.processor.base - processing page PHYS_0001
11:06:15.111 ERROR ocrd.processor.base - Failure on page PHYS_0001: cannot identify image file 'None'
Traceback (most recent call last):
  File "/data/ocr-d/ocrd_all/core/src/ocrd/processor/base.py", line 710, in process_workspace_handle_page_task
    task.result()
  File "/data/ocr-d/ocrd_all/core/src/ocrd/processor/base.py", line 124, in result
    return self.fn(*self.args, **self.kwargs)
  File "/data/ocr-d/ocrd_all/core/src/ocrd/processor/base.py", line 1157, in _page_worker
    _page_worker_processor.process_page_file(*input_files)
  File "/data/ocr-d/ocrd_all/core/src/ocrd/processor/base.py", line 793, in process_page_file
    page_ = page_from_file(input_file)
  File "/data/ocr-d/ocrd_all/core/src/ocrd_modelfactory/__init__.py", line 102, in page_from_file
    return page_from_image(input_file)
  File "/data/ocr-d/ocrd_all/core/src/ocrd_modelfactory/__init__.py", line 59, in page_from_image
    exif = exif_from_filename(input_file.local_filename)
  File "/data/ocr-d/ocrd_all/core/src/ocrd_modelfactory/__init__.py", line 39, in exif_from_filename
    with Image.open(image_filename) as pil_img:
  File "/data/ocr-d/ocrd_all/venv38/lib/python3.8/site-packages/PIL/Image.py", line 3298, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file 'None'
11:06:15.141 ERROR ocrd.processor.helpers.run_processor - Failure in processor 'ocrd-segment-replace-page'
```

So none of our prior checks to prevent exactly that seem to have taken effect:
https://github.com/OCR-D/core/blob/769919a659a82462066f3395afc55b405ea31c4a/src/ocrd_modelfactory/__init__.py#L55-L59
https://github.com/OCR-D/core/blob/769919a659a82462066f3395afc55b405ea31c4a/src/ocrd_modelfactory/__init__.py#L37-L39

[...]

PR #1325, opened 2025-04-17
mets_server.OcrdFileModel: do not convert None to str

bertsky:
~~partial~~ complete fix for #1324


src/ocrd/mets_server.py
```diff
@@ -46,7 +46,7 @@ def create(
     ):
         return OcrdFileModel(
             file_grp=file_grp, file_id=file_id, page_id=page_id, mimetype=mimetype, url=url,
-            local_filename=str(local_filename)
+            local_filename=str(local_filename) if local_filename else None
```
kba, line 49: Sensible in any case. `"None"` helps no one.

bertsky:
We should add comprehensive tests for this entire area.

PR #1329, opened 2025-05-07
--page-id negation operator and logical structmap attribute matching

### LINKED ISSUE #1328: allow filtering by `mets:div/@TYPE`
In case the METS already contains a **logical structMap** (or we run a processor like `ocrd-anybaseocr-layout-analysis` as first step), there will be information on the type of every page (content vs title vs empty vs cover etc). This in turn might be very useful to allow filtering the page range when processing – complementary to the `--page-id` selection mechanism we already have (i.e. list or regex of physical page `@ID`, `@ORDER`, `@ORDERLABEL`, `@LABEL`, or `@CONTENTIDS`).

We might tie the list of types to consider to [DFG Strukturdatenset](https://dfg-viewer.de/strukturdatenset) – but in many cases, libraries extend that anyway (or use different spellings, like titlecase).

For example, a user may want to rule out processing `cover_front,cover_back,binding` pages. (Not only to avoid wasting resources, but because some processors may not handle these well, or even crash.)

Our current syntax does not allow for negative selection, so we'd need some new operator. I suggest `~`, since `!` might be interpreted by the shell via history expansion (even when quoted to prevent pathname expansion).

For `..` range expressions, negation only makes sense when applied to the entire expression. For comma-separated lists, every atom could be negated individually. And for regexes, the entire expression.

So how about `--page-id ~PHYS_0001..PHYS_0004`, `--page-id ~cover_front,~cover_back,~binding`, `--page-id "~//(cover_(front|back)|binding)"`?

(We might even want to support selecting logical div `@ID` in addition to `@TYPE`, e.g. `--page-id ~LOG_0001..LOG_0005` ...)

bertsky:
fixes #1328, depends on https://github.com/OCR-D/assets/pull/104


kba (approved):
Very impressive and powerful. `~`-negation is useful in itself but the combination with search patterns on logical `mets:div/@TYPE` attributes is awesome, because we can exclude bindings, colour checker and more directly with a `--page-id` option.

kba (approved):

bertsky:
@kba Just to be certain: we do want to keep the lower-cased syntax (e.g. `physical:orderlabel:` instead of `PHYSICAL:ORDERLABEL:`), despite the fact that in METS these are all upper case? (Changing that is of course trivial, but should happen before introducing this new feature.)

kba:
> @kba Just to be certain: we do want to keep the lower-cased syntax (e.g. `physical:orderlabel:` instead of `PHYSICAL:ORDERLABEL:`), despite the fact that in METS these are all upper case? (Changing that is of course trivial, but should happen before introducing this new feature.)

Yes, lowercase is different from how it appears in METS, but it is easier to type, consistent and less error-prone. We might even want to `pageId_token = pageId_token.lower()` where applicable but the logic for the patterns is already pretty complex, so let's not :)

tests/model/test_ocrd_mets.py
```diff
@@ -104,17 +123,82 @@ def test_find_all_files(sbb_sample_01):
...
+    assert len(mets.find_all_files(pageId='physical:id://PHYS_0005,physical:id:PHYS_0001..PHYS_0002')) == 35, '35 files in //PHYS_0005,PHYS_0001..PHYS_0002'
+    assert len(mets.find_all_files(pageId='physical://PHYS_0005,physical:PHYS_0001..PHYS_0002')) == 35, '35 files in //PHYS_0005,PHYS_0001..PHYS_0002'
+    assert len(mets.find_all_files(pageId='physical:id://PHYS_0005,physical:id:PHYS_0001..PHYS_0002')) == 35, '35 files in //PHYS_0005,PHYS_0001..PHYS_0002'
+    assert len(mets.find_all_files(pageId='physical:1..10')) == 35, '35 files in @ORDER range 1..10'
+    assert len(mets.find_all_files(pageId='physical:order:1..10')) == 35, '35 files in @ORDER range 1..10'
+    assert len(mets.find_all_files(pageId='physical:1..5')) == 35, '35 files in @ORDER range 1..10'
+    assert len(mets.find_all_files(pageId='physical:order:1..5')) == 35, '35 files in @ORDER range 1..10'
+    assert len(mets.find_all_files(pageId='physical:page 1..page 2,physical:5')) == 35, '35 in PHYS_0001,PHYS_0002,PHYS_0005'
+    assert len(mets.find_all_files(pageId='physical:orderlabel:page 1..page 2,physical:order:5')) == 35, '35 in PHYS_0001,PHYS_0002,PHYS_0005'
```
kba, line 164: Maybe we should put these tests on the website, since the syntax is getting powerful and complex and few users are probably aware of the extensive unit tests.

bertsky: Yes, this should be documented extensively. I recommend the Readme in src/ocrd_models or src/ocrd as nearest place. Next off, point to that from the ocrd-website pages (guides or Wiki).

PR #1330, opened 2025-05-07
skip input files

bertsky:
This is a case of not seeing the forest for the trees: In the v3 changes, we did introduce `MissingInputFile` exception (for `OCRD_MISSING_INPUT=ABORT`). But this only applies narrowly: if there is no `mets:file` in one of the input fileGrps.

However, the more important case seems to be that there is a `mets:file` (regardless of the number of input fileGrps), but it is not available locally, and `OCRD_DOWNLOAD_INPUT=False` was set deliberately – to either trigger `OCRD_MISSING_INPUT=ABORT` or `OCRD_MISSING_INPUT=SKIP`. 

But we did not implement this case yet – instead, we let the function proceed without download, which will then fail (without the `SKIP` behaviour) at `page_from_file`.

So now the ABORT and SKIP behaviours are enforced for locally missing files, too.

### DIFF
--- a/src/ocrd/processor/base.py
+++ b/src/ocrd/processor/base.py
@@ -783,11 +783,16 @@ class Processor():
         page_id = input_files[input_pos].pageId
         self._base_logger.info("processing page %s", page_id)
         for i, input_file in enumerate(input_files):
+            grp = self.input_file_grp.split(',')[i]
             if input_file is None:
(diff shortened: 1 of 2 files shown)


bertsky:
I don't understand what's going on with the [network integration tests](https://github.com/OCR-D/core/actions/runs/15161980594)...

PR #1331, opened 2025-06-18
Fix network agent type click

bertsky:
attempts to fix the recent CI failures of the network integration test

### DIFF
--- a/src/ocrd/decorators/ocrd_cli_options.py
+++ b/src/ocrd/decorators/ocrd_cli_options.py
@@ -56,7 +56,7 @@ def ocrd_cli_options(f):
         # subcommands. So we have to work around that by creating a
         # pseudo-subcommand handled in ocrd_cli_wrap_processor
         argument('subcommand', nargs=1, required=False,
-                 type=click.Choice([AgentType.PROCESSING_WORKER, AgentType.PROCESSOR_SERVER])),
+                 type=click.Choice(list(map(str, AgentType)))),
     ]
     for param in params:
         param(f)
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,5 +74,5 @@ COPY requirements_test.txt .
 RUN pip install -r requirements_test.txt
 RUN mkdir /ocrd-data && chmod 777 /ocrd-data
 
-CMD ["yes"]
+CMD yes > /dev/null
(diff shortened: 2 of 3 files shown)


src/ocrd/decorators/ocrd_cli_options.py
```diff
@@ -56,7 +56,7 @@ def cli(**kwargs):
         # subcommands. So we have to work around that by creating a
         # pseudo-subcommand handled in ocrd_cli_wrap_processor
         argument('subcommand', nargs=1, required=False,
-                 type=click.Choice([AgentType.PROCESSING_WORKER, AgentType.PROCESSOR_SERVER])),
+                 type=click.Choice(list(map(str, AgentType)))),
```
MehmedGIT, line 59: Why does the previous implementation no longer work? Is it because of the updated click version?

MehmedGIT (approved):
> Ideally we should also add some direct tests for the worker CLIs. 

What exactly do you mean with that? 

> And when the network integration services do not start up successfully, we still need a proper error message... (currently just the `docker exec ... pytest` waits indefinitely)

Would it help if all docker test container logs of the processing server and processing workers are written to a file, the printed by the `ocrd_core_test` instance instead of `yes`?

src/ocrd/decorators/ocrd_cli_options.py
bertsky: Click 8.2 introduced direct support for enum types (without the `list` wrapper and the `.value` function). But that is not even available on Python 3.8 or 3.9. And it does not explain why the previous setting ever worked. I don't know TBH.

bertsky:
> > Ideally we should also add some direct tests for the worker CLIs.
> 
> What exactly do you mean with that?

A new test case for the processor CLI like `ocrd-dummy worker --help` and `ocrd-dummy worker --database mongodb://... --queue amqp://...`

> > And when the network integration services do not start up successfully, we still need a proper error message... (currently just the `docker exec ... pytest` waits indefinitely)
> 
> Would it help if all docker test container logs of the processing server and processing workers are written to a file, the printed by the `ocrd_core_test` instance instead of `yes`?

Yes! That would be ideal. With that we could directly `timeout 40s grep` in the log file for certain events before running pytest.

PR #1334, opened 2025-07-05
run_workflow: pass comma-separated page ids instead of interval range

### LINKED ISSUE #1333: Processing Server: expand_page_ids is overly strict
I have a workspace with **heterogeneous page identifiers**:
- f_1875-Himnos-EMU-0040
- f_1878-Helpin-EMU-0027
- f_1890-Evange-UMI-0007
- f_1892-Womans-EMU-0015
- f_192x-Gospel-MTS-0007

When I try to process this in the Processing Server, it fails, because it cannot convert this to a **single numerical range**:
```
   File "/build/core/src/ocrd_network/processing_server.py", line 744, in run_workflow
     responses = await self.task_sequence_to_processing_jobs(
   File "/build/core/src/ocrd_network/processing_server.py", line 693, in task_sequence_to_processing_jobs
     response = await self.validate_and_forward_job_to_network_agent(
   File "/build/core/src/ocrd_network/processing_server.py", line 431, in validate_and_forward_job_to_network_agent
     page_ids = expand_page_ids(data.page_id)
   File "/build/core/src/ocrd_network/utils.py", line 59, in expand_page_ids
     page_ids += generate_range(*page_id_token.split(sep='..', maxsplit=1))
   File "/build/core/src/ocrd_utils/str.py", line 231, in generate_range
     raise ValueError(f"Range '{start}..{end}' differ in their non-numeric part: '{start[:-len(start_num)]}' != '{end[:-len(end_num)]}'")
 ValueError: Range 'f_1875-Himnos-EMU-0040..f_192x-Gospel-MTS-0007' differ in their non-numeric part: 'f_1875-Himnos-EMU-' != 'f_192x-Gospel-MTS-'
```

This is with `page_wise=False`.

IMO if splitting is necessary, it should be across integer ranges (pure ordinal page numbers), ignoring page IDs.

bertsky:
fixes #1333

### DIFF
--- a/src/ocrd_network/processing_server.py
+++ b/src/ocrd_network/processing_server.py
@@ -91,7 +91,7 @@ class ProcessingServer(FastAPI):
         log_file = get_processing_server_logging_file_path(pid=getpid())
         configure_file_handler_with_formatter(self.log, log_file=log_file, mode="a")
 
-        self.log.info(f"Loading ocrd all tool json")
+        self.log.info("Loading ocrd-all-tool.json")
         self.ocrd_all_tool_json = load_ocrd_all_tool_json()
         self.hostname = host
         self.port = port
@@ -208,7 +208,8 @@ class ProcessingServer(FastAPI):
             methods=["DELETE"],
             tags=[ServerApiTags.WORKFLOW, ServerApiTags.PROCESSING],
             status_code=status.HTTP_200_OK,
-            summary="!! Workaround Do Not Use Unless You Have A Reason !! Kill all METS servers on this machine that have been created more than 60 minutes ago."
+            summary="!! Workaround Do Not Use Unless You Have A Reason "
+            "!! Kill all METS servers on this machine that have been created more than 60 minutes ago."
         )
         self.include_router(others_router)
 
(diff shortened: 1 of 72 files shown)


src/ocrd_network/processing_server.py
```diff
@@ -736,9 +736,7 @@ async def run_workflow(
         self.validate_tasks_agents_existence(processing_tasks, agent_type)
 
         page_ids = get_page_ids_list(self.log, mets_path, page_id)
-
-        # TODO: Reconsider this, the compact page range may not always work if the page_ids are hashes!
-        compact_page_range = f"{page_ids[0]}..{page_ids[-1]}"
+        compact_page_range = ','.join(page_ids)
```
MehmedGIT, line 739: That is a good fix for non-integer page_ids. However, for integer page_ids it will bloat the database with unnecessarily big `compact_page_range` values. Instead of `PHYS_0001..PHYS_1234`, the value will be a very big string `PHYS_0001,PHYS_0002,PHYS_0003...PHYS_1234`. Is there maybe a way to keep it as it is for integer page_ids, and join for non-integer ones? For the sake of optimization.

MehmedGIT (changes requested):

src/ocrd_network/processing_server.py
MehmedGIT: @bertsky, if that works, yes!

src/ocrd_network/process_helpers.py
```diff
@@ -28,23 +30,58 @@ def invoke_processor(
...
+handlers=
+qualname=shapely.geos
+[handler_fileHandler]
+class=FileHandler
+formatter=defaultFormatter
+args=('{log_filename}','a+')
+[formatter_defaultFormatter]
+format=%(asctime)s.%(msecs)03d %(levelname)s %(name)s - %(message)s
+datefmt=%H:%M:%S
+""".format(log_filename=log_filename))
```
MehmedGIT, line 67: Could you put that constant string as a variable above the method definition and just use the variable in the method?

MehmedGIT (approved):

src/ocrd_network/process_helpers.py
bertsky: Or perhaps another time – should I track this as a separate issue?

PR #1355, opened 2026-04-27
docker: install jq in base image

kba:
For doing JSON query ops in derived containers. Small and few dependencies and we do need it for the `readinessProbe` of the processing worker pods:

```bash
curl -u "*:*" -s http://$(OCRD_RABBITMQ_SERVICE_HOST):15672/api/queues | \
jq '.[] | select(.name == "$(OCRD_PROCESSOR_NAME)")' >/dev/null || exit 1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,6 +51,7 @@ RUN apt-get update && apt-get -y install software-properties-common \
         bzip2 \
         sudo \
         git \
+        jq \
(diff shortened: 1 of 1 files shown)


kba:
Test failures are due to the issue fixed in https://github.com/OCR-D/core/pull/1356

bertsky (approved):
Absolutely!

PR #1357, opened 2026-04-29
remove pinning cryptography, update paramiko to versions without the warning issue from #1260

kba:
The problem with paramiko raising warnings about deprecated cryptography algorithms should have been fixed since at least 3.4.1 (https://github.com/paramiko/paramiko/pull/2421).

Also, paramiko dropped support for Python 3.9 in 4.0.0, so another `python_version` conditional requirement.


bertsky (approved):
LGTM

PR #1358, opened 2026-05-05
ocrd network client workflow run: compare options in boolean context

kba:
Comparison between path to workflow and tasks provided must be in a boolean context, otherwise we're comparing string to boolean.

Bug introduced in https://github.com/OCR-D/core/pull/1337/changes#diff-71be56139c9cf750a2a524a33517ef092a7a010fc8115d79370c292fb7f01f5cL212

How I did not spot this before I don't know 😨


bertsky (approved):
absolutely

PR #1367, opened 2026-09-13
OcrdPage: ensure .etree gets the same ns_prefix as .pcgts

bertsky:
This is necessary to avoid producing PAGE output w/ mixed (and thus invalid) namespaces prefixes `pc` (from generateDS / DOM) and `page` (from lxml / etree) during serialization.

Use-case: ocrd-page-transform, which relies entirely on core here:

```python
    def process_page_pcgts(self, *input_pcgts: Optional[OcrdPage], page_id: Optional[str] = None) -> OcrdPageResult:
        pcgts = input_pcgts[0]
        result = self.xslt(pcgts.etree, **self.xsltparams)
        for error in self.xslt.error_log:
            self.logger.error(error)
        root = result.getroot()
        assert root is not None, "transform yields non-XML result; try setting `mimetype` parameter correctly"
        root = PcGtsType.factory()
        root.build(result.getroot())
        return OcrdPageResult(OcrdPage(root, result, {}, {}))
```

What used to happen:
1. ocrd_modelfactory's OcrdPage constructor would just call `parseEtree` without `nsmap` override; this created an  `.pcgts.ns_prefix_ = 'pc'` but an `.etree.prefix = 'page'`
2. XSLT processing would not change that
3. the new PcGtsType would have the `pc` prefix, but the passed-in etree to be built would still have `page`, causing a mixup

Since it is not allowed to change the `.prefix` in an lxml.etree.Element post-hoc, this must be done during construction. Hence the PR.


bertsky:
CI fails at:
```
Error: Unable to resolve action `homebrew/actions@master`, unable to find version `master`
```

I don't know what that means...

kba:
LGTM, thanks!

